PR #1573, opened 2015-04-10
allow users to optionally bypass the trash even if it is available on their OS

thisfred:
I have:

  - added a setting to bypass the trash, with a default of "false", which, when set to "true" will allow users to delete files, even if their OS supports using a trash bin.
  - renamed can_trash to use_trash

Since no tests existed around deleting/trashing files, as far as I could tell, I did not add any, but am happy to add them if you think it makes sense, and if you can tell me which mocking library you prefer.

I *did* test this setting manually, and confirmed that:

  1. With no action taken by the user, on non-darwin POSIX, QL still uses the trash by default.
  2. After changing `bypass_trash` in `.quodlibet/config` to `"true"`, QL shows the delete icon in the track right-click menu, and actually deletes the file when choosing that menu item.
  3. When removing the `bypass_trash` setting from `.quodlibet/config`, the behavior reverts to what it was before this setting existed.

Motiviation: I download and listen to a lot of new tracks almost daily. Consequently I delete a lot of files every day also to clear up disk space, and I don't want to have to take two actions (move to trash, empty trash) to do so.

### DIFF
--- a/quodlibet/quodlibet/util/trash.py
+++ b/quodlibet/quodlibet/util/trash.py
@@ -15,6 +15,7 @@ import time
 from os.path import join, islink, abspath, dirname
 from os.path import isdir, basename, exists, splitext
 
+from quodlibet import config
 from quodlibet.util.path import find_mount_point, xdg_get_data_home
 
 
@@ -121,9 +122,11 @@ def trash_free_desktop(path):
         raise
 
 
-def can_trash():
+def use_trash():
     """If the current platform supports moving files into a trash can."""
-    return (os.name == "posix" and sys.platform != "darwin")
+    return (
+        os.name == "posix" and sys.platform != "darwin" and
+        not config.getboolean("settings", "bypass_trash"))
 
 
 def trash(path):
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -114,6 +114,10 @@ INITIAL = {
         # tags which get searched in addition to the ones present in the
         # song list, separate with ","
         "search_tags": "",
+
+        # If set to "true" allow directly deleting files, even on systems that
+        # support sending them to the trash.
+        "bypass_trash": "false",
     },
     "rename": {
         "spaces": "false",
--- a/quodlibet/quodlibet/qltk/delete.py
+++ b/quodlibet/quodlibet/qltk/delete.py
@@ -141,7 +141,7 @@ class TrashDialog(WarningMessage):
 
 
 def TrashMenuItem():
-    return (MenuItem(_("_Move to Trash"), "user-trash") if trash.can_trash()
+    return (MenuItem(_("_Move to Trash"), "user-trash") if trash.use_trash()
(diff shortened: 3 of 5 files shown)


quodlibet/quodlibet/util/trash.py
```diff
@@ -121,9 +122,13 @@ def trash_free_desktop(path):
...
 
-def can_trash():
+def use_trash():
     """If the current platform supports moving files into a trash can."""
-    return (os.name == "posix" and sys.platform != "darwin")
+    if config.has_option("settings", "bypass_trash"):
+        bypass = config.getboolean("settings", "bypass_trash")
+    else:
+        bypass = False
+    return os.name == "posix" and sys.platform != "darwin" and not bypass
```
thisfred: One thing I considered and did not do yet, because it might be premature optimization, is move the bypass check into a separate method and call that as the third clause, so that we'd only have to do that work if the first two checks aren't False.

quodlibet/quodlibet/util/trash.py
```diff
@@ -121,9 +122,13 @@ def trash_free_desktop(path):
         raise
 
 
-def can_trash():
+def use_trash():
     """If the current platform supports moving files into a trash can."""
-    return (os.name == "posix" and sys.platform != "darwin")
+    if config.has_option("settings", "bypass_trash"):
```
lazka: This isn't needed. config keys defined in config.py are always available.

quodlibet/quodlibet/util/trash.py
lazka: This only gets called once, so performance shouldn't be an issue.

lazka:
I don't quite get the motivation as one click to clear the trash after many clicks for deleting files doesn't add much, but it seems like a small enough change.

A test would be nice (trash is sadly not tested..) but not required. See other tests doing config.init() in setUp. We currently don't use a mocking library, not sure if we need one...

declension:
Yeah, I'm not entirely seeing the usefulness of this either: we spent a while making it _safer_ to use delete in QL and this feels a little against that effort. That said I suppose if it's entirely hidden behind config (and tested) then there's little risk for others users.

As to mocking, there have been only a few times I've wanted a mocking framework for QL, but not used any in Python*. Whether that could significantly help with the intricacies of cross-platform file trashing / deletion, etc, anyway is another matter.

 
--
*Just noticed 3.3 has [unittest.mock](https://docs.python.org/3/library/unittest.mock.html), looks interesting, but that's some way off for us anyway... though there is a backport, hmm...

thisfred:
I removed the check for the setting, and added some tests. I didn't add mocking, since simple monkey patching seemed to already be used elsewhere in the tests and I didn't need anything more complicated.

Taking out the check for the existence of the setting broke test_qltk_delete.py, so I added a setUp and tearDown there to initialize and tear down the config. I'm not sure that's the right thing t

lazka:
The monkeypatching seem a bit excessive as we test on all platforms, but can't hurt I guess. Thanks

Issue #559, opened 2015-03-14
Use folder name as album key if MBID, labelid are missing


lazka:
Original [issue 559](https://code.google.com/p/quodlibet/issues/detail?id=559) created by foxy@on-the-hill.me.uk on 2010-08-13T20:53:24.000Z:

When there are two albums with the same name but by different artists in the collection, they are merged in the album view. They should be shown separately under different artists.

lazka:
We use albumartist now to separate albums with the same name (besides musicbrainz tags, which should always give the right result).

PR #1576, opened 2015-04-12
Makes album-heuristics consider date-tag

funkyfuture:
This is a simple heuristic that ensures that albums with the same name are considered distinct if they have different date-tags.

As by their nature albums have an unambigious release date, there should be nothing messed by this.

This solves e.g., different Albums named "In Dub" to be merged into one.
This should also solve the initial behaviour described in #559. Though this patch also allows for same-named albums to be in the same directory.


lazka:
We've added "albumartist" for this some time ago. Wouldn't that be enough?

I'm worried that this makes it harder to merge tracks in case of not fully tagged libraries and "date" being used as recording time and not the release time in some cases.

lazka:
Please open a bug report if you want to discuss this further.

PR #1583, opened 2015-04-19
Fix #1049 : allow to configure cover size in animosd plugin

### LINKED ISSUE #1049: Make size of album art image configurable in animosd plugin
Original [issue 1049](https://code.google.com/p/quodlibet/issues/detail?id=1049) created by dencrypt on 2012-11-01T14:06:18.000Z:

Just a comment in the code for anyone with big screens (think 24&quot;+) and not really satisfied with the tiny thumbnail showing beside the other info in animosd.

Might be suitable for other users as well though.

brunob:
For review, not sure if it's ok cause i have to preview twice in order to see the new size after setting change.


lazka:
Ideally (also considering the bug report) we should just be more smart about the size so no config is needed. But for now this seems fine.

PR #1593, opened 2015-04-23
Playlists Browser: enable searching of contents

### LINKED ISSUE #1588: Add search to Playlist Browser
I would like to search in playlists without switching the browser.

declension:
@lazka - as discussed really. Please have a look, we can either then merge or update first.

 * Add searchbar as per other browsers, mostly.
 * Query is normal QL syntax
 * Searches persist across playlists selection
 * Text-compatible filtering supported too (TODO: ~playlists tag should select playlist)

Closes #1588.

### DIFF
--- a/quodlibet/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/quodlibet/browsers/playlists/main.py
@@ -14,6 +14,10 @@ from quodlibet.plugins.playlist import PLAYLIST_HANDLER
 from quodlibet import config
 from quodlibet.browsers._base import Browser
 from quodlibet.formats._audio import AudioFile
+from quodlibet.qltk.completion import LibraryTagCompletion
+from quodlibet.qltk.searchbar import SearchBarBox
+from quodlibet.qltk.songlist import SongList
+from quodlibet.query import Query
 from quodlibet.util import connect_obj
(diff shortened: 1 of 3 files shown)


declension:
@lazka - Thanks. 

Weird, I can't replicate that UI behaviour; the queue stays shut for me, and if I don't show the songpane, I don't get a songpane at all it seems. 

As for the `active_filter()`... I see. Is it in fact possible to _remove_ it here entirely, as newly added songs will never be auto-added to a playlist (well, not without a whole lot new work)? That said, songs removed from the library should get removed from the songlist I guess, and it seems safer to implement it as a filter on both active playlist and text filter... so maybe I'll do that.

lazka:
I mean if you hide the queue completely, not just close the expander.

for ``active_filter()``, removals are still handled in the song list directly atm, so you can probably just remove it for now.

declension:
Oh, right, I forgot you could even _do_ that (again...)! 

I've fixed that behaviour in the packing (AFAICT), and as for the `active_filter`, made it consistent with the others for now.

Perhaps we can think about cleaning that interface up at some point? IMO it feels a bit inconsistent with the rest of the Browser interface, or something.

declension:
@lazka - are you happy for me to merge the updated version?

PR #1618, opened 2015-05-22
Update Dutch translation

plsng:
I'm not entirely sure on some of the terminology and there are probably some small errors here and there but at least it's better than before.


lazka:
Thanks

PR #1623, opened 2015-05-30
Add a song "energy" value that operates much like song rating.

ZDBioHazard:
This patch adds an arbitrary "energy" rating for songs that operates just like regular song ratings for most intents and purposes.

If a user has a wide gamut of musical tastes, this allows them to filter for more or less "intense" songs (or whatever they decide "energy" means), depending on what they are in the mood for, without making huge manual playlists, or usually unsatisfactory genre, artist or album filters.

This is probably too niche a feature to merge. Just thought I'd post it in case people are interested. I've been using it since 3.0.1 and have been enjoying the easy mood filtering.

### DIFF
--- a/quodlibet/po/POTFILES.in
+++ b/quodlibet/po/POTFILES.in
@@ -138,6 +138,7 @@ quodlibet/qltk/delete.py
 quodlibet/qltk/downloader.py
 quodlibet/qltk/edittags.py
 quodlibet/qltk/_editutils.py
+quodlibet/qltk/energymenu.py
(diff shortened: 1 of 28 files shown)


lazka:
This seems overly specific, but I don't even use ratings, so maybe someone else has some thoughts on this.

declension:
I agree. Rating is pretty much universal (albeit with widely varying implementations and ideas), but this is clearly custom tag territory.

However perhaps there is a case for allowing more configuration of custom tags though to allow them to behave in, say, more rating-like ways..

PR #1624, opened 2015-05-30
Add a plugin for editing ~#playcount and ~#skipcount.

ZDBioHazard:
This songs menu plugin lets you easily edit the `~#playcount` and `~#skipcount` of selected songs.

It supports incrementing multiple songs, and resets `~#laststarted` and `~#lastplayed` when setting `~#playcount` to 0.

### DIFF
--- a/quodlibet/po/POTFILES.in
+++ b/quodlibet/po/POTFILES.in
@@ -86,6 +86,7 @@ quodlibet/ext/songsmenu/cddb.py
 quodlibet/ext/songsmenu/console.py
 quodlibet/ext/songsmenu/custom_commands.py
 quodlibet/ext/songsmenu/duplicates.py
+quodlibet/ext/songsmenu/editplaycount.py
 quodlibet/ext/songsmenu/embedded.py
 quodlibet/ext/songsmenu/exact_rating.py
 quodlibet/ext/songsmenu/filterall.py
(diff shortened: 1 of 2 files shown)


declension:
Whilst I'm a little uneasy about hacking the internal tags, doesn't seem too dangerous. 

Could we not generalise this plugin to editing _any_ (numeric) tags?

ZDBioHazard:
> Could we not generalise this plugin to editing any (numeric) tags?

Sure, I can create a config UI with an editbox to specify a custom tag list and auto-generate the main UI. Though I'm not sure what other tags it might be useful to be able to change with an integer spinner.

Is there an official way to verify a tag is internal and numeric? If not, would this sort of logic be reliable?:
```python
# Would use USER_TAGS, but ~#playcount and ~#skipcount aren't in USER_TAGS
from quodlibet.util.tags import _TAGS

def is_internal_numeric_tag(name):
    try:
        tag = _TAGS[name]
    except KeyError:
        return False

    if tag.internal is True and tag.numeric is True:
            return True
    return False
```

declension:
Yes, you probably want:

    [t.name for t in _TAGS.values() if t.numeric and not t.internal]

Note it's the _non_-internal ones we want here (no point editing computed tags like `date`, `originalyear`) as having just looked at the tags they seem to make even less sense to edit (`~#length`, `~#bitrate`?). 

`~#added` is the perhaps the only other one I'd like to change personally (and quite related to this PR, perhaps), but that's really a date masquerading as a numeric.

I've created a separate issue as some of these tags now seem confused (and/or confusing. I'm confused about some of them either way...)

datakid:
Hey, love this plugin. Been looking for this since I installed QL and my #(playcount<1) filter was filled with long podcast backups.

Thanks.

Issue #1401, opened 2015-03-15
[PATCH] User option to disable rating hotkeys.


lazka:
Original [issue 1401](https://code.google.com/p/quodlibet/issues/detail?id=1401) created by zdbiohazard2 on 2014-05-16T19:58:30.000Z:

Not knowing the rating hotkey feature existed, I've accidentally rated a bunch of songs into oblivion over the years by hitting number keys with the wrong input focus.

I've attached a patch that adds an option to the preferences window to disable it. :)

For those of you that like the rating hotkeys, this patch also generates the hotkey list based on your rating precision setting, so now all of the 0-9 keys will work, depending on how many rating levels you use. :D

PR #1625, opened 2015-05-30
Add a setting to enable/disable rating hotkeys.

ZDBioHazard:
This is sort-of a continuation of #1401.

Changing the hotkeys to `<ctrl>-[0-x]` nearly eliminated mis-ratings for me, but now I have another app that uses `<ctrl>-[0-9]` in it's UI, so I updated this patch just to be safe. :wink:


lazka:
Would you be OK with it not getting exposed in the UI and only in the config file?

ZDBioHazard:
Works for me, though I just had a thought that I don't think the rating hotkeys are mentioned anywhere in-UI, so the option being present in the preferences dialog might clue users in that rating hotkeys are a thing they can use.

PR #1627, opened 2015-05-30
Add a plugin to migrate metadata between sets of songs.

ZDBioHazard:
This songs menu plugin lets you easily copy the quodlibet-specific metadata (`~#rating`, `~#playcount`, `~#skipcount`, `~#added`, `~#laststarted`, `~#lastplayed`) of a selection of songs, and lets you paste it to another selection of songs.

When more than one track is selected, songs can be matched by disc and track number.

This is useful when you have an existing collection of music files encoded in one format that you want to replace with files encoded in another format, such as migrating an old MP3 collection to newly-ripped FLACs, without losing your ratngs and play counts.

The fields that are used are able to be selected at both copy and paste time.

### DIFF
--- a/quodlibet/po/POTFILES.in
+++ b/quodlibet/po/POTFILES.in
@@ -114,6 +114,7 @@ quodlibet/ext/songsmenu/importexport.py
 quodlibet/ext/songsmenu/k3b.py
 quodlibet/ext/songsmenu/lastfmsync.py
 quodlibet/ext/songsmenu/makesorttags.py
+quodlibet/ext/songsmenu/migratemetadata.py
 quodlibet/ext/songsmenu/playlist.py
 quodlibet/ext/songsmenu/refresh.py
 quodlibet/ext/songsmenu/replaygain.py
--- a/quodlibet/quodlibet/ext/songsmenu/migratemetadata.py
+++ b/quodlibet/quodlibet/ext/songsmenu/migratemetadata.py
@@ -23,21 +23,7 @@ class MetadataCopier(SongsMenuPlugin):
     PLUGIN_NAME = _("Migrate Metadata")
     PLUGIN_VERSION = "1.0"
     PLUGIN_ICON = Icons.EDIT_COPY
-    PLUGIN_DESC = _("Copies the quodlibet-specific metadata (~#rating, "
-                    "~#playcount, ~#skipcount, ~#added, ~#laststarted, "
-                    "~#lastplayed) of a selection of songs, and lets you "
-                    "paste it to another selection of songs."
-                    "\n\n"
-                    "When more than one track is selected, songs can be "
-                    "matched by disc and track number."
-                    "\n\n"
-                    "This is useful when you have an existing collection of "
-                    "music files encoded in one format that you want to "
-                    "replace with files encoded in another format, such as "
-                    "migrating an old MP3 collection to newly-ripped FLACs."
-                    "\n\n"
-                    "The fields that are used are able to be selected at "
-                    "both copy and paste time.")
+    PLUGIN_DESC = _("Copies the quodlibet-specific metadata between songs.")
 
     def plugin_songs(self, songs):
         global songinfo
--- a/quodlibet/quodlibet/ext/songsmenu/migratemetadata.py
+++ b/quodlibet/quodlibet/ext/songsmenu/migratemetadata.py
@@ -1,0 +1,157 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013-2015 Ryan "ZDBioHazard" Turner <zdbiohazard2@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


ZDBioHazard:
> Perhaps this list should be in the plugin's config, especially if it includes non-standard tags (`~#energy`)

Oh, whoops, I thought I picked this commit from my master branch, not my energy branch.

> Maybe it should go in the [documentation](http://quodlibet.readthedocs.org/en/latest/guide/overview.html#plugins).

Makes sense, though I'm personally a fan of having the documentation in-app when possible. 

I don't see a section in the documentation about individual plugins. Should I create one in this PR? Probably should create a separate issue for that project.

> FYI these are (almost) all available for free by using `quodlibet.util.tags.readable()`

Awesome, this changes everything. The only reason the checkboxes aren't auto-generated based on `quodlibet.formats._audio.MIGRATE` is because I didn't have a way to get translated tag names for the UI.

The only thing that bothers me about `readable()` is the output isn't properly capitalized for display. Since I'm not very i18n-savvy, would calling `readable(...).capitalize()` violate any i18n rules?

I'll start work on a new version that fixes everything else mentioned, and auto-generates the checkboxes based on `quodlibet.formats._audio.MIGRATE`. I'm not going to have time for a few days though.

> `is True` seems unnecessary here, no?

[PEP-20](https://www.python.org/dev/peps/pep-0020/): "Explicit is better than implicit." :wink:

> Are the casts to `int` not done by the formatter anyway?

```python-traceback
>>> print("%d" % "1")
TypeError: %d format: a number is required, not str
```

declension:
On 04/06/15 03:32, Ryan Turner wrote:
>
>     Maybe it should go in the documentation
>     <http://quodlibet.readthedocs.org/en/latest/guide/overview.html#plugins>.
>
> Makes sense, though I'm personally a fan of having the documentation 
> in-app when possible.
>
> I don't see a section in the documentation about individual plugins. 
> Should I create one in this PR? Probably should create a separate 
> issue for that project.
>

Yep, good point (and why I made a note to self). We used to have some 
docs, but when we moved plugins to the core we should have put some more 
effort into this I guess. Perhaps I'll rustle something up shortly.

The main reason for keeping it a bit shorter is for UI consistency - we 
keep promising / threatening to re-design that plugins management UI 
(we've made a few tweaks along the way but nothing major).

> Awesome, this changes everything. The only reason the checkboxes 
> aren't auto-generated based on |quodlibet.formats._audio.MIGRATE| is 
> because I didn't have a way to get translated tag names for the UI.
>

Cool


> The only thing that bothers me about |readable()| is the output isn't 
> properly capitalized for display. Since I'm not very i18n-savvy, would 
> calling |readable(...).capitalize()| violate any i18n rules?
>
Me neither really. In general probably, yes, completely (Arabic? 
Mandarin?). There's https://pypi.python.org/pypi/PyICU it seems, but we 
don't use that. FWIW the [Python String 
docs(https://docs.python.org/2/library/stdtypes.html#str.capitalize) 
hint it's safe (_this method is locale-dependent_), but who knows.

I think it's probably safe to do as we've undoubtedly got similar 
problems elsewhere (safer of course is to not process it though).

> I'll start work on a new version that fixes everything else mentioned, 
> and auto-generates the checkboxes based on 
> |quodlibet.formats._audio.MIGRATE|. I'm not going to have time for a 
> few days though.
>
Great, thanks.

>     |is True| seems unnecessary here, no?
>
> PEP-20 <https://www.python.org/dev/peps/pep-0020/>: "Explicit is 
> better than implicit." :wink:
>
OK, personal taste I guess :)



>     Are the casts to |int| not done by the formatter anyway?
>
> >>> print("%d" % "1")
> TypeError:%d format: a number is required, not str

Cool, good to know. Wishful thinking on my part clearly, though it's 
probably a good thing.

zsau:
Wouldn't it make more sense to export the metadata to a file, like `importexport.py` does? That would make it easier to support cases like library migration (#1747).

ZDBioHazard:
I rewrote this plugin to generate the tag list from `quodlibet.formats._audio.MIGRATE` and made the suggested changes to the description and the track count text. Sorry to leave you all hanging this long. I forgot this PR was still open.

It looks like the Travis build failed due to a temporary X11 error in the Windows test. I'm not sure how to retest without pushing new code.

> Wouldn't it make more sense to export the metadata to a file, like importexport.py does? That would make it easier to support cases like library migration (#1747).

This plugin doesn't match tracks against anything but selection index or track number. Migrating data between moved files isn't really the purpose of this plugin. Maybe I should rename this plugin to "copy/paste metadata" or something?

About import/export to an intermediate file; I figured it would be easier for everyone involved to avoid dealing with writing, parsing, storing and supporting external files, and just do a super simple one-click type of thing.

declension:
I've retriggered, green now.

lazka:
added some fixes in bcee734bb06becee5f8b0dbd2597ece36e439e59 after merging

PR #1643, opened 2015-06-29
Fix --show-window description in man page

SimonLarsen:


lazka:
Thanks. These files are generated and they shouldn't be changed manually. The source is here: https://github.com/quodlibet/quodlibet/tree/master/quodlibet/docs/guide/commands ; the make file generates the man pages

SimonLarsen:
Oh yeah, of course. Changed the source file and ran make. Should be okay now.

lazka:
Thanks

lazka:
Added a readme now: 0085ff4e4ae359d7a613

Issue #157, opened 2015-03-14
Browser and Library plugins 


lazka:
Original [issue 157](https://code.google.com/p/quodlibet/issues/detail?id=157) created by steven.strobe.cc on 2009-03-20T17:24:11.000Z:

Quod Libet should become even more extensible by allowing browsers and
libraries to be loaded as plugins. Their tight integration with other code
may require a model for loading plugins that differs from the
enabled/disabled/destroyed framework of existing plugins, but ideally a way
can be found to preserve that logic across all plugins.

PR #1657, opened 2015-08-01
Implement GstAppSrcPlugins, so plugins can play arbitrary audio to QL

akdor1154:
I am currently implementing Spotify support in Quod Libet. This plugin interface makes this support a lot neater and means the internals of QL don't have to know anything about Spotify (as, quite rightly, they shouldn't!)

This plugin interface should be by no means specific to Spotify; in fact, it should enable the creation of plugins that play arbitrary data through Quod Libet.

The Gstreamer player selects which source to use based on the protocol of the uri of the song to be played. So, my own plugin will handle uris like 'spotify:track:blahblah', but you could also feasibly handle uris like 'googleplay://blahblah', or 'generatesinewave:440', or anything you like, really. :)


nemethf:
But then wouldn't it make sense to update quodlibet/ext/songsmenu/fingerprint/analyze.py and quodlibet/ext/songsmenu/replaygain.py as well? 

Calculating fringerprints and replaygain for local files with special URIs  would be probably useful.  Or should these calculations be disabled for any GstAppSrcPlugins plugin?  Or maybe the plugin writer could somehow indicate whether a music file is steamed or not.

akdor1154:
It would, but I believe that's beyond the scope of this -- currently, QL will offer to run the Fingerprint or Replaygain plugins to an internet radio track, which should probably be excluded from those plugins in a similar fashion.

lazka:
Otherwise looks good. Sorry for the delay.

akdor1154:
Cool, I'll have a look at the ref cycle thing.
Oooh, is_setup/play_after_setup is a bit yucky, yes. It either requires some clear documentation on how plugin authors need to use them, or a big rethink on how to set up the appsrc (and I've already attempted the latter and failed). To illustrate how I'm using them now:
```python
class SpotifySource(AppSrcPlugin, spotify.sink.Sink):
    PLUGIN_ID = "spotify-appsrc"
    PLUGIN_NAME = _("Spotify Play Component")
    PLUGIN_DESC = _("Handles playing files with a spotify: protocol")

    uri_protocol = 'spotify'

    def source_setup(self, playbin, appsrc):
        super(SpotifySource, self).source_setup(playbin, appsrc)
        self._appsrc.set_caps(self._caps)
        self._appsrc.set_stream_type(GstApp.AppStreamType.SEEKABLE)
        self._appsrc.set_property('format', Gst.Format.TIME)
        self._appsrc.set_property('max-bytes', 100000)
        self._appsrc.set_property('block', False)

        self._is_setup = True
        if self._play_after_setup is not None:
            self._play_after_setup()
            self._play_after_setup = None
```

In order to leave the gstplayer largely untouched, source_setup only gets called when set_uri is called. This starts playback (am I correct in that, for vanilla quodlibet, just setting the uri on gst's playbin is all that's required to start playback?). Unfortunately if we need to do any other setup for playback to work, as we do for the appsrc, we need to postpone the actual playback until this is done. I don't think I can handle this in the base class, otherwise plugin authors can't do any setup before their playback starts (or they have to call super() AFTER their setup code, which leads to other problems).
Ideas? I'm tending towards leaving a small note for plugin authors to just put the snippet 
```python
        self._is_setup = True
        if self._play_after_setup is not None:
            self._play_after_setup()
            self._play_after_setup = None
```
after their setup, but if there is a cleaner solution that would be great.

akdor1154:
http://stackoverflow.com/a/5790334 has a decent solution, would that be a bit cleaner? Obviously with source_setup instead of __init__.
```python
class Base:
    def __init__(self, args, *additional_args):
        # perform some pre-initialization
        # ...

        # perform any futher initialization needed by derived classes
        self.subclass_setup(*additional_args)

        # now call method "calculate"
        self.calculate()

    def subclass_setup(self, *args):
        pass

class Derived(Base):
    def __init__(self, args, additional_attr):
        super().__init__(args, additional_attr)

    def subclass_setup(self, additional_attr):
        # do some work and create new instance attributes
        # ...
        self.additional_attr = additional_attr
```

lazka:
Can you include the spotify plugin here once that starts to work? It's a bit 
hard to see how things fit together without a real implementation.

> This starts playback (am I correct in that, for vanilla quodlibet, just setting the uri on gst's playbin is all that's required to start playback?).

The internal elements of the playbin get setup on the READY->PAUSED playbin state change afaik. In gapless mode setting the new uri is enough as playbin handles everything itself in that case. Or do you mean something else?

akdor1154:
SpotifyPlugin:
https://github.com/akdor1154/quodlibet/blob/d4ce0cdf4ea1a157f906d64ef2738d4106147066/quodlibet/quodlibet/ext/gstappsrc/spotifysource.py

It's still a little on the hacked-together side, and this particular plugin is useless on its own (e.g. I'm also working on a Browser, a SpotifyFile(RemoteFile) class, and some non-hacky way of glueing all this together), but the plugin structure itself should be pretty clear. Music playback and control works perfectly well.

declension:
Cool. Note there's a _very_ long-standing Issue about Browser (and Library) plugins in #157.

lazka:
What's the status here?

akdor1154:
uni caught up with me, sorry. If I can get this all together in three weeks before my last semester I might  try and push it again. I'll at least rebase it on current git for now.

lazka:
ok, just checking.., there are no deadlines here :)

lazka:
Maybe of interest: https://jodal.no/2016/02/18/guide-to-poor-api-management/

lazka:
If there is anything new feel free to reopen/make a new pull request/open an issue.

PR #1666, opened 2015-08-10
How do I add custom tags to header?

MikeiLL:
I think this might clarify the relationship between custom tags and custom headers a bit.


declension:
Thanks. Included a version of this as part of revision 7041c6ff83011d369b5cca052c4e0a20bb3ec7f7

Issue #1628, opened 2015-05-30
Revamp the lyrics plugin


CreamyCookie:
There are two tabs:
* **File** - editor area that displays the lyrics that are stored in the tags of the currently playing song
* **Online** - browser area that shows the [LyricWiki](http://lyrics.wikia.com/Lyrics_Wiki) page or search results for the song

Whenever a song gets played, the plugin checks if the song has lyrics tags.
* If yes, the **File** tab is selected.
* If not, the **Online** tab (containing the browser area) is selected and the plugin checks whether lyrics exist in the LyricWiki.
  * If they do, the lyrics page is displayed.
  * If they don't, a search (the search URL can be configured) is started with `<author> - <title>`

For this enhancement, bug #280 needs to be fixed first. Implementing this enhancement issue fixes #1485.

dannn-o:
Another vote for the embedded lyrics to be displayed.  Unlike what CreamyCookie stated, my experience is that the View Lyrics plugin does not display embedded lyrics (i.e. those stored in the MP3 file's ID3 tag).

declension:
> Unlike what CreamyCookie stated, my experience is that the View Lyrics plugin does not display embedded lyrics (i.e. those stored in the MP3 file's ID3 tag).

View lyrics definitely uses embedded (or file-based) lyrics. There may be errors, of course. If this is happening please try running with --debug and filing an error.

declension:
This issue is several things really
 * Inline editing of lyrics (within viewlyrics I guess)
 * Merging the two plugins

Note also:
 * #1485 is now fixed.
 * There is a separate UI piece around [viewlyrics becoming a sidebar (#2553)](#2553). This at least brings the two plugins closer physically and conceptually.

dannn-o:
With the recent updates on this thread, I've taken another look at Quod Libet and the lyrics plug-in.  To ensure I'm at the latest (stable) version with Linux Mint 17.3, I added the stable PPA (ppa:lazka/ppa) and updated, but was already at the latest version available (3.8.1).  It seems that toggling the "View Lyrics" option within the Plugins dialog does not really do anything...no lyrics at all appear beneath the song list in the main window.  The "Lyrics Window" plugin does appear to open and display lyrics from some sort of online source (Fandango?) if a match is found, but my preference, again, is to view embedded lyrics (which are present in my entire library of audio files).

[...]

Issue #1485, opened 2015-03-15
"Edit lyrics" button doesn't work (it opens a website instead)


lazka:
Original [issue 1485](https://code.google.com/p/quodlibet/issues/detail?id=1485) created by joschuaga@yahoo.de on 2014-10-31T17:17:25.000Z:

<b>\- What did you try to do?</b>
1. Click Information / Lyrics on a song.
2. Click on &quot;Edit lyrics&quot;

<b>\- What did you expect to happen?</b>
A way to edit the lyrics.

<b>\- What did happen instead?</b>
A website opens. http://lyrics.wikia.com/{artist_name}

<b>\- Which version of Quod Libet?</b>
3.2.2.

<b>\- Which operating system including version (Ubuntu 14.04, Win7, Debian</b>
<b>sid)?</b>
Ubuntu 14.04 / 64bit

declension:
Fixed in https://github.com/quodlibet/quodlibet/commit/949444c516d5a988461d26609fa0c19dd11127ca (renamed the button for clarity).

The idea was you can edit _Lyricwiki's_ version of the lyrics online using that button, but I'm sure most people  use this to _view_ them online.

Editing lyrics happens in-place.

PR #1709, opened 2015-10-14
Add a new LyricsWindow events plugin

### LINKED ISSUE #1629: The lyrics plugin window should get opened if it was open in the last Quod Libet session
At the moment, every time I start Quod Libet, I have to reopen the lyrics window.

CreamyCookie:
This plugin (an EventPlugin) is based on the [LyricWiki plugin](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/ext/songsmenu/lyricwiki.py) (a SongsMenuPlugin), which it aims to replace.

Compared to it, it has the following advantages:

* The window gets opened when Quod Libet is launched. Closes  #1629.
* The user can configure the zoom level of the web view.
* Window position and size are (re)stored.

Additionally I might implement #1485 and parts or the entirety of #1628 in a later pull request, which would make the plugin use a search engine when there are no lyrics in LyricWikia. This is also the reason why it was renamed to **lyricswindow.py**. 

This is how it looks with my current configuration (zoom = 1.4):

lyricswindow

---

One thing I'd like to do, but don't really know how, is to set `LyricsWebViewWindow.set_transient_for(quod_libet_main_window)` ([doc](http://www.pygtk.org/pygtk2reference/class-gtkwindow.html#method-gtkwindow--set-transient-for)), so that when bringing Quod Libet to the front, the same would happen to the lyrics window. How do I get main window in the plugin?

Regarding the Copyright notice: Can I leave it the way it is?


lazka:
> This plugin (an EventPlugin) is based on the LyricWiki plugin (a SongsMenuPlugin), which it aims to replace.

I guess we should allow merging EventPlugins and SongsMenuPlugin so we can have both enabled() and the song menu actions.

> Compared to it, it has the following advantages:
>
>    The window gets opened when Quod Libet is launched. Closes #1629.

OK

>    The user can configure the zoom level of the web view.

Why not hardcode a good value if it is too small?

>    Window position and size are (re)stored.

There is qltk.window.PersistentWindowMixin for that as well. But it's hardcoded to same to "memory" atm, and we don't want plugin state there.

> Additionally I might implement #1485 and parts or the entirety of #1628 in a later pull request, which would make the plugin use a search engine when there are no lyrics in LyricWikia. This is also the reason why it was renamed to lyricswindow.py.

Sounds good.

> This is how it looks with my current configuration (zoom = 1.4):

> One thing I'd like to do, but don't really know how, is to set LyricsWebViewWindow.set_transient_for(quod_libet_main_window) (doc), so that when bringing Quod Libet to the front, the same would happen to the lyrics window. How do I get main window in the plugin?

```
from quodlibet import app
app.window
```
SongMenuPlugins have a ``plugin_window`` attribute set to the window which owns the menu.

> Regarding the Copyright notice: Can I leave it the way it is?

You should add your name

lazka:
> One thing I'd like to do, but don't really know how, is to set LyricsWebViewWindow.set_transient_for(quod_libet_main_window) (doc), so that when bringing Quod Libet to the front, the same would happen to the lyrics window. How do I get main window in the plugin?

Btw. you linked to gtk+2 docs. Here are the gtk+3 ones: https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Window.html#Gtk.Window.set_transient_for

CreamyCookie:
Okay, I updated the code according to your feedback. :)

> I guess we should allow merging EventPlugins and SongsMenuPlugin so we can have both enabled() and the song menu actions.

Yeah, that would be useful. In regards to this I changed the `plugin_on_song_started` event, so it dosen'- reopen the window if it's closed, when a new song starts.

> There is qltk.window.PersistentWindowMixin for that as well. But it's hardcoded to same to "memory" atm, and we don't want plugin state there.

What would need to change in `PersistentWindowMixin` to make it usable in this case?

> Btw. you linked to gtk+2 docs. Here are the gtk+3 ones: https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/Window.html#Gtk.Window.set_transient_for

Ah nice. Thank you! :+1:

lazka:
OK, thanks and sorry for the delay. There are still a few no-gos. The Window can't be reopened easily and having two plugins doing the same thing but in slightly different way is confusins (plus the code duplication).

I'll look into fixing PersistentWindowMixin and to allow making a mixed songmenu/event plugins before continuing here.

CreamyCookie:
@lazka Okay, cool. I will delete the original plugin file in my branch. (I first wanted to get feedback before doing that.)

I might add the search function, which I mentioned earlier, in the meantime.

> The Window can't be reopened easily

I guess this would be possible when PersistentWindowMixin is fixed, right?

CreamyCookie:
@lazka I have added the search functionality and refactored the code a bit. (lyricswindow and another plugin were using the same classes.) That took way too much time, haha.. But hey, I learned the basics of git.

Please feel free to code review the heck out of this. :smile:

quodlibet/quodlibet/plugins/__init__.py
```diff
@@ -500,3 +500,69 @@ def ConfigCheckButton(cls, label, name, default=False):
             cls.config_set(name, default)
         return ConfigCheckButton(label, PM.CONFIG_SECTION,
                                  option, populate=True)
+
+
```
declension: Unless I've missed the point, this seems like a lot of OOP to do what `PluginConfigMixin` is already doing (`config_get_bool`, `config_set` etc)?

quodlibet/quodlibet/ext/events/lyricswindow.py
```diff
@@ -0,0 +1,368 @@
...
+        if self._window is None:
+            self._open_webview_window()
+
+    def disabled(self):
+        if self._window is not None:
+            self._window.close()
+
+    @cached_property
+    def Conf(self):
+        return get_config(self.PLUGIN_ID)
```
declension: This now seems very similar to what `PluginConfigMixin` does, essentially (but less configurable for the config section actually)...

CreamyCookie:
Hmm, would be cool, if GitHub allowed me to directly respond on a certain "line note".

> Unless I've missed the point, this seems like a lot of OOP to do what PluginConfigMixin is already doing (config_get_bool, config_set etc)?

I guess that's true. There are a few advantages to the `ConfProp` classes though:

* You can just access and set them like normal attributes (e.g. `conf.zoom_level = 3`).
* You don't need to know the exact type. (You don't need to care whether `zoom_level` is a float or int.) 
* The code seems more readable, because it isn't littered with `config_get`'s.
* It allows your IDE to provide code completion.


Note that these classes weren't my idea. They came from the [Animated On-Screen Display plugin](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/ext/events/animosd/config.py). Hmm.. I think there's another advantage shown here, regarding `ColorConfProp`: You can easily extend the configuration classes.

I don't know if that's worth it though. So, if you give me the sign, I'll use `config_get_bool` etc. everywhere. I'd also add `config_get_float` then, if that's okay.

> This now seems very similar to what PluginConfigMixin does, essentially (but less configurable for the config section actually)...

What do you mean about being less configurable?

declension:
Hi. Yes it's annoying about line comment threads.

Anyway I see the benefits of the more OOP approach, to an extent (and looked again at Animated OSD, which is probably due a bit of refactoring anyway)... but with over 70 plugins in the codebase managing the diversity amongst them is quite a task sometimes, so I'm keen to get _more_ consistency first, and in the interests of getting the PR merged, keep _that_ work separate from extended the common config code significantly. So for now I'd be happiest if this used the `config_get_bool` etc (yes though please add related `float` / `int` ones if they're necessary)

Afterwards, I'm happy to revisit the, and fix, de-duplicate and improve the config system _across_ plugins as a separate commit / PR / Github Issue.

> Could you explain what you mean by the solution being less configurable?

Oh... it isn't that important, but it's down to the fallback / legacy support mechanism used (via `CONFIG_SECTION` in that class)

@lazka do you have any further comments?

lazka:
> @lazka do you have any further comments?

 re config? I don't  care much about how it's done. The animosd way is using descriptors because it used to use a special config object before and at the last refactor I just kept the interface as is and connected it to the config.

I've recently moved some plugins to PluginConfig (https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/ext/events/trayicon/util.py#L18) which acts as a proxy to the real config. I prefer it over the mixin but ymmv.

declension:
@lazka - meant about the PR entirely really (LGTM other than above)

Re: config - yes, I see now, that's quite cool too. I guess _any_ of these (three?) existing methods for config is probably good I guess, would be good to settle on one or two and migrate slowly, but outside of the scope for this here..

lazka:
I don't like losing the song menu item for this plugin. But I don't feel that strongly about it.. so feel free to merge.

Tamriel:
I would love to see this merged!
A lyrics window with custom zoom level is exactly what I wished for.

CreamyCookie:
> so feel free to merge.

Sorry for responding so late. How do I merge? I don't have write access.

---

For reference, the following are the tasks that should probably be done in other pull requests, because they would increase the complexity of this one even more.

* fix PersistentWindowMixin so that it can be used for plugins, instead of having to manually set the size / position of windows
* creating a class that inherits from both EventPlugins and SongsMenuPlugin, so the window can be easily reopened via the context menu of a song
* possibly decide for one method of storing configurations in plugins (ConfProp, PluginConfigMixin, and PluginConfig)

lazka:
> Sorry for responding so late. How do I merge? I don't have write access.

This was directed at @declension, only project members can merge.

> * fix PersistentWindowMixin so that it can be used for plugins, instead of having to manually (re)store the size / position of windows
* creating a class that inherits from both EventPlugins and SongsMenuPlugin, so the window can be easily (re)opened via the context menu of a song
* possibly decide for one method of storing configurations in plugins (ConfProp, PluginConfigMixin, and PluginConfig)
>
> Should I add these as bug reports?

Yes please, sounds good.

declension:
Yes - I'm largely happy (ongoing config decisions aside), but the branch is conflicting now, so it needs fixing first...

CreamyCookie:
@declension What needs fixing?

declension:
The branch has merge conflicts so it needs rebasing / merging with `master`, that's all

CreamyCookie:
Yay, I did it (my first git rebase as far as I remember). I squashed a few commits via `fixup` in the interactive rebase, so the commits are more clear.

CreamyCookie:
Yay, I did it (my first git rebase as far as I remember). I squashed a few commits via `fixup` in the interactive rebase, so the commits are more clear. It should be ready to merge into master now.

CreamyCookie:
Yay, I did it! :)

I squashed a few commits via `fixup` in the interactive rebase, so the commits are more clear.  Furthermore, the WebView shouldn't make any sounds on its own anymore. (There were some annoying audio ads that started autoplaying, which were driving me crazy, because I didn't know that they were coming from the lyrics window.)

@declension Is this alright?

declension:
Thanks

PR #1717, opened 2015-10-18
Proper language codes

piotrdrag:
For Czech, "cs" should be used instead of "cs_CZ", and Galician should use "gl" instead of "gl_ES".


lazka:
Is there a list somewhere for what should be used as language codes?

piotrdrag:
I don't think so, but GTK+ would be a good reference point: https://git.gnome.org/browse/gtk+/tree/po

declension:
I'd say the authoritative _language_ source should ISO-639-1, in particular [as on the GNU site](https://www.gnu.org/software/gettext/manual/html_node/Usual-Language-Codes.html#Usual-Language-Codes) with ISO 3166 _country codes_ as [they list here](https://www.gnu.org/software/gettext/manual/html_node/Country-Codes.html#Country-Codes).

fitojb:
Unfortunately, this change left behind a Galician file with an unofficial spelling system. Due to this PR being so old, I’ll try to merge back the file into the current file on Weblate.

PR #1723, opened 2015-10-25
Add new events plugin - Synchronized Lyrics

elfalem:
This plugin reads .lrc files and shows timed lyrics.

### DIFF
--- a/quodlibet/quodlibet/ext/events/synchronizedlyrics.py
+++ b/quodlibet/quodlibet/ext/events/synchronizedlyrics.py
@@ -1,0 +1,305 @@
+# -*- coding: utf-8 -*-
+# Synchronized Lyrics: a Quod Libet plugin for showing synchronized lyrics.
+# Copyright (C) 2015 elfalem
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+
+
+"""Provides `Synchronized Lyrics` plugin for showing synchronized lyrics."""
+
+import os
+
+from datetime import datetime
+
+from gi.repository import Gtk, Gdk, GLib
+
+from quodlibet import app
+from quodlibet import qltk
+from quodlibet import config
+
+from quodlibet.plugins.events import EventPlugin
+
+
+class SynchronizedLyrics(EventPlugin):
+
+    PLUGIN_ID = 'SynchronizedLyrics'
+    PLUGIN_NAME = _('Synchronized Lyrics')
+    PLUGIN_DESC = _('Shows synchronized lyrics from .lrc file with same name \
+as the track.')
+
+    SYNC_PERIOD = 3000
+
+    _defaultBgColor = '#343428282C2C'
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/ext/events/synchronizedlyrics.py
```diff
@@ -0,0 +1,305 @@
...
+        s.set_numeric(True)
+        s.set_text(str(self.getFontSize()))
+        t.attach(s, 1, 2, 4, 5)
+        s.connect('value-changed', self.setFontSize)
+
+        vb.pack_start(t, False, False, 0)
+        return vb
+
+    def getTextColor(self):
+        v = config.get("plugins", self.PLUGIN_ID + "_txt",
```
declension: There's actually a class to wrap all this (consistently). 
See [PluginConfigMixin](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/plugins/__init__.py#L456).

quodlibet/quodlibet/ext/events/synchronizedlyrics.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def getBgColor(self):
+        v = config.get("plugins", self.PLUGIN_ID + "_bg", self._defaultBgColor)
+        return v[:3] + v[5:7] + v[9:11]
+
+    def getFontSize(self):
+        return int(config.get("plugins", self.PLUGIN_ID + "_fsize",
+                              self._defaultFontSize))
+
+    def setTextColor(self, button):
```
declension: ..and for set etc.

quodlibet/quodlibet/ext/events/synchronizedlyrics.py
```diff
@@ -0,0 +1,305 @@
...
+            #parse lyricsLine
+            if not lyricsLine[1].isdigit():
+                continue
+            closeBracket = lyricsLine.find("]")
+            timeObject = datetime.strptime(lyricsLine[1:closeBracket],
+                                           '%M:%S.%f')
+            timeStamp = (timeObject.minute * 60000 + timeObject.second * 1000
+                         + timeObject.microsecond / 1000)
+            words = lyricsLine[closeBracket + 1:]
+            if words == "":
```
declension: Is `if not words`, or better, `if words` with reversed logic sufficient here? 

If not, depending on the difference between various _falsey_ things (empty string, `None`, `False` etc) is risky.

declension:
Thanks. It looks good to me, other than a few tweaks. 

Can we ensure adherence to [PEP 8 naming](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) - notably underscores for method names, not camel case.

elfalem:
Thanks for the comments. I've followed PEP 8 conventions and used PluginConfigMixin.

I'm not sure I understand the comment about depending on falsey things. That area of the code will act on strings such as `[03:28.87][03:09.36][01:37.52][00:48.97]Some words` broken up by the brackets. The first three slices would match the empty string condition while the fourth one would fail.

declension:
Great - thanks for the plugin.

PR #1776, opened 2016-01-03
Save and read extensive statistics in id3 and xiph tags.

HoverHell:
The patch still lives!

Somehow, this still wasn't in the code (and the patches even applied automatically).


lazka:
This is a bit of "too many changes in one", "no progress in a year", "missing tests".. closing.

Issue #1339, opened 2015-03-15
trash.py trows exeption due to inproper rights-handling, QL stalls and has to be killed manually via  SIGKILL


lazka:
Original [issue 1339](https://code.google.com/p/quodlibet/issues/detail?id=1339) created by marcus.rottensteiner on 2014-02-27T20:23:32.000Z:

<b>Please describe the problem thoroughly and accurately. If possible paste</b>
<b>the contents of the ~/.quodlibet Dump or MiniDump file associated with the</b>
<b>problem; otherwise, you must at a minimum report the version of Quod Libet</b>
<b>and the operating system you are using.</b>

<b>- What did you try to do?</b>

I wanted to delete a mp3 file.

<b>- What did you expect to happen?</b>

That file is deleted 
OR
If not possible, because of any reason, I would expect the operation to be canceled and that a info dialog is shown + after ACK qoodlibet continues with normal operation.

OLYMP: Best would be if quodlibet asks me if (not yet defined by user) where the Trash-Folder should reside, which would also avoid probably wrong guessing of the application... 

<b>- What did happen instead?</b>
trash.py throws exeption due to inproper rights-handling and quodlibet has to be killed with -9 because no respond to any buttons nor closing the application window
was possible.

get_fd_trash_dir(path) delivers Trash-Path based on mount-point root-directory
Rights of this directory are not checked, so Trash can not be created.

=========================
    # --&gt; cat MiniDump_20140227_172244.txt 
    === SYSTEM INFORMATION:
    Quod Libet 3.0.2
    Mutagen 1.22
    Python 2.7.6 (default, Jan 12 2014, 08:16:39) 
    [GCC 4.8.2] linux2
    Platform Linux-3.13.0-12-generic-x86_64-with-Ubuntu-14.04-trusty
    
    === STACK TRACE
    Traceback (most recent call last):
    
      File &quot;/usr/lib/python2.7/dist-packages/quodlibet/qltk/songsmenu.py&quot;, line 321, in __delete
        removed = dict.fromkeys(d.run())
    
      File &quot;/usr/lib/python2.7/dist-packages/quodlibet/qltk/delete.py&quot;, line 102, in run
        trash.trash(filename)
    
      File &quot;/usr/lib/python2.7/dist-packages/quodlibet/util/trash.py&quot;, line 131, in trash
        trash_free_desktop(path)
    
[...]

PR #1782, opened 2016-01-13
Issue1339: fix cross-device trashing

achadwick:
quodlibet/quodlibet#1339 was annoying me in 3.5.1, so I threw together a quick fix. Implementation is now a bit closer to <http://standards.freedesktop.org/trash-spec/trashspec-latest.html>. This handles my silly cross-volume music library fine even for broken or unwriteable trash structures, provided the fallback is still OK.

### DIFF
--- a/quodlibet/quodlibet/util/trash.py
+++ b/quodlibet/quodlibet/util/trash.py
@@ -28,24 +28,54 @@ def is_sticky(path):
     return bool(os.stat(path).st_mode & stat.S_ISVTX)
 
 
-def get_fd_trash_dir(path):
-    """Returns the right trash directory for the given path."""
+def _get_fd_trash_dirs(path):
+    """Returns verified trash folders for the given path.
+
+    Returns (rootdir, filesdir, infodir), or raises a TrashError if a
+    valid trash folder structure could not be found. The returned trash
+    folders are not guaranteed to be on the same volume as the original
+    path: a fallback may be returned instead.
+
+    Returned paths are absolute. This method may create partial or
+    complete trash directory structures as part of its search.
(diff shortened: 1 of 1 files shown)


lazka:
Thanks for looking into it.

What about using https://lazka.github.io/pgi-docs/#Gio-2.0/interfaces/File.html#Gio.File.trash instead?

achadwick:
Is QuodLibet moving to Gio everywhere? Then that would be a better API. `trash_free_desktop()` doesn't accept GFiles as parameters though, and I've not looked further.

Conversely, and may not be an issue: fallbacks are allowed by the spec, and are nice to have. However `g_file_trash()` isn't described as doing that. Instead, you'll have to catch a specific Gio.IOErrorEnum case via a GError and handle it with an additional direct-deletion dialog (like Nautilus).

lazka:
> Is QuodLibet moving to Gio everywhere? Then that would be a better API. trash_free_desktop() doesn't accept GFiles as parameters though, and I've not looked further.

No plans atm, but I can't think of a reason why we couldn't use it in this case (except pygobject+python3 only supporting utf-8 paths).

> Conversely, and may not be an issue: fallbacks are allowed by the spec, and are nice to have. However g_file_trash() isn't described as doing that. Instead, you'll have to catch a specific Gio.IOErrorEnum case via a GError and handle it with an additional direct-deletion dialog (like Nautilus).

Yeah, deleting in case of an error wouldn't be a good idea :) We currently don't do fallbacks so that wouldn't be a regression at least.

lazka:
Merged for now, thanks. We should look into using Gio sometime...

PR #1783, opened 2016-01-13
In song lists augment sort tag for tag patterns with conditional tag …

pfps:
…pattern that selects sort version of tag if present.

All that this does is replace any <tag> that has a sort version with <sorttag|<sorttag>|<tag>> when building the sort tag from a tag pattern.

### DIFF
--- a/quodlibet/quodlibet/qltk/songlist.py
+++ b/quodlibet/quodlibet/qltk/songlist.py
@@ -160,7 +160,7 @@ def get_sort_tag(tag):
             tag = tag.replace("<%s>" % key, "<%s>" % value)
         for key, value in TAG_TO_SORT.iteritems():
             tag = tag.replace("<%s>" % key,
-                              "<{1}|<{1}>|<{0}>>".format(key, value))
+                               "<{1}|<{1}>|<{0}>>".format(key, value))
         tag = Pattern(tag).format
     else:
         tags = util.tagsplit(tag)
--- a/quodlibet/quodlibet/qltk/songlist.py
+++ b/quodlibet/quodlibet/qltk/songlist.py
@@ -159,8 +159,8 @@ def get_sort_tag(tag):
         for key, value in replace_order.iteritems():
             tag = tag.replace("<%s>" % key, "<%s>" % value)
         for key, value in TAG_TO_SORT.iteritems():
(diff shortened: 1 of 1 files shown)


lazka:
Looks good. Can you fix the travis tests? https://travis-ci.org/quodlibet/quodlibet/jobs/102230304#L1067

pfps:
It appears that the problem was a line longer than 79 characters.  I
reformatted the line, and uploaded the change.

peter

PS:  I wonder why there is a check for lines less than 80 characters.

lazka:
> PS: I wonder why there is a check for lines less than 80 characters.

We run pep8 and pyflakes, both with some errors and warnings disabled/ignored which we found too strict/annoying. It makes sure that the whole code base remains readable and has a consistent style. The 80 char limit is mostly there because pep8 has it by default and we haven't found the need to disabled/change it. If you prefer a different limit feel free to open a bug report about it.

You can invoke the tests yourself locally using "./setup.py quality". See https://quodlibet.readthedocs.org/en/latest/development/testing.html#testing

pfps:
I was just wondering about the line limit.  I'm not going to agitate to change it.

I tried to run the tests locally, but pep8 was not installed.  It took some digging to find it.  And then the output is strange

idefix quodlibet> ./setup.py quality
running quality
TPEP8 (1):                E                                                   0
================================================================================
ERROR: test_all (tests.quality.test_pep8.TPEP8)
--------------------------------------------------------------------------------
  File "/home/local/SoftwareDownloads/quodlibet-fork/quodlibet/tests/quality/test_pep8.py", line 86, in test_all
    raise Exception("\\n".join(errors))
Exception: 

[lots of blank lines]

TPyFlakes (1):            +                                                   1
0 test failure(s) and 1 test error(s), as detailed above.
idefix quodlibet> 

This is on an up-to-date Fedora distribution.


Anyway, not a severe problem, it just means that I have to push and look at the builds on github for this.

pfps:
Aah, you have to
dnf install python-pep8
not
pip install pep8
and then pep8 works OK on Fedora.

lazka:
Thanks

pfps:
Great!

By the way, how easy was it to set up the automatic checks?  Is there a cookbook for doing it?

PR #1796, opened 2016-01-21
use sort tags in paned browser; union tied tags in paned browser

pfps:
Code changes:

1/ list_separate and list_sort (new) in quodlibet/formats/_audio.py
   Only uses are in the paned browser

a) Union all results for tied tags instead of concatenating them in list_separate.  (This is
   a change, but it only affects the paned browser.)

b) Return display,sort pairs when a tag has an associated sort key with 
   different values.

2/ list_separate and format_list in quodlibet/pattern/_pattern.py
   Only uses are in the paned browser

Handle display,sort pairs appropriately.

3/ paned browser

Use sort value for sorting if present.

User-visible changes:

1/ Paned browser uses sort tags when present.
2/ Paned browser unions results for tied tags.

Caveats:

1/ If a tag value has different sort values (including no sort value) in
   different songs then in panes using the tag songs with this tag value will
   sort to one of the sort values at random and may have more than one entry. 
2/ If multiple tag values have the same sort value then in panes using the
   tag sorting of these tag values will be non-deterministic and there may
   be multiple entries for some of the tag values.  (This problem already
   existed.)  
3/ Some changes to documentation are indicated.

### DIFF
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -29,6 +29,7 @@ from quodlibet.compat import iteritems, string_types
 
 from ._image import ImageContainer
 
+from itertools import izip_longest
 
 MIGRATE = {"~#playcount", "~#laststarted", "~#lastplayed", "~#added",
            "~#skipcount", "~#rating", "~bookmark"}
@@ -532,25 +533,35 @@ class AudioFile(dict, ImageContainer):
             v = self.get(key)
             return [] if v is None else v.split("\n")
 
-    def list_separate(self, key, connector=" - "):
-        """Similar to list, but will return a list of all values
-        for tied tags instead of one comma separated string.
-
-        In case of tied tags the result will be unicode, otherwise
-        it returns the same as list()
+    def list_sort(self, key):
+        """Like list but return display,sort pairs when appropriate
+        and work on all tags
         """
+        display = decode_value(key,self.__call__(key))
+        display = display.split("\n") if display else []
+        sort = []
+        if key in TAG_TO_SORT:
+            sort = decode_value(TAG_TO_SORT[key],self.__call__(TAG_TO_SORT[key]))
+            # it would be better to use something that doesn't fall back
+            # to the key itself, but what?
+            sort = sort.split("\n") if sort else []
+        if len(display) > 0 and len(sort) > 0 and display != sort:
+            return [(d if d is not None and d != "" else s,
+                     s if s is not None and s != "" else d)
+                    for d, s in izip_longest(display, sort)]
+        else:
+            return display if len(display) > 0 else sort
 
-        if key[:1] == "~" and "~" in key[1:]:
-            vals = \
-                filter(None,
-                map(lambda x: isinstance(x, basestring) and x or str(x),
-                map(lambda x: (isinstance(x, float) and "%.2f" % x) or x,
-                (self(tag) for tag in util.tagsplit(key)))))
-            vals = (val.split("\n") for val in vals)
+    def list_separate(self, key):
+        """For tied tags return the list union of the display,sort values 
+           otherwise just do list_sort
+        """
+        if key[:1] == "~" and "~" in key[1:]: # tied tag
+            vals = [self.list_sort(tag) for tag in util.tagsplit(key)]
             r = [j for i in vals for j in i]
             return r
         else:
-            return self.list(key)
+            return self.list_sort(key)
 
     def list_unique(self, keys):
         """Returns a combined value of all values in keys; duplicate values
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -546,9 +546,8 @@ class AudioFile(dict, ImageContainer):
             # to the key itself, but what?
             sort = sort.split("\n") if sort else []
         if len(display) > 0 and len(sort) > 0 and display != sort:
-            return [(d if d is not None and d != "" else s,
-                     s if s is not None and s != "" else d)
-                    for d, s in izip_longest(display, sort)]
+            return [((d if d != "" else s, s) if s != "" else d)
+                    for d, s in izip_longest(display, sort, fillvalue="")]
         else:
             return display if len(display) > 0 else sort
 
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -533,7 +533,7 @@ class AudioFile(dict, ImageContainer):
             return [] if v is None else v.split("\n")
 
     def list_separate(self, key, connector=" - "):
-        """Similar to list, but will return a list of all combinations
+        """Similar to list, but will return a list of all values
         for tied tags instead of one comma separated string.
 
         In case of tied tags the result will be unicode, otherwise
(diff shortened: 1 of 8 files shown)


quodlibet/quodlibet/formats/_audio.py
```diff
@@ -532,26 +533,35 @@ def list(self, key):
             v = self.get(key)
             return [] if v is None else v.split("\n")
 
-    def list_separate(self, key, connector=" - "):
-        """Similar to list, but will return a list of all combinations
-        for tied tags instead of one comma separated string.
-
-        In case of tied tags the result will be unicode, otherwise
-        it returns the same as list()
+    def list_sort(self, key):
```
lazka: Please add a test for list_sort()

quodlibet/quodlibet/formats/_audio.py
```diff
@@ -532,26 +533,35 @@ def list(self, key):
...
+        if key in TAG_TO_SORT:
+            sort = decode_value(TAG_TO_SORT[key],
+                                self.__call__(TAG_TO_SORT[key]))
+            # it would be better to use something that doesn't fall back
+            # to the key itself, but what?
+            sort = sort.split("\n") if sort else []
+        if len(display) > 0 and len(sort) > 0 and display != sort:
+            return [((d if d != "" else s, s) if s != "" else d)
+                    for d, s in izip_longest(display, sort, fillvalue="")]
+        else:
```
lazka: I think it would be nicer to just always return a list of tuples even if there are no sort values. Also throw out sort values if there are too many.

quodlibet/quodlibet/formats/_audio.py
```diff
@@ -532,26 +533,35 @@ def list(self, key):
...
-        """Similar to list, but will return a list of all combinations
-        for tied tags instead of one comma separated string.
-
-        In case of tied tags the result will be unicode, otherwise
-        it returns the same as list()
+    def list_sort(self, key):
+        """Like list but return display,sort pairs when appropriate
+        and work on all tags
         """
+        display = decode_value(key, self.__call__(key))
```
lazka: note that ``self.__call__()`` is the same as ``self()``

pfps: OK, fixed (in two places).

peter

quodlibet/quodlibet/formats/_audio.py
pfps: I was trying to be somewhat efficient, i.e., if there are no useful sort
values, then don't go through the bother of creating both the display and sort
versions.  The consuming code already handled the mixed situation, because
that is what comes out of the pattern code, so there was no extra downstream
work (except a little bit at the very end) to do things this way.

I also wanted to distinguish between the situation where there is a sort value
and where there isn't so that a subsequent sort value (from a different song,
like where "The Beatles" is sometimes missing the "Beatles, The" sort value)
could be used in the paned browser.

As far as using extra sort values, I was thinking that extra information
should not be dropped.  If they are truely unneeded then they can be removed
from the data.

If you don't think that either of these are useful just let me know and I'll
make the changes.  (Both should be easy.)

peter

PS:  I noticed that using izip doesn't work in Python 3.   Should I include
the compatability magic to make later migration to Python 3 easier?

lazka: > I was trying to be somewhat efficient, i.e., if there are no useful sort
values, then don't go through the bother of creating both the display and sort
versions.  The consuming code already handled the mixed situation, because
that is what comes out of the pattern code, so there was no extra downstream
work (except a little bit at the very end) to do things this way.

I'd prefer the pattern to output tuple lists as well. Makes things more readable and less likely to make errors in handling the output. If we have a use case were only display versions are needed we can add another method just for that in the future.

> I also wanted to distinguish between the situation where there is a sort value
and where there isn't so that a subsequent sort value (from a different song,
like where "The Beatles" is sometimes missing the "Beatles, The" sort value)
could be used in the paned browser.

If this isn't needed in this case I'd ignore it and take empty strings as "no sort tag".

> As far as using extra sort values, I was thinking that extra information
should not be dropped.  If they are truely unneeded then they can be removed
from the data.

If we need it in the future we can easily change the code then. Maybe add the limitations regarding different value count in the docstring.

>If you don't think that either of these are useful just let me know and I'll
make the changes.  (Both should be easy.)

> peter
>
> PS:  I noticed that using izip doesn't work in Python 3.   Should I include
the compatability magic to make later migration to Python 3 easier?

Sure, why not.

quodlibet/quodlibet/formats/_audio.py
pfps: Done

quodlibet/quodlibet/formats/_audio.py
pfps: Done

lazka:
Any update on this?

pfps:
Is the ball in my court?  The last change I made was to always return pairs and ignore extra sort values, as requested.  This didn't change any observable behaviour.

I have been running with the modification, with no problems.  The effects of the changes are confined to the paned browser (unless add-ons can access this far down into quodlibet).  I think that the changes are ready to go.

If the change is made, there probably should be some updates to the documentation.  I'm willing to propose some updates.  (How would this be done, by the way?)

lazka:
I assumed it's not done since I saw some isinstance(tuple) checks. i'll have another look.

pfps:
Hmm.   Some of these are defensive, just in case a non-pair leaks through.  However, format_list might need some work.  I'll take a look.

pfps:
The instance(..., tuple) checks in models.py are defensive.  If a non-tuple
leaks through, these checks do the right thing with it.

Some checks in format_list are needed.  The output from a pattern execution
contains non-pairs for literal strings, which have to be handled correctly.
format_list did not turn these strings into pairs.  I have added a new patch
which does the conversion so that format_list always returns pairs.  This
simplied the logic in format_list, but does consume some extra resources (to
build up the pairs when they would not otherwise be needed).  I added an
explicit test for a pattern with just a literal string, and adjusted the
other test where a non-pair was being generated.

The special list_separate function inside the pattern function in
util/__init__.py is used to generate column headings.  This does not
generate display,sort pairs, but that's OK as it is never used in a setting
where pairs can happen.

pfps:
The failure appears to be that some resource is not currently available.  This should be temporary.  Is there a way to rerun the test?

lazka:
This seems to fail on Windows: https://travis-ci.org/quodlibet/quodlibet/jobs/109157054#L2479

pfps:
Aah.

The reason is running the results of format_list through _post.    Should this be done for format_list at all?  
I was only doing that for the display string, not the sort string.  I will put in a patch to _post for both display and sort, which should "fix" the testing, but someone who understands what _post is supposed to be doing to figure out whether _post is a good idea for format_list, even before the changes in this pull request.

lazka:
_post on both seems good.

pfps:
OK, _post is on both display and sort, and all the required tests look good.   I'm running the new patch locally and the results look fine.  I have lots of songs with both sort values and without and the separating and sorting all looks correct even for complex patterns in the paned browser.

lazka:
Thanks!

pfps:
Yeah, well, I did this mostly to fix something that was bugging me!

I'll take a look at the documentation and suggest changes there.  Should I just open an issue for that?

lazka:
Whatever is easier for you.

pfps:
It looks as if I can just do a pull request.  I'll try that. 

(I should  have made the doc change part of this pull request, but I guess it's too late now.)

Issue #1492, opened 2015-03-15
Allow (numeric) tags in RHS of numeric binary operator expressions


lazka:
Original [issue 1492](https://code.google.com/p/quodlibet/issues/detail?id=1492) created by joshua@paddle.com on 2014-11-14T08:08:45.000Z:

My request: Allow comparison operators to apply between two numeric fields.

For example: #(skipcount &lt; playcount)

My reasoning:
I have a lot of music, some songs of which I like more than others. Usually I just leave it on random play, with the result that I hit the &quot;skip&quot; key a couple of times an hour. I recently looked at the stats, and discovered that there some songs I usually skip, and am rarely in the mood for. Rather than trying to manually categorize my music into &quot;songs I always like&quot; and &quot;songs I rarely want to hear&quot;, it would be simpler if I could just filter them by whether I skip them more often than I play them.

So I was astonished to discover that although I can filter by e.g. #(playcount &gt; 10), I can't filter by #(playcount &gt; skipcount). It would make a lot more sense to allow any number in a comparison to be substituted by a numeric attribute.

CreamyCookie:
This was implemented with #1799. (I just tested `#(skipcount < playcount)` and it works.)

Issue #198, opened 2015-03-14
Calculations in a #() Term


lazka:
Original [issue 198](https://code.google.com/p/quodlibet/issues/detail?id=198) created by sp4zzpp2 on 2009-05-03T15:05:34.000Z:

I'd like to see quodlibet to be albe to search something like this:
# (lastplayed &lt; mul(1-rating,5) days)

&quot;mul&quot; represents a multiplication function with 2+ parameters

pschwede:
bump

declension:
Whilst I still think this is an "interesting" feature piece, I'd like to collect people's actual (real-world) use-cases here, i.e. what kind of queries, and operators would be useful (and for what kinds of tags and libraries)

This way, we can make sure we're not missing part of a bigger picture, and so that we can address any quick wins before implementing what could quickly turn into a (slow) generalised numerical expression parser with an awkward, non-infix syntax set.

Obligatory mention of my experiments with [other ways of thinking about queries](https://github.com/quodlibet/quodlibet/issues/821) here too.

pschwede:
#821 is very similar to this idea. We can move the discussion there, I think.

CreamyCookie:
Closing as #1799 addressed the main issue and #821 and #2803 address usability issues.

PR #1799, opened 2016-01-27
New query parser with numeric expressions and query plugins

faubiguy:
This is a rewrite of the code for parsing queries. This allows numeric expressions such as "length + 8 seconds", "playcount * 2", and "(1-rating)*5 days" to be used on both the right and left side of numeric comparisons. Any numerical comparisons that parsed before should still work the same.

This expansion of numeric comparisons addresses issue #1492, and to some extent #198.

This also adds a "@(plugin)" syntax for plugins to extend queries with custom types of searches. This is backwards incompatible with some queries containing '@', however. Three query plugins are included: conditional queries, python expression queries, and including saved searches in queries.

The syntax for the query plugins:
- Conditional queries: `@(if: condition query: then query: else query)`
- Include saved search: `@(saved: some search name)`
- Python expressions: `@(python: expression)`. The `s` variable is the song being matched.

Implementation-wise, this parser parses through each character individually instead of having a separate lexer, which allows the numerical expressions to be much more easily parsed.

The interface of the quodlibet.query.Query class is identical to before, although the internal implementation has changed somewhat.

The following test cases have been added to cover the additions:

tests/test_query.py: TQuery_is_valid.test_extension
tests/test_query.py: TQuery_is_valid.test_numexpr
tests/test_query.py: TQuery.test_numexpr
tests/test_query.py: TQuery.test_numexpr_date
tests/test_query__match.py: TQueryMatch
tests/plugin/test_query.py: TQueryPlugins

The test case TNumericOp in tests/test_query__match.py was removed because it contained only tests pertaining to the previous parser implementation.

I've tried to be backwards compatible with the old parser wherever possible. As far as I can tell, the only exception to this is queries containing '@' (which is now a special character).

This does keep backwards compatibility with date expressions such as '2007-07-19' in numeric comparisons by interpreting them as dates if the comparison contains a 'date' tag and as subtraction otherwise.


pschwede:
Exciting! Is it also possible to have two variables in one @-expression? E.g.: `@(~#playcount / ~#skipcount)`

faubiguy:
Numeric comparisons with more than one tag work, like `#(playcount > skipcount)` or `#(length <  6 minutes * rating)`.

Extensions using `@()` can only have one plugin and argument each, but like other queries can still be combined using `&()` and `|()`.

pschwede:
Okay thanks for clearing up. So, I could filter out `#(0.5 * playcount > skipcount)` which is new in your queries, which is pretty useful, too.

However, these queries (conditions, exactly)  won't work for column headers in browsers, right?

lazka:
Sorry for the delay..

On the pull request:

* That amount of change seems a bit risky for not asking first or opening an
  issue... but ok..
* This pull requests really implements two separate things so it should be split
  in two pull requests.
* Both features seem like nice additions.

On the implementation/features:

Why isn't it possible to implement the numeric expressions on top of the
existing parser or as an additional parser?

faubiguy:
I figured this pull request would be fairly unexpected, but after writing it primarily for my own use I decided to make a pull request in case it's worth including in the main project.

As for the implementation, the lexer parsed all `/` as the beginning of a regular expression, preventing it from also being used as the division operator in different context. Making it lexerless also greatly simplified the numerical expression parsing by preventing having to parse series of tokens like `playcount * `, `(`, `length - 2 minutes`, `)` as a single expression.

If the loss of backwards compatibility with `@` needs to be avoided, I think an option can be added to the preferences menu to enable both the loading of query plugins and the `@()` rule in the parser.

lazka:
> As for the implementation, the lexer parsed all / as the beginning of a regular expression, preventing it from also being used as the division operator in different context. Making it lexerless also greatly simplified the numerical expression parsing by preventing having to parse series of tokens like playcount *, (, length - 2 minutes, ) as a single expression.

I see, thanks. I would hav epreferred if the numexp parsing was separate, but the new code looks readable/documented enough..

> If the loss of backwards compatibility with @ needs to be avoided, I think an option can be added to the preferences menu to enable both the loading of query plugins and the @() rule in the parser.

I don't think that's a big problem.. 

One problem I see is that spaces in tags no longer work and I escaping them doesn't help. e.g. "my tag=foo" no longer works.

faubiguy:
Oh, the tag parsing must have been overly restrictive. Tags containing spaces should now work properly.

pschwede:
`#(rating >= (playcount+1) / (skipcount+1))` is reeeally useful! :+1:

spxxxk:
Can you provide examples for the `@()` syntax ? I can't figure it out on my own ...

faubiguy:
With the python expression query plugin:
* `@(python: s('genre') in s('title'))` find songs where the title contains the genre name. That's a bit contrived, but basically any python expression with the tags works.

With the saved search plugin:
* `&(@(saved: favorites), genre=rock)` finds songs in the saved search titled 'Favorites' whose genre is rock.
* `|(@(saved: mysearch), @(saved: another))` finds songs in either the saved search 'mysearch' or the search 'another'

With the conditional plugin:
* `@(if: #(rating > 0.7), genre=|(classical, electronic), genre=classical)` finds classical music regardless of rating and electronic music which is rated highly.

None of these work unless the corresponding plugin is first enabled in the plugins menu.

spxxxk:
Thanks , _enabled in the plugins_ was the missing bit... 
Nice additions. I will definitely have a use for them.

Note : @-queries appear valid (turn green) even if plugins aren't enabled. Not sure what can be done about it.

spxxxk:
By any chance, does the `@(python: expression)` syntax support sorting by tag ? For example, sorting by last played time ?

lazka:
> By any chance, does the @(python: expression) syntax support sorting by tag ? For example, sorting by last played time ?

No, only filtering.

PR #1810, opened 2016-02-07
Improve lyrics-in-tags support (globally and for ID3 tags in particular)

### LINKED ISSUE #919: Save lyrics to ID3
Original [issue 919](https://code.google.com/p/quodlibet/issues/detail?id=919) created by freyraphael on 2012-02-18T16:08:38.000Z:

I would highly appreciate if the lyrics entered in the Lyrics tab (right click on song -&gt; Information -&gt; Lyrics) would be saved to ID3 instead of saving it locally to ~/.lyrics. And, if a file has an ID3 lyrics tag, Quodlibet should show it in the Lyrics tab.

This would allow viewing lyrics in other music players (Exaile, iTunes, ...) and viewing lyrics in Quodlibet that has been saved to ID3 using another player.

intelfx:
* add support for filling `lyrics` tag from `USLT` frame of ID3v2 tags
(only empty desc and empty lang — others are ignored but not deleted)
* synthesize tag `~lyrics` by looking both at the tags and the external file (in this order)
* in the lyrics pane and the lyrics editor, use tag `~lyrics` for reading
* in the lyrics editor, write both to the `lyrics` tag and to the file

This should fix #919.

### DIFF
--- a/quodlibet/quodlibet/formats/_id3.py
+++ b/quodlibet/quodlibet/formats/_id3.py
@@ -167,6 +167,9 @@ class ID3File(AudioFile):
             elif frame.FrameID == "TLAN":
                 self["language"] = "\n".join(frame.text)
                 continue
+            elif (frame.FrameID == "USLT" and frame.desc == "" and
+                  frame.lang == "\x00\x00\x00"):
+                name = "lyrics"
             else:
                 name = self.IDS.get(frame.FrameID, "").lower()
 
@@ -180,6 +183,9 @@ class ID3File(AudioFile):
                 text = "\n".join(map(unicode, frame.text))
             elif id3id == "COMM":
                 text = "\n".join(frame.text)
+            elif id3id == "USLT":
+                # lyrics are single string, not list
+                text = frame.text
             elif id3id.startswith("W"):
                 text = frame.url
                 frame.encoding = 0
@@ -303,7 +309,7 @@ class ID3File(AudioFile):
             else:
                 tag.add(Kind(encoding=enc, text=text))
 
-        dontwrite = ["genre", "comment", "musicbrainz_trackid"] \
+        dontwrite = ["genre", "comment", "musicbrainz_trackid", "lyrics"] \
             + RG_KEYS + self.TXXX_MAP.values()
 
         if "musicbrainz_trackid" in self.realkeys():
@@ -356,6 +362,13 @@ class ID3File(AudioFile):
             t = self["comment"].split("\n")
             tag.add(mutagen.id3.COMM(encoding=enc, text=t, desc=u"",
                                      lang="\x00\x00\x00"))
+
+        tag.delall("USLT::\x00\x00\x00")
+        if "lyrics" in self:
+            enc = encoding_for(self["lyrics"])
+            # lyrics are single string, not array
+            tag.add(mutagen.id3.USLT(encoding=enc, text=self["lyrics"],
+                                     desc=u"", lang="\x00\x00\x00"))
 
         # Delete old foobar replaygain ..
         for frame in tag.getall("TXXX"):
--- a/quodlibet/quodlibet/qltk/lyrics.py
+++ b/quodlibet/quodlibet/qltk/lyrics.py
@@ -38,12 +38,11 @@ class LyricsPane(Gtk.VBox):
         view.set_wrap_mode(Gtk.WrapMode.WORD)
         sw.set_policy(Gtk.PolicyType.AUTOMATIC, Gtk.PolicyType.AUTOMATIC)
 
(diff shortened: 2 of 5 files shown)


intelfx:
It would be good if someone advised me how to do "detection" in the lyrics editor to save lyrics to file only if saving to tags is unsupported for this file.

quodlibet/quodlibet/formats/_id3.py
```diff
@@ -357,6 +363,12 @@ def write(self):
             tag.add(mutagen.id3.COMM(encoding=enc, text=t, desc=u"",
                                      lang="\x00\x00\x00"))
 
+        if "lyrics" in self:
```
lazka: Please add a test for reading/writing those frames in test_formats__id3.py

quodlibet/quodlibet/qltk/lyrics.py
```diff
@@ -113,23 +113,45 @@ def __search(self, song, buffer, refresh, add):
             GLib.idle_add(buffer.set_text, text)
             GLib.idle_add(refresh.set_sensitive, True)
 
-    def __save(self, save, lyricname, buffer, delete):
+    def __save(self, save, song, buffer, delete):
+        start, end = buffer.get_bounds()
+        text = buffer.get_text(start, end, True)
+
+        # First, write back to the tags.
+        song["lyrics"] = text
```
lazka: text isn't unicode here I think, `.decode("utf-8")`

quodlibet/quodlibet/qltk/lyrics.py
```diff
@@ -113,23 +113,45 @@ def __search(self, song, buffer, refresh, add):
...
+        try:
+            song.write()
+        except NotImplementedError:
```
lazka: Why the NotImplementedError?

quodlibet/quodlibet/qltk/lyrics.py
```diff
@@ -113,23 +113,45 @@ def __search(self, song, buffer, refresh, add):
...
+            pass
+
+        # Then, write to file.
```
lazka: What's this about?

quodlibet/quodlibet/qltk/lyrics.py
```diff
@@ -113,23 +113,45 @@ def __search(self, song, buffer, refresh, add):
...
         except EnvironmentError as err:
             encoding = util.get_locale_encoding()
             print_w(err.strerror.decode(encoding, "replace"))
         delete.set_sensitive(True)
         save.set_sensitive(False)
 
-    def __delete(self, delete, lyricname, save):
+    def __delete(self, delete, song, save):
+        # First, delete from the tags.
+        song["lyrics"] = ""
```
lazka: song.remove("lyrics") or song.pop("lyrics", None)

quodlibet/quodlibet/qltk/lyrics.py
```diff
@@ -113,23 +113,45 @@ def __search(self, song, buffer, refresh, add):
...
+        try:
+            song.write()
+        except NotImplementedError:
```
lazka: Why the NotImplementedError?

quodlibet/quodlibet/formats/_id3.py
intelfx: Done.

quodlibet/quodlibet/qltk/lyrics.py
intelfx: Done. So you mean that the text in the buffer is in the system's global encoding which is not necessarily utf-8?

If yes, then we also need to convert from utf-8 to system encoding on load — how to do that? (/me is not familiar with Python at all, this is probably my first python patch ever)

quodlibet/quodlibet/qltk/lyrics.py
intelfx: Done.

quodlibet/quodlibet/qltk/lyrics.py
intelfx: The default implementation of `write()` in `class AudioFile` throws that.

quodlibet/quodlibet/qltk/lyrics.py
intelfx: Hm? We write the modified text to file's tags, then write the same text to the external file (previous behavior).

BTW, it would be good if you told me how to detect whether the format handler really writes lyrics to tags (and does not ignore them, as it was with ID3 before this patch). If it really does, then we could skip writing to the file altogether. Or keep it that way for forward-compatibility with earlier Quod Libet versions?

intelfx:
@lazka: ping?

PR #1813, opened 2016-02-08
Display systray issue on plasma

### LINKED ISSUE #1756: Minimize to tray on Plasma5 (KDE)
Is there an option or a plugin to install to minimize to tray?

I'm using Plasma5 on Archlinux and when I quit QuodLibet, it quits for real...

Thx

ochurlaud:
XDG_CURRENT_DESKTOP is set with "KDE". 

I asked #plasma on IRC and it's the expected value.

fix #1756

### DIFF
--- a/quodlibet/quodlibet/util/environment.py
+++ b/quodlibet/quodlibet/util/environment.py
@@ -34,7 +34,7 @@ def xdg_get_current_desktop():
 def is_plasma():
     """If we are running under KDE/plasma"""
 
-    return "plasma" in xdg_get_session_desktop()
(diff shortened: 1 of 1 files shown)


lazka:
but this is checking XDG_SESSION_DESKTOP not XDG_CURRENT_DESKTOP

ochurlaud:
I tested both, and both are 'KDE'. Maybe gdm doesn't implement it right? I've no idea.

lazka:
Should be fixed now.. I can't get sddm to work, plasma ends up in a endless crash loop after login :(

PR #1830, opened 2016-02-18
fix determination of tag patterns in songlists

pfps:
This fix makes the determination of whether a tag is a tag pattern to be whether the tag contains a <, not whether it starts with <


pfps:
OK, done

PR #1846, opened 2016-02-22
Fixes #1461: Store ratings in the tags, even if they're the default value.

### LINKED ISSUE #1461: Ratings are not stored in tags (tested with mp3 and FLAC)
Original [issue 1461](https://code.google.com/p/quodlibet/issues/detail?id=1461) created by joschuaga@yahoo.de on 2014-08-30T14:19:07.000Z:

The plugin &quot;Force Write&quot; and &quot;saving rating and play count&quot; are enabled. I use &quot;Banshee&quot; as the email, because I hoped that I could get my old ratings (made by Banshee) this way, but it doesn't seem to work.

<b>- What did you try to do?</b>
I rated something and then checked with mutagenx if the tags for rating are correctly set.

<b>- What did you expect to happen?</b>
I expect it to set the rating tag.

<b>- What did happen instead?</b>
It didn't set the rating tag. (On one mp3 file the rating and the play count was set, but it was 0 instead of the expected 255, because I voted it 4 of 4 stars.)

<b>- Which version of Quod Libet?</b>
3.2.1

<b>- Which operating system including version (Ubuntu 14.04, Win7, Debian</b>
<b>sid)?</b>
Ubuntu 14.04 / 64bit

CreamyCookie:
Currently, if a rating is the same as the default value (usually `0.5`), the rating isn't stored in the tags. Since QL can now internally differentiate between a song that has a rating and one that has the default value, this pull request makes use of that.

### DIFF
--- a/quodlibet/quodlibet/formats/_id3.py
+++ b/quodlibet/quodlibet/formats/_id3.py
@@ -10,7 +10,6 @@ import mutagen.id3
 
 from quodlibet import config, const, print_d
 from quodlibet import util
-from quodlibet.config import RATINGS
 from quodlibet.compat import iteritems
 from quodlibet.util.iso639 import ISO_639_2
 from quodlibet.util.path import get_temp_cover_file
--- a/quodlibet/quodlibet/formats/_id3.py
+++ b/quodlibet/quodlibet/formats/_id3.py
@@ -423,8 +423,7 @@ class ID3File(AudioFile):
                 tag.add(f)
 
         if (config.getboolean("editing", "save_to_songs") and
-                (self("~#rating") != RATINGS.default or
-                 self.get("~#playcount", 0) != 0)):
+                (self.has_rating or self.get("~#playcount", 0) != 0)):
             email = config.get("editing", "save_email").strip()
             email = email or const.EMAIL
             t = mutagen.id3.POPM(email=email,
--- a/quodlibet/quodlibet/formats/xiph.py
+++ b/quodlibet/quodlibet/formats/xiph.py
@@ -255,9 +255,8 @@ class MutagenVCFile(AudioFile):
 
         if config.getboolean("editing", "save_to_songs"):
             email = email or const.EMAIL
-            rating = self("~#rating")
-            if rating != RATINGS.default:
-                comments["rating:" + email] = str(rating)
+            if self.has_rating:
+                comments["rating:" + email] = str(self("~#rating"))
             playcount = self.get("~#playcount", 0)
             if playcount != 0:
                 comments["playcount:" + email] = str(playcount)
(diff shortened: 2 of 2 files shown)


lazka:
https://travis-ci.org/quodlibet/quodlibet/jobs/111043582#L1094

other than that, looks good, thanks

CreamyCookie:
Should be fine now. The failed checks aren't related to my changes, right?

**Slightly related:** The [Force Write plugin](https://github.com/quodlibet/quodlibet/blob/6bd54a9a081fecbf8d5ae0e910c66981739444ab/quodlibet/quodlibet/ext/songsmenu/forcewrite.py) sets `song._needs_write = True`. Shouldn't the same happen when "Save ratings and playcounts" is enabled, whenever playcount / rating have changed?

Currently `_needs_write` is only set in [three files](https://github.com/quodlibet/quodlibet/search?utf8=%E2%9C%93&q=_needs_write). (btw. I think the attribute should be public, because it's not just internally used in `SongWrapper`.)

lazka:
> Should be fine now. The failed checks don't seem related to my commits, or are they?

You removed all usage of RATINGS in that files, so yes.

> Slightly related: The Force Write plugin sets song._needs_write = True. Shouldn't the same happen when "Save ratings and playcounts" is enabled, whenever playcount / rating have changed?
>
> Currently _needs_write is only set in three files. (btw. I think the attribute should be public, because it's not just internally set in SongWrapper.)

We pass "fake" songs to plugins and check them for changes when the plugin finishes and if they have changed write them to disk. The "_needs_write" just forces a rewrite to disk when nothing has changed in the tags and should only be needed in that one plugin.. so no need to expose it more.

CreamyCookie:
> You removed all usage of RATINGS in that files, so yes.

I was referring to the current failed checks, after removing RATINGS.

>  The "_needs_write" just forces a rewrite to disk when nothing has changed in the tags and should only be needed in that one plugin.. so no need to expose it more.

Okay, that makes sense. I'll have to see if that problem of requiring force write for storing ratings still persists.

lazka:
>> You removed all usage of RATINGS in that files, so yes.
>
>I was referring to the current failed checks, after removing RATINGS.

Ah, yeah that was just a transient network error on the builder.. that seems to happen from time to time.

>>  The "_needs_write" just forces a rewrite to disk when nothing has changed in the tags and should only be needed in that one plugin.. so no need to expose it more.
>
>Okay, that makes sense. I'll have to see if that problem of requiring force write for storing ratings still persists.

I don't think we write back tags in case of rating changes atm, so the plugin is probably still needed.

CreamyCookie:
> I don't think we write back tags in case of rating changes atm, so the plugin is probably still needed.

Ah, right! I've created an enhancement report for this: #1852

PR #1916, opened 2016-04-16
Support ReplayGain in mp4

bp0:
Now QL and the QL ReplayGain Scanner will read and write ReplayGain tags from mp4/m4a files in the format used by the foobar2000 and deadbeef media players. There are no standard ReplayGain tags for mp4, it is simply the flac/vorbiscomment named tags and data format as mp4 freeform tags. 

This is not support for Apple's "SoundCheck" iTunNorm tag.

### DIFF
--- a/quodlibet/quodlibet/formats/mp4.py
+++ b/quodlibet/quodlibet/formats/mp4.py
@@ -57,11 +57,11 @@ class MP4File(AudioFile):
             "musicbrainz_albumtype",
         "----:com.apple.iTunes:MusicBrainz Album Release Country":
             "releasecountry",
-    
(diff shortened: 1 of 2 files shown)


lazka:
Please make sure `./setup.py quality` passes.

Also tests would be nice, see `tests/test_format_mp4.py`

bp0:
Now passes ./setup.py quality, although it passed locally even with those trailing spaces, so that is strange.
Added _assert_tag_supported() tests for each of the four tags.

lazka:
Thanks

lazka:
> Now passes ./setup.py quality, although it passed locally even with those trailing spaces, so that is strange.

Probably because pep8 or pyflakes wasn't installed and it skipped those tests. I've made it more strict now in  c4d4f4dd79d

PR #1921, opened 2016-04-20
Added shortcut of <Primary>Delete for moving files to trash (fixed)

victoriahayes:


quodlibet/quodlibet/qltk/songlist.py~
```diff
@@ -0,0 +1,1187 @@
+# -*- coding: utf-8 -*-
```
lazka: This file shouldn't be here I guess :)

victoriahayes: I'll clean it up and repost in a second

PR #1941, opened 2016-05-13
Use App indicator when running under Enlightenment

J5lx:
Some time ago, Enlightenment dropped its XEmbed support since it had never functioned well anyway (See e.g. <https://twitter.com/_Enlightenment_/status/538000507315314688>). Therefore, an app indicator has to be used when running under Enlightenment as well.

I'm not a Python programmer, so I'm not quite sure whether the way I indented the long lines is "correct" and I couldn't find good examples in QL's source tree within a reasonable amount of time.


declension:
LGTM

declension:
Thanks @J5lx

PR #1946, opened 2016-05-17
bs1770gain related ReplayGain tag handling

bp0:
The complete set of ReplayGain tags that is produced by bs1770gain is now viewable and editable in the tag editor. The set includes replaygain_[reference_loudness|algorithm|track_range|album_range].
Also, the ReplayGain scanner plugin can now delete those tags when it scans, so the set of ReplayGain tags is consistent and correct if scanning files that were previously scanned with bs1770gain or something else that might create them. This is a user preference, that is on (delete unused) by default.
The replaygain_reference_loudness patch that was previously rejected is also included, because why not? The value is available from gstreamer, even if it is currently a constant, so why not have the option to store it? It is off (not stored) by default, a user will have to request that it be stored.


lazka:
Sorry, but "why not" is not a good reason. I've pushed 04ed141cd9f7c4c310658a4595d7befe2aaf2ccb and 249cdfffd4e92fe929ad0e9568f1937a3d844d44 for this.

Issue #1967, opened 2016-06-18
Add option to display lenght in seconds instead of hh:mm:ss


MartinBorg:
I prefer the length of tracks displayed in seconds rather than hours:minutes:seconds.
Sounds odd, but is useful in some situations, at least for me.

lazka:
In which context do you want it to display seconds (in the UI)?

MartinBorg:
Practically any context.

lazka:
OK, thanks. Could you go into detail why you prefer seconds and in which situations it is useful?

MartinBorg:
1. When I estimate roughly the length of multiple tracks, addition of decimal numbers is easier for me than with a combination of decimal and babylonian, sexagesimal numbers (hh:mm:ss).
2. When I need the length in seconds to use in another program, e.g. I'm looking for a track of a certain length for a video and this video (scene) is given in seconds.

I assume, that not many people would make use of this feature. That's why I put the option "hidden" in the config file. But I would change the implementation to look up the config option only once at startup.

declension:
Fixed in #2021

PR #1968, opened 2016-06-18
add option to display length of tracks in seconds (#1967)

MartinBorg:


lazka:
(Sorry for the delay)

I'd really like to avoid adding an option for this... -> #1967

MartinBorg:
Don't you like the feature at all or do you prefer another implementation?
If the latter, would it be possible to write a plugin?

(Btw. the implementation in this pull request is inefficient, because the config object is accessed many times. Maybe it would be better to just override `format_time()` and `format_time_long()` once after parsing the config file.)

declension:
This is now implemented as part of the changes in #2021, so closing this.

Issue #1716, opened 2015-10-18
Usability: Remove «View» –> «Disable Browser» (double)


timpulver:
In «View» there are two nearly similar items: «Browser» and «Disable Browser». «Browser» is enough to toggle the browser pane, so «Disable Browser» can be removed.

lazka:
Disable Browser also is without a search bar, so there is a difference. But yeah, we could remove the "Disable Browser" one and merge the "Search Browser" with the "Paned Browser" by adding a button to toggle the panes.

timpulver:
Didn’t use the search browser yet, have to get used to all the functions…

Just noticed that clicking »Browser« when it is checked does not result in the browser pane being hidden. When you have an element associated to a checkmark, it’s confusing when you cannot enable and disable it by clicking. It should represent a boolean state which is editable. On all elements in the view menu I would assume they behave like a boolean.

lazka:
Ah, I see. gtk-mac-integration translates the radio items to checkboxes for the global menu.

lazka:
This is how it's supposed to look like:

![screenshot from 2015-10-20 15-26-56](https://cloud.githubusercontent.com/assets/991986/10608382/1582968e-773f-11e5-9010-4b71d23bc78a.png)

timpulver:
This is how it looks on OS X 10.10:

<img width="611" alt="bildschirmfoto 2015-10-20 um 15 26 59" src="https://cloud.githubusercontent.com/assets/1055819/10608394/1f9611d2-773f-11e5-9dd8-286bc89ef54e.png">

timpulver:
I guess this is not a bug but a usability issue.

lazka:
Looks like OSX doesn't support radio items in menus. So not sure what to do there.

Looking at your your screenshot, I think one thing contributing to the confusion is that some items are actions while others are nouns. Changing them all to nouns might help ("Kein Browser", "Suchansicht",...)

timpulver:
That’s one thing yes. Verb items near subject items are strange.
If you could get rid of «Deactivate Browser» there would only be «Search Library» left. That one should go to an «Edit» menu imo.

Issue #1598, opened 2015-04-29
Tidy up application menu


CreamyCookie:
The current application menu has some redundancies and some items that aren't ordered semantically. Here's my suggestion:

* **Quod Libet**
  * Preferences
  * Plugins
  * ------------------
  * Online Help
  * Search Help
  * About
  * ------------------
  * Quit

* **View** # no change

* **Library**
  * Add a file
  * Add a folder
  * Add a location
  * ------------------
  * Refresh Library
  * Search Library

* **Filters** # no change

* **Control**
  * Previous
  * Play
  * Next
  * Stop after this song

* **Current song**
  * Jump to playing song
  * Add bookmark
  * Edit bookmark
  * ------------------
  * Edit Tags
  * Information

----

**Related enhancements / bugs:**
* #1527 - Add the context menu of the current song under "Control" in the menu bar
* #1601 - "Filter on genre" is missing in the context menu of the current song
* #1602 - Add all "saved searches" to the "Filters" submenu in the menu bar

declension:
Some good points there (having a _Library_ top-level menu makes a lot of sense), and a few that don't work for me.

@CreamyCookie, do you think you could split this ticket into logically distinct tasks, as there are too many things to implement for this to get done any time soon.

CreamyCookie:
@declension Thanks!

1. I would leave this bug report here and extract / create the following 2 additional ones. Is that okay with you?
  * insert saved searches have (including the current 4 default ones: all, recently played, recently added, top40)
  * "Filter on genre" seems to be missing in the context menu of the current song.
2. Which are the ones you don't like? Maybe I can win convince you (or you me).. :smile:

declension:
@CreamyCookie that sounds great. Perhaps reduce this ticket to being about a Library top-level menu and add others as necessary.

@lazka and I agreed recently that _extra_ top-level menus should probably be avoided if possible (there are already enough), especially with Ubuntu (15) adding its own (currently).

lazka:
The Ubuntu menu is gone now #1599. For a plugin menu we have #152

Here is the Gnome HIG btw: https://developer.gnome.org/hig/stable/menu-bars.html.en
(but given that Gnome is moving away from menu bars, ignoring it shouldn't do much harm)

CreamyCookie:
@declension I have extracted the two bug posts and cleaned up this one a bit. (Is that wording correct, or should it be "cleaned this one up a bit"?)

[...]

PR #1975, opened 2016-07-05
Complete re-think of application menu (#1598)

declension:
* Disable Browser aka `EmptyBar` is now gone (this fixes some of #1716 etc)
 * Make top-level menus closer to Gnome HIG
 * Add a few missing accels in menu
 * _Filters_ now belongs in _Browse_
 * Simplify menu creation code a bit
 * Browser menu widget accessing is now centralised (to allow easier rethinking of menus..)
 * Update `POTFILES` and documentation
 * Fix some broken tests


declension:
@lazka some significant changes (UI - code change not so much), so wanted to make sure you were happy with these, or had any more suggestions for this part...

lazka:
Random thoughts without looking at the code:

* The filter menu in extra browser menus seems gone
* I don't like the "Information" in "View" as it doesn't affect the view but opens a new window. What about renaming "Edit" to "Song" (aka current song) and put it there as well?
* In the filters it might be a good idea to rename the top filters to "Filter on _current_ Genre" to make it clear where the input value is coming from.

declension:
Thanks @lazka 

    The filter menu in extra browser menus seems gone

Hmm, I didn't even notice that there all this time. Thanks, looks like an oversimplification on my part


    I don't like the "Information" in "View" as it doesn't affect the view but opens a new window. What about renaming "Edit" to "Song" (aka current song) and put it there as well?

I see. Yes that could work I suppose


    In the filters it might be a good idea to rename the top filters to "Filter on current Genre" to make it clear where the input value is coming from.

Yes, makes sense. Those filters are a bit awkward to fit in anywhere but I'll rename them

Issue #1828, opened 2016-02-17
Add Soundcloud Browser


declension:
This would be nice; Clementine has one as do a few other things.

Imagine it would be either similar (or even _an extension_) to the Internet Radio Browser. Obviously this would require users signing up.

There is a [lightweight API](https://developers.soundcloud.com/docs/api/guide#uploading) and everything.

Ideally this would be implemented using Browsers as plugins (see #157).

nodiscc:
It would be a great addition. There is https://github.com/soundcloud/soundcloud-python -  A Python wrapper around the Soundcloud API. Do you know if one would be able to use this plugin (browsing/searching/playing soundcloud tracks) without a soundcloud user account?

declension:
@nodiscc - no it would / will be account-only for now. The API is geared that way, and most of the interesting features come with that anyway. It wouldn't require a pro account anyway..

declension:
Some experimental work in the [souncloud-browser branch](https://github.com/quodlibet/quodlibet/tree/soundcloud-browser).

Pretty rough still, but the authentication flow, basic UI, searching, and integration with QL is working (for me).

Needs the `soundcloud` module, though longer term might try to ditch that and use the GTK+ libraries.

lazka:
> Some experimental work in the souncloud-browser branch.

Cool, got it working, kinda. Please tell if you want feedback.

> Needs the soundcloud module, though longer term might try to ditch that and use the GTK+ libraries.

As soundcloud isn't in debian (I think?) that would make it easier. On Windows/OSX shipping soundcloud from pypi isn't a problem.

declension:
Good point. I've not used the Soup stuff but it looks OK so I'll try duplicating as little of the helper methods as needed to get it working.

And yes feedback would be useful.... in particular I'm debating how to integrate searching with QL in  a way that makes sense for a/any remote "library" of millions of tracks... that supports almost none of the querying we do... Currently it's keeping a local cache of results and then using normal filtering which feels "native" but will fail miserably with non-text searches...

lazka:
Also have a look at util.thread.call_async() + if you want we can use requests instead of urllib. Might make the update to py3 easier anyway.

(I don't know what the API exposes..)
I'd just try to replicate the online version and only allow text queries, for starters.
Have a combobox selecting between Tracks/Playlists/People/Groups. + a sidebar containing the filters + the results for playlist/people/group search.

declension:
[...]

PR #1990, opened 2016-07-19
Add Soundcloud browser

declension:
See #1828 for background.

This collection of, err, ~~40~~ 34 (+ merge) commits is what hopefully now represents a minimum viable (and compliant, as far as my checks have gone) feature for Soundcloud, plus a bunch of "free extras" I did along the way (see checklists in #1828).

The technical commentary is generally in the various detailed commit messages (so please don't squash).

Seems working from here now, but probably contains a few bugs / inefficiencies / gotchas still so it would be good to validate this on a few different systems. Features / improvements can be added in later PRs, as this one has got far too big really... :wink:


lazka:
Two blockers imo, but if that's get handled after merge, fine with me.

* I don't think we can ship the icons
* Does the login still depend on a custom URI scheme? We need something that works on OSX/Windows.

declension:
Thanks

>    I don't think we can ship the icons

Ah, yes that. What's the specific concern - GPL? Conversely, SC seems to be quite keen on _always_ displaying the icons (["If you are using SoundCloud's data in your app or site, we ask that you place these logos where the data is being used/shown"](https://developers.soundcloud.com/docs/api/buttons-logos) etc). Might need to look at that web-caching idea you had then...

>    Does the login still depend on a custom URI scheme? We need something that works on OSX/Windows.

It still uses it where available (it's definitely a UX win), but I added a third state for the button (logging in) whereby you can enter the code manually copied from the redirect page (even pastes it from the clipboard by default). This should allow all users then, but I think the OS X registration might be doable too

lazka:
>>    I don't think we can ship the icons
>
>Ah, yes that. What's the specific concern - GPL? Conversely, SC seems to be quite keen on _always_ displaying the icons (["If you are using SoundCloud's data in your app or site, we ask that you place these logos where the data is being used/shown"](https://developers.soundcloud.com/docs/api/buttons-logos) etc). Might need to look at that web-caching idea you had then...

A Gtk.Image subclass which does an async http request shouldn't be too hard. I can look into that if you want.

>>    Does the login still depend on a custom URI scheme? We need something that works on OSX/Windows.
>
> It still uses it where available (it's definitely a UX win), but I added a third state for the button (logging in) whereby you can enter the code manually copied from the redirect page (even pastes it from the clipboard by default). This should allow all users then, but I think the OS X registration might be doable too

Yeah, I've used the manual code input. But once you log in it redirects you to a browser error page, and you have to figure out that part of the URL has to be copied to QL. No UX win in that case.

declension:
Redirect problem resolved in https://github.com/quodlibet/quodlibet.github.io/commit/f4892bdb485d83b3cd22d1e9105f25b46bfca5b0

declension:
> A Gtk.Image subclass which does an async http request shouldn't be too hard. I can look into that if you want.

@lazka that would be great, if you have time

lazka:
here is a basic implementation: https://gist.github.com/lazka/87b59993b3cf01ed420d78dda03e1eac

lazka:
Nice, one more thing: As the images are defined globally the http requests happen on every QL start, even if another browser is selected.

Either move them locally, or use our ``@cached_func`` decorator.
Or we could only start loading them when the mapped in screen, but I'd prefer the first solution as I'd like to avoid creating Gtk.Widget instances at import time.

lazka:
Sorry for the conflicts.

Feel free to merge

declension:
No worries - merging then :)

PR #1993, opened 2016-07-21
Add new uri helpers

lazka:
I'd like to get rid of the URI class as most of it is unused


declension:
I guess the removal itself is in a forthcoming commit?

lazka:
Yeah, WIP. I was using travis to run the windows tests

PR #2020, opened 2016-08-29
tests: switch to pytest as the main test runner

lazka:
Allows us to get rid of a bunch of test discovery code
and gives us a few new features.

"--to-run=TFoo,TBar" still works by passing the class
names to pytest using the "-k" switch. This accepts
more stuff, but for the basic class name list it
should give the same results.

To skip network and code quality tests pytest markers
are used:
  @pytest.mark.quality
  @pytest.mark.network


lazka:
Still needs travis changes and the windows/osx packages are missing dependencies...

@declension any objections/thoughts?

declension:
I haven't used pytest but it's meant to be good, and less is more etc.

Will we still get colourised output?

lazka:
You can give it a try on master

```
sudo apt install python-pytest
py.test tests/test_qltk_x.py
py.test tests
```

declension:
IDE support seems good (better, even) too :100:

declension:
Oh and regarding look-and-feel, [pytest-sugar](https://pypi.python.org/pypi/pytest-sugar) is oh so sweet.

I'm _definitely_ sold now.. :smile:

PR #2023, opened 2016-09-01
Include senf

lazka:
Work in progress.... (so I can make travis run windows/osx tests for me)

All our platform specific path/environ/argv/print code is moved to an external lib, but still included in QL since we can't depend on it.


declension:
Cool. Couldn't we use a Git submodule here? (though nobody seems to like them)

lazka:
Maybe? Never used it.

lazka:
Looking a bit into submodule, it looks kinda complicated. Also we need a subdirectory of the other repo which doesn't seem to work.

PR #2031, opened 2016-09-11
Update nb.po

akrosikam:
- translated all new and changed strings
- replaced awkward anglicistic phrases and words with more common and proper Norwegian ones (see http://i18n.skolelinux.no/nb/Fellesordl.eng-no.html)


lazka:
Thanks

Issue #1411, opened 2015-03-15
Play Order rework


lazka:
Original [issue 1411](https://code.google.com/p/quodlibet/issues/detail?id=1411) created by reiter.christoph on 2014-06-06T18:19:31.000Z:

Meta bug for play order changes...

Current design: http://f.666kb.com/i/cozwusc91t7xzr57s.png

* Two modes, shuffle/inorder
* Menu to select a modifier(check item) e.g. follow cursor / variant(radio item) e.g. weighted random

We can implement that on top of the current interface by grouping all orders in two groups... a good first step I guess. For the things below we probably need two kinds of plugins, orders and order wrappers.

Feedback welcome.

Menu examples:
```
--------------------------- <- modifiers
| [x] follow cursor (falls back to variant if cursor doesn't change)
| [ ] queue only (only affects clicks)
| [ ] autofill queue
| [ ] one song (falls back to variant on explicit track changes)
| [ ] track repeat (submenu to select count?) (falls back to variant on explicit track changes)
|-------------------------- <- variants
| o weigthed (random only)
| o spotify (random only)
| o reverse (inorder only)
---------------------------
```

The queue could have:

```
----------
| [x] stop after empty
| [ ] consume songs (remove after played)
| [ ] shuffle (or a menu button like in the main UI but only with variants, to keep things simple)
----------
```

Other bugs which should be considered:

* Issue #1407 - Spotify shuffle
* Issue #1111 - Repeat-block based play mode modifier
* Issue #903 - &quot;Next&quot; should not shuffle a new song if within a string of already played songs
* Issue #894 - Add &quot;party shuffle&quot; play order
* Issue #703 - Support un-shuffleable tracks for DJ mixes / classical pieces

Queue related:

* Issue #442 - The Play Queue should use the Play Order setting
* Issue #43 - rfe: stop playing once queue runs out

Implementation related:

* Issue #575 - Play order plugins should handle song list changes

timpulver:
Very hard to follow this discussion and bring in new ideas. Could you post a summary of the problem and what might be planned?

declension:
The closer I get to a refactoring path the more Kafkaesque the codebase seems. So some collected thoughts / plan:

### Highlights
* No complete rethink right now. The good ideas around pre-filling a queue e.g.  #1407, #904 would make this far too big to take on right now unfortunately.
* However, enable more powerful and flexible combinations of repeat and shuffle
[...]

Issue #152, opened 2015-03-14
Plugins should have some space on the main window (menu or hbox)


lazka:
Original [issue 152](https://code.google.com/p/quodlibet/issues/detail?id=152) created by steven.strobe.cc on 2009-03-17T04:07:21.000Z:

There's plenty of space in the main window for plugins to expose extra
functionality, and many things that plugins could do with something that
has constant presence. Two places seem like obvious candidates: a new
'Plugins' menu, and in the HBox along the bottom of the screen which
currently holds only selection length (and scanning progress at startup)
and order controls.

Plugins that use these spaces should not depend on them, and should have a
seamless fallback which works on versions without this mechanism.

declension:
Additional idea: When exposing SongsMenu plugins, allow some method of DnD of songs onto them.

declension:
Currently toying with a simple but promising proof of concept involving a sidebar (most screens are wide nowadays so going using horizontal space is logical), available to `EventPlugin`s only (others don't make so much sense).

Seems reasonable to also expose icon space straight after the shuffle / repeat buttons, as this is more likely to be useful (one example: Track Repeat configuration)

@lazka - any tips on making a widget (containing a `Gtk.Notebook` in this case) resizeable, a bit like the Queue, but without being a `Gtk.Expander` I guess? Also, for making this sidebar toggleable - I was thinking of resurrecting the View Menu, and adding a menu item for the ever-confusing queue toggle as well, what do you reckon?

lazka:
I imagined something like this for starters:

```
... | Plugins | Help
     ---------
      Plugin 1 (opens plugin prefs)
      Plugin 2 -> | Custom Action 1 (provided by the plugin)
                  | Custom Action 2 (..same)
                  | Custom Toggle 1 (..same)
                  |-------------------
                  | Plugin Preferences
```

What's the content of the notebook? If you want it to be resizable you have to put it into a qltk.Paned.

declension:
Thanks, I'll try using the `qltk.Paned`.

I guess I didn't explain well - the change is mainly adding a sidebar area to the main window, at one side of the window:
```
/----------------------------------\
|    INFO                     ART  |
------------------------------------
|                           |      |
|     S O N G S             |   S  |
|                           |   I  |
|                           |   D  |
|                           |   E  |
\----------------------------------/
```
[...]

PR #2036, opened 2016-09-17
Added ModeBar

bernd-wechner:
My very first QL patch. Am getting my head around git, and github in the process and trying to set up up a workflow. 

Am very open to feedback. But also seriopusly appreciate accepting the request. Happy to modify it so it's optional if preferred, and take advice on how to make it optional. Or to look at internationalizing the label (could not see how to do that easily) or otherwise to make it fit the visions of look and feel I guess.

Real estate in this area is abundant on any normal sized screen and I found the two mode buttons cryptic at first and find their status hard to see (contrast between up and down images is low). So I added a very rough shod simple text statement as to the current mode. 

All feedback or improvement suggestions welcome. Am testing the ropes wit QL, PyDev and Git..

### DIFF
--- a/quodlibet/quodlibet/qltk/quodlibetwindow.py
+++ b/quodlibet/quodlibet/qltk/quodlibetwindow.py
@@ -9,7 +9,7 @@
 
 import os
 
-from gi.repository import Gtk, Gdk, GLib, Gio, GObject
+from gi.repository import Gtk, Gdk, GLib, Gio, GObject, Pango
 from senf import uri2fsn
 
 import quodlibet
@@ -407,6 +407,22 @@ class RepeatButton(Gtk.ToggleButton):
         self.connect('toggled', toggled_cb)
 
 
+class ModeBar(Gtk.Label):
+    def __init__(self, repeatmode, shufflemode):
+        super(ModeBar, self).__init__()
+
+        self.set_text(repeatmode, shufflemode)
+        self.set_ellipsize(Pango.EllipsizeMode.END)
+
+        self.show_all()
+
+    def set_text(self, repeatmode, shufflemode):
+        rm_text = "Repeat" if repeatmode else ""
+        sm_text = "Shuffle" if shufflemode else ""
+        conj = " and " if repeatmode and shufflemode else ""
+        super(ModeBar, self).set_text(sm_text + conj + rm_text)
+
+
 class QueueButton(Gtk.ToggleButton):
 
     def __init__(self):
@@ -436,11 +452,17 @@ class StatusBarBox(Gtk.HBox):
 
         self.order = order = PlayOrder(model, player)
         self.pack_start(order, False, True, 0)
+        shuffle = order._shuffle
+        shuffle.connect('toggled', self.__shuffle, model)
+        model.shuffle = shuffle.get_active()
 
         self.repeat = repeat = RepeatButton()
         self.pack_start(repeat, False, True, 0)
         repeat.connect('toggled', self.__repeat, model)
         model.repeat = repeat.get_active()
+
+        self.modebar = ModeBar(model.repeat, model.shuffle)
+        self.pack_start(self.modebar, False, True, 0)
 
         self.statusbar = StatusBar(TaskController.default_instance)
         self.pack_start(self.statusbar, True, True, 0)
@@ -453,6 +475,11 @@ class StatusBarBox(Gtk.HBox):
 
     def __repeat(self, button, model):
         model.repeat = button.get_active()
+        self.modebar.set_text(model.repeat, model.shuffle)
+
+    def __shuffle(self, button, model):
+        model.shuffle = button.get_active()
+        self.modebar.set_text(model.repeat, model.shuffle)
 
 
 class AppMenu(object):


lazka:
> My very first QL patch. Am getting my head around git, and github in the process and trying to set up up a workflow. 

:)

> Am very open to feedback. But also seriopusly appreciate accepting the request. Happy to modify it so it's optional if preferred, and take advice on how to make it optional. Or to look at internationalizing the label (could not see how to do that easily) or otherwise to make it fit the visions of look and feel I guess.

For translations you just have to wrap all translatable text with the magic `_` function e.g. `_("My text")`. But you have to keep in mind that translators will only see the text without context. So instead of `" and "` it would be better use something like `"%(repeat)s and %(shuffle)s"`, or just provide three fixed strings for all cases.

> Real estate in this area is abundant on any normal sized screen and I found the two mode buttons cryptic at first and find their status hard to see (contrast between up and down images is low). So I added a very rough shod simple text statement as to the current mode. 

Imo the two icons are a well established UI pattern by now and I'm not sure if adding a text label will make it easier/faster (besides for screen readers perhaps) to parse the state. Maybe someone else has an opinion to add here..

Not sure how easy that is to implement, but what about changing the color of the icons when they are active instead?

> All feedback or improvement suggestions welcome. Am testing the ropes wit QL, PyDev and Git..

declension:
Thanks

+1 to what @lazka said really, i.e. once those i18n problems fixed there would be no _code_ problems, but... whilst I agree the current icons could be a bit clearer (or even larger), having pulled the PR, I don't see that _text_ here would be a UI improvement for users given the clutter it brings.

That said, these might make it more palatable to users:
 * Feature made optional via a prefs widget
 * Using the full name (_One Track_ / _Track Repeat_ etc) might be more useful
 * The text should probably be in parentheses i.e.  _(Weighted Shuffle, Repeat)_, hinting it's a translation of the button states to an extent.

There was a lot of discussion around on a big overhaul of play order (partially implemented now) in #1411, it's worth a look.

bernd-wechner:
Thanks for the feedback so far. Very much appreciated (Am well into a rework of the library scanner on another local branch right now, more on that to come).

Am rearing to fix this one though as suggested. But some questions if you can spare a moment to handhold a rapidly upskilling newb:

@lazka 

1) What do you mean by `"%(repeat)s and %(shuffle)s"` ? 
I'd have though plurals and 'and' are all language specific so a conjunction like `" " + _("and") + " " ` seems appropriate.

2) WRT IMO a text feedback is pretty customary on other players and makes it readily visible. That said in my experience it's usually near the what's playing features, that show you the playing track etc, and it might be more at home there. For now was aiming at a MVP (minimum viable product) for a newb to offer up and get comment on.

@declension 

3) Sounds great and I like. I'll look at a prefs widget to propose as that's an area I want to upskill in. But if UI clutter is an issue, a config file setting with no UI is an option too. It is Quod Libet after all, Whatever You Like ... and that's one reason I've landed with it, namely a confidence that the community culture at least supports mainlining configurability to individual tastes and preferences.

4) Any tips on how to divine the current nuance of Repeat and Shuffle modes would please me, but I'll poke around too in the iterim. Reason I ask is the Repeat button I see (and I'm only using on browser, the File System one for now) simply toggles on/off doesn't have any nuancal options and Shuffle has only two `In Order` and `One Song` and I admit I find that wholly cryptic (i.e. I have no idea what it means). 

5) Rendering in parenthesis or not or fonts or location is all up for advice as I'm pretty unfussed. The UX desire is to see clearly with a glance at screen the current play mode I guess. That's what I'm angling for.

6) Will read #1411 later. A quick check now suggests much to digest and I'm not focussed enough right now.

7) As a total side observation `_` seems overloaded sadly, which surprised me for a language that doesn't support overloading ;-). But I'm getting warnings on unused variables on a list unpack from a function/generator return and so I reflexively replace that variable with `_`. Which is the custom and works fine. But if that is done in the class then the `_` can no longer be used for internationalization. Odd. At a  total guess for now I just named it dummy and I get no more unused variable warning (seems the semantics of that name are respected). Ah, am still learning python after a year or two of coding with it - but have never done internationalization before I guess)

declension:
> 3) Sounds great and I like. I'll look at a prefs widget to propose as that's an area I want to upskill in. But if UI clutter is an issue, a config file setting with no UI is an option too.

Yes that's possibly better in cases like this - see the advanced preferences plugin for this purpose. However, as this discussion has prompted me to revisit #1411, including clarifying the confusing situation with the repeat button you mention. Note also that play-order plugins are exposed here, so it's not as simple as it first looks. I've got a few pending changes, so perhaps wait and see how they affect this before going any further.

In fact, the real solution for your aim here is a plugin (automatically and easily solving the prefs issue) once #152 is finally started, which I've also started thinking about now). Plugins are by far the most  useful, powerful and readily accepted development option especially for a first commit.  See the [plugin development documentation](https://quodlibet.readthedocs.io/en/latest/development/plugins.html). (note: in future, please file an issue here first before pull requests... and we do have over 400 open...)


> 7) As a total side observation _ seems overloaded sadly, which surprised me for a language that doesn't support overloading ;-). But I'm getting warnings on unused variables on a list unpack from a function/generator return and so I reflexively replace that variable with _. Which is the custom and works fine. But if that is done in the class then the _ can no longer be used for internationalization.

As a side reply: yes, it's pretty annoying. This is because `_` is `gettext` thus more of a C thing. See the [Python gettext docs](https://docs.python.org/2/library/gettext.html). Some people use `__` as the throwaway variable, though I usually like to give it a name of what it was anyway.

lazka:
> 1) What do you mean by `"%(repeat)s and %(shuffle)s"` ? 
I'd have thought plurals and 'and' are all language specific so a conjunction like `" " + _("and") + " " ` seems appropriate.

See https://docs.python.org/2/library/stdtypes.html#string-formatting

No idea, my language skills are limited. If in doubt try to make it easier to translate I guess. You can have a look at the translation .po files to get a feeling for what translators see: https://github.com/quodlibet/quodlibet/tree/master/quodlibet/po

----

Regarding "_": master now imports those functions explicitly, so this should be less confusing.

----

Regarding this PR, still not convinced that this is an improvement, so closing.

PR #2046, opened 2016-09-30
Added Waveform seek bar

0x1777:


quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def update(self, rms_vals, player):
+        self._rms_vals = rms_vals
+        self.queue_draw()
+        self._player = player
+
+    def do_draw(self, cr):
+        # Paint the background
+        context = self.get_style_context()
+        bg_color = context.get_background_color(Gtk.StateFlags.NORMAL)
```
lazka: This no longer works on newer gtk ( and is probably the source of 100% CPU) You need to do:

```
context.save()
context.set_state(Gtk.StateFlags.NORMAL)
bg_color = context.get_background_color(context.get_state())
context.restore()
```

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -0,0 +1,277 @@
...
+            # The index of the closest rms value
+            i = int(x * ratio_width)
+            cr.move_to(x, 20)
+            # Draw line up and down
+            cr.line_to(x, 20 - self._rms_vals[i] / ratio_height * 0.5)
+            cr.line_to(x, 20 + self._rms_vals[i] / ratio_height * 0.5)
+
+            cr.stroke()
+
+    def do_realize(self):
```
lazka: Why no subclass a Gtk.EventBox?

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -0,0 +1,277 @@
...
+        height = allocation.height
+        value_count = len(self._rms_vals)
+        max_value = max(self._rms_vals)
+        ratio_width = value_count / float(width)
+        ratio_height = max_value / float(height)
+
+        elapsed_color = Gdk.RGBA()
+        elapsed_color.parse(get_fg_color())
+
+        remaining_color = context.get_color(Gtk.StateFlags.SELECTED)
```
lazka: same as above

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -0,0 +1,277 @@
...
+
+        pipeline.set_state(Gst.State.PLAYING)
+
+        self._pipeline = pipeline
+        self._rms_vals = []
+
+    def _on_bus_message(self, bus, message):
+        if message.type == Gst.MessageType.ERROR:
+            error, debug = message.parse_error()
+            print("Error received from element {name}: {error}".format(
```
lazka: Please use `quodlibet.print_d` instead of print() everywhere. It adds logging and works better under Windows.

quodlibet/quodlibet/ext/events/waveformseekbar.py
0x1777: Yeah that's a better solution. Fixed the other things too.

lazka:
Thanks for the update. A few more things I noticed while using it:

* It crashes when clicking on it after enabling the plugin before switching songs
* When enabling it it needs a song change for it to generate the waveform
* When no song is active (after the playlist is finished) the last waveform stays visible
* It would be nice if there was some kind of "computing state" when switching songs before the waveform is computed instead of showing the old one. Maybe just a thick line in the same style or something. The current behavior makes it hard to tell if the shown waveform is the current one or it just wasn't computed yet.

Except for the first point (the crash), this can be deferred to future PR/commits if you want.

nodiscc:
Thank you! Definitely looks like a killer feature to navigate in long tracks, etc. I have not tested it yet, looking forward to it.

> It would be nice if there was some kind of "computing state" when switching songs

What usually works is to show a striped pattern (eg light gray diagonal stripes ` ///////`) in place of the waveform. In any case it should not be confusing whether waveform is computing/track is broken/track is silence/track is 0db noise, so the pattern must be remarkable.

0x1777:
Ok, I think I fixed all 4 points. It simply displays a horizontal line now when no waveform is available.

lazka:
> What usually works is to show a striped pattern (eg light gray diagonal stripes ///////) in place of the waveform. In any case it should not be confusing whether waveform is computing/track is broken/track is silence/track is 0db noise, so the pattern must be remarkable.

Valid point.

> Ok, I think I fixed all 4 points. It simply displays a horizontal line now when no waveform is available.

Looks good!

lazka:
One more thing: Any thoughts on using a theme color for the "elapsed time" part of the waveform instead of the configurable value? That would make it work with all themes.

0x1777:
Sounds good, that would make it look more consistent.
I'd say the colors of the scale widget would be good. But I don't know how to get them from the theme.

I tried `context.get_color()` and `context.get_background_color()` with different `Gtk.StateFlags.*` values but they are all the same.

`elapsed_color = context.lookup_color("selected_bg_color")[1]` seems to work with some themes (like Arc) but not with others (like Adwaita).

lazka:
The fg color for the LINK state should work. That's for inline links in
labels.

0x1777:
Yes that works. It'll use the theme's color by default now unless the user overrides it in the config.
I'll open a new pull request then.

PR #2048, opened 2016-10-01
Use GTK theme color by default

0x1777:


quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -206,18 +206,21 @@ def draw_placeholder(self, cr, width, height, color):
...
-        context.save()
-        context.set_state(Gtk.StateFlags.SELECTED)
         fg_color = context.get_color(context.get_state())
+        context.set_state(Gtk.StateFlags.LINK)
+        elapsed_color = context.get_color(context.get_state())
         context.restore()
 
+        # Check if the user set a different color in the config
+        elapsed_color_config = get_elapsed_color()
+        if Gdk.RGBA().parse(elapsed_color_config):
```
lazka: Gdk.RGBA.parse does not take None which get_elapsed_color() returns by default

0x1777: Oh, right. Fixed.

Issue #1861, opened 2016-02-29
Paned browser not updating on playlist-creation


timpulver:
After I created a playlist in playlist-browser and adding music to it, the paned browser with a `~playlists`-column does not update.

declension:
At a guess because that's a (very) internal tag, and adding songs to a playlist is not then firing the `changed` signal for them. It's never come up before though, so would have to investigate.

Issue #518, opened 2015-03-14
Separate library for playlist songs.


lazka:
Original [issue 518](https://code.google.com/p/quodlibet/issues/detail?id=518) created by reiter.christoph on 2010-05-25T00:05:00.000Z:

Like in the file system view.

declension:
The only way I can see to deal with missing / moved files (surviving restarts) is to keep some metadata in the `Playlist` collection (now I see it I remember extracting this class hierarchy to indicate we might try other ways of implementing `Playlist` as well as the current `FileBackedPlaylist`... :thinking: )  

Despite liking the playlist-is-a-local-URL-encoded-file-path metaphor, it seems this is what's holding this up. If it had a structured format with enough metadata cached there (ugh: invalidation), then something could be displayed if it's missing (e.g. see #3105) or it could be repaired / re-found if moved externally. 

It's conceivable that we could do two in one here - XSPF (see #1122) would be more than enough structure to support our needs I believe, and some of the management code from the Playlists Browser could be moved into a `Library` layer for a first start, maybe?

PR #2051, opened 2016-10-06
Update paned browser when updating a playlist

### LINKED ISSUE #2017: Browse panes not updated
It appears that in the specific case of adding playlists to the panelbrowser, with `~playlists`, see #2011, and then adding or removing songs to playlists (which I do with the right click menu), the panel view isn't update until I restart quodlibet. So I suppose the pane is populated once, and not dynamically updated. Is that true for all panes?

weyfonk:
When songs are added to or deleted from a playlist, the paned
browser is updated in case a custom `~playlists` pane
has been configured. Same goes when a new playlist is created
or deleted.
This fixes #2017 and #1861.

### DIFF
--- a/quodlibet/quodlibet/browsers/playlists/menu.py
+++ b/quodlibet/quodlibet/browsers/playlists/menu.py
@@ -9,6 +9,7 @@ from gi.repository import Gtk, Pango
 
 from quodlibet import qltk
 from quodlibet import ngettext, _
+from quodlibet.browsers.paned import PanedBrowser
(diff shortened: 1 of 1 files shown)


declension:
Thanks for the PR.

Whilst I like the simplicity (and that it fixes the immediate problems), quite keen not to introduce coupling across the browsers; in fact the existing tight (but less controversial) coupling between `Playlist` and `PlaylistsBrowser` was something I was hoping to get rid of.

That said, until #518 is solved, doing this properly and for _all_ browsers, is hard (I've got a patch that mostly solves #2017 though).

PR #2052, opened 2016-10-06
Seek bar plugin: invert scrolling directions

### LINKED ISSUE #1958: Alternative seek bar plugin: wheel down now seeks backward and wheel up forward
It's the opposite of the original seek bar.

weyfonk:
This commit makes the alternative seekbar's scrolling directions
consistent with the default seekbar's behaviour, ie scrolling down
seeks forward while scrolling up seeks backward.
This fixes #1958.


declension:
Thanks

PR #2063, opened 2016-10-14
Waveform seek plugin: performance / drawing improvements

declension:
* Enhance resolution for lower file lengths which was blocky:
waveform-resolution-problems-for-short-songs

 * Make sure waveform is at least a pixel high, otherwise you can't see the end of the track.
 * Speed up drawing by only drawing one line, not two
 * Speed up computation by maxing out callbacks at a reasonably large value.
 * Use warning message for unexpected Gstreamer data
 * Average, don't _sample_ data for more accurate (and smoother) display.
 * Port all config to Conf Proxies
 * Add Hi-res option - single pixel
 * Bugfix: Use float in position calculation, was getting stuck sometimes
 * Bugfix: make sure quotes in filename (I have some) are escaped

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -1,15 +1,21 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 0x1777
+#           2016 Nick Boultbee
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


declension:
cc @0x1777 - assuming you're still happy I'll merge this shortly.

0x1777:
Yeah looks good.

declension:
Thanks

PR #2068, opened 2016-10-23
Use python futures. Fixes #2067

### LINKED ISSUE #2067: lots of wakeups
QL does 10 wakeups per second when idle.

* Using "strace -y -f python quodlibet.py" shows there are lots of select syscalls in a thread.
* Attaching "gdb --pid $(pgrep quodlibet)", "break select", "cont"

Points to https://hg.python.org/cpython/file/2.7/Lib/multiprocessing/pool.py#l327

:(

lazka:


declension:
Cool :smile:

Issue #241, opened 2015-03-14
New kind of display for the albums list


lazka:
Original [issue 241](https://code.google.com/p/quodlibet/issues/detail?id=241) created by brunobergot on 2009-06-24T17:40:51.000Z:

Hi and sorry for my poor english,

I really like the look and feel of the albums list. This list can be very
long when you have a lot of albums, and generally i like to search an album
only by scanning the covers &quot;au hasard&quot;.

I think it would be great to allow users to modifiy some settings about the
display of this list :
- show textuals informations or not
- show this information when the mouse hover a cover
- size of the covers

Some years ago i was using a plugin called album list for winamp. I would
love to have this type of display in the album list of QL.

Thanks for reading

brunob:
Work is on in my fork here : https://bitbucket.org/b_b_/quodlibet

brunob:
Switched my work here : https://github.com/brunob/quodlibet/tree/covergrid

brunob:
See also #1489 (in order to centralize issues about this feature).

Lalopx:
Is there a way to change the album view, or add the option to the right clivk menu?? i would love to help with some coding. Also, hello to evryone :+1:

darthoctopus:
@brunob's fork seems to have died. I've taken the liberty of getting covergrid to work with 3.7.1, but only after brutally commenting out a bunch of calls to GtkIconView.set_cursor. Would you be open to a pull request for covergrid as-is?

brunob:
@darthoctopus i haven't touched to my fork since a while cause i never knowed if it would be merge one day... I think it would be better to ask to @lazka to have a look at this, maybe a pull request of your fork would suit to him.

nodiscc:
Duplicate of https://github.com/quodlibet/quodlibet/issues/1101?
Related https://github.com/quodlibet/quodlibet/issues/1665.

brunob:
@nodiscc #1101 is the duplicate of thie one present here ;)

nodiscc:
I've just drafted what an album art based view should look at in my humble opinion. Feedback welcome.

![ql-albugrid](https://cloud.githubusercontent.com/assets/3950390/19727202/c257fca4-9b8e-11e6-9c5b-961686d67135.png)
![ql-hover](https://cloud.githubusercontent.com/assets/3950390/19727203/c29725dc-9b8e-11e6-9231-c4e6c2af5303.png)

Clicking anywhere on an album art selects it and display details in the left pane.
[...]

PR #2071, opened 2016-10-27
Covergrid Browser

darthoctopus:
Updated @brunob's covergrid to work with latest commit (see #241)


lazka:
Thanks. As you can see tests are failing, see https://quodlibet.readthedocs.io/en/latest/development/testing.html for more details.

For the browser in general, I think it's still a bit rough, and performance is horrible, but maybe including it will push us to fix things and prevent further bitrot

@declension any thoughts?

darthoctopus:
could you elaborate on what you mean by being rough? e.g. UI not sufficiently polished? I think one sorely needed thing (not sure how to implement this) would be the ability to set a different cover size from the one used by the Album browser in GUI preferences (perhaps also toggle between RHPaned and RVPaned, not sure how to implement either)

quodlibet/quodlibet/browsers/covergrid/prefs.py
```diff
@@ -0,0 +1,103 @@
...
+from quodlibet import _
+from quodlibet.browsers._base import FakeDisplayItem, EditDisplayPatternMixin
+from quodlibet.formats import PEOPLE
+from quodlibet.qltk import Button, Icons
+from quodlibet.qltk.ccb import ConfigCheckButton
+from quodlibet.util import format_rating
+from quodlibet.util.i18n import numeric_phrase
+
+PEOPLE
+_SOME_PEOPLE = "\n".join([util.tag("artist"), util.tag("performer"),
```
declension: Hmm we should really extract all this code, but that can be done separately

quodlibet/quodlibet/browsers/covergrid/prefs.py
```diff
@@ -0,0 +1,103 @@
...
+                         util.tag("composer"), util.tag("arranger"), ])
+
+_EMPTY = _("Songs not in an album")
+DEFAULT_PATTERN_TEXT = """[b]<album|<album>|%s>[/b]<date| (<date>)>
+[small]<~discs|<~discs> - ><~tracks> - <~long-length>[/small]
+<~people>""" % _EMPTY
+
+
+class Preferences(qltk.UniqueWindow, EditDisplayPatternMixin):
```
declension: Does this in fact extend `browsers.albums.Prefs`? It looks almost identical

declension:
@darthoctopus 

Firstly thanks for your (and @brunob's) work so far. It's definitely one of the most common requests for browsers and it's starting to look good.

### Fixes
* As to the current state, now that the tests are green, the most pressing problem I had was on clicking the Preferences menu item:

```
  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/__init__.py", line 753, in wrap
    return handler(that, *args)
  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/browsers/covergrid/prefs.py", line 62, in __init__
    cb.set_active(config.getboolean("browsers", "album_text"))
  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/config.py", line 173, in getboolean
    raise Error(e)
ConfigParser.Error: <exception str() failed>
```
* The display pattern doesn't seem to be centred properly relative to the Album Art (when large): 
covergrid alignment issue


### Performance
I agree with @lazka really - the performance is still the single killer for this PR. On this machine (4GHz I7) it takes:

 *  7-8 seconds to clear a query to none (i.e. redraw all albums)
 *  7-8 seconds to re-render the window on resize with no query selected

Both are occasionally enough for the Unity to darken the window, i.e. indicate an unresponsive UI.

I've noticed that as soon as the query is specific enough (e.g. `artist=nas`) to narrow down the result set a lot, the performance is actually very fast. You can see this happen in realtime as the first letters are  typed (`artist=n`). (at the very least, this browser should override the timeout for query updating).

I have around 8500 songs (across 1800 albums) in the library, which is an order of magnitude less than some of our users.

### Discussion
* _Should_ this be a subclass of AlbumList (currently seems like it works well as such)
* Perhaps then, the prefs should subclass its prefs too as there's large overlap.
* Maybe we could pad the browser window, not sure.
* Does the _All Albums_ item make sense in this browser?

Thanks

lazka:
> ### Discussion
> * _Should_ this be a subclass of AlbumList (currently seems like it works well as such)

I'd avoid subclassing browsers, this has always made things complicated in the past. 

> * Perhaps then, the prefs should subclass its prefs too as there's large overlap.

Same, at least until we are happy with how things are.

but ymmv

Regarding the performance, the only way I know to make things faster with iconview (which has same nasty O(n^3)) is to move the text inside the cover and make the cover size fixed. Another might be to use Gtk.FlowBox (not available on Ubuntu 14.04)

declension:
> > Should this be a subclass of AlbumList (currently seems like it works well as such)

> I'd avoid subclassing browsers, this has always made things complicated in the past.

I tend to agree actually. @darthoctopus probably better to start a _new_ browser here, extract any _necessarily_ common code to mixins.

quodlibet/quodlibet/browsers/covergrid/prefs.py
darthoctopus: I'm not sure how to properly extend browsers.albums.Preferences without also invoking the AlbumList preferences.

darthoctopus:
> this browser should override the timeout for query updating

where can I find documentation on this? Sorry, I'm quite new to both Gtk and quodlibet

darthoctopus:
I've implemented @lazka's suggestion to fix item width, could you try again to see if this improves performance? I've also fixed the centering issue.

lazka:
Still not really usable here, but if all fails we can disable it for the next release.

declension:
Thanks all

brunob:
(y)

urielz:
Just chiming in my feedback. I have a relatively small library (around 250 albums) and it is very slow on my end as well. However, I’d rather have this than no cover grid at all! So very happy to see this pr merged.
 
Question: are images cached? does the file size of images have an impact on performance? (All cover images I have are fairly big, .5 to 1Mb each, and nothing is embedded as metadata...)

lazka:
> Question: are images cached? does the file size of images have an impact on performance? (All cover images I have are fairly big, .5 to 1Mb each, and nothing is embedded as metadata...)

thumbnails of external images are cached

Issue #2083, opened 2016-11-05
"Random album broken" since...


draxil:
Looking at the random album plugin to make some changes.. Found that it's not working in HEAD, backtracked it to the commit:

commit dcfb0471f666a6a1fc8eed2b08be97f2ac442a6d
Play Order: address PR comments

If I check out the previous:

commit c3d6b8ec9e975576d9b4d0ab0458c78583bdaf09

It works. Might look into fixing this but wanted to report this detective work at least in case I loose the thread.

draxil:
Should also mention that this is in the "Album list" browser that I have been testing, not sure about other browsers.

Issue #2072, opened 2016-10-27
The "random album" plugin thinks it can function in the playlist browser


draxil:
The random album plugin thinks it can function in the playlist browser, I assume because the playlist browser reports it can filter by album.

The way this results for me most of the time is it searches the playlist browser for an album that produces no  results and hence the music stops. 

I am not sure what the right fix is here. Probably some other means of telling the random album plugin to determine that it doesn't play well when the playlist browser is up?

I'd be happy to contribute to a fix, but I'm not really sure on the correct path.

declension:
Hmm, yes. The semantics of filtering are a little weird at best in the playlist browser. Yes, it does the right thing (uses the generated query to filter the results of the _current_ playlist), but this is not helpful with the random album plugin.

I'd suggest:
*  the ugly quick fix is indeed to special-case the playlists browser in the Random Album plugin.
* Alternatively rethinking filtering support in playlists browser, though this would be removing features
* The better fix, which I've had in mind since the play order extensions, is to rework this (excellent) plugin to be a shuffle plugin. This would also allow track-level shuffling with the same powerful weighting preferences as this. Possibly needs #894 or similar - i.e. some form of queing / pre-deciding the next tracks a long way (entire album, here) in advance.

draxil:
My pull request contains the "ugly fix". The others sound better with more time and more work :)

lazka:
I guess this can be closed now.

PR #2084, opened 2016-11-05
Fixed broken random album plugin

draxil:
The logic got inverted in commit dcfb0471f666a6a1fc8eed2b08be97f2ac442a6d, fairly innocently I think due to confusing wording


draxil:
By the way this fixes #2083

draxil:
Thrown my other change to address #2072 in here as well.

lazka:
> The logic got inverted in commit dcfb047, fairly innocently I think due to confusing wording

looks good

> Thrown my other change to address #2072 in here as well.

please don't mix multiple fixes if possible

PR #2099, opened 2016-11-14
Create Tagging

John-Peters:
I am using version 3.7.1 Linux version (Ubuntu 64 bits).
I have two questions:
When I create tags from the file name, how can I remove the file extension?
After adding the album photo how do I save it for all files in the album?
Thanks for your help

### DIFF
--- a/Tagging
+++ b/Tagging
@@ -1,0 +1,1 @@
+When you create the tags from the filename, how can you remove the file extension?


lazka:
>  When I create tags from the file name, how can I remove the file extension?

It should be removed automatically. Can you give an example for a filename and the tag pattern?

>  After adding the album photo how do I save it for all files in the album?

In case you are referring to embedding the image in the files:

There is a "Edit Embedded Images" plugin which, if activated, exposes the follow context menu item: "Edit Embedded Images" > "Embed Current Image"

John-Peters:
Merci pour votre aide.
En fait, le problème semble être que quand je clique sur "aperçu" il ne se passe rien......
Je vous joins la capture d'écran
etiquettes par chemin

John-Peters:
Sorry, I'll start again in English ...!!
In fact the problem seems to be that when I click on Preview I don't see anything happen.
I attach the screen image
John Peters

lazka:
Try `<tracknumber> <title>` instead. The format in the text entry at the top has to match the format of the filenames.

John-Peters:
Same situation. Must be something basic I haven't understood....
capture

lazka:
Make sure there is only one space between `<tracknumber>` and `<title>`

John-Peters:
OK GREAT !!!! I had TWO spaces. With one space it works fine.
Many thanks

lazka:
:)

John-Peters:
Here's the screenshot. That's a REALLY useful feature. Thanks
capture ok

Issue #1827, opened 2016-02-15
Decide for one method of storing the configurations of plugins


CreamyCookie:
Currently there seem to be multiple options for storing the configuration of a plugin (ConfProp, PluginConfigMixin, and PluginConfig). Changing this would make the codebase more consistent, and thus easier to understand and maintain.

declension:
Thanks

lazka:
I prefer PluginConfig as it allows to set defaults in one place and doesn't use inheritance.

CreamyCookie:
I prefer the `ConfProp` classes, because compared to something like [this](https://github.com/quodlibet/quodlibet/blob/6bd54a9a081fecbf8d5ae0e910c66981739444ab/quodlibet/quodlibet/ext/playorder/track_repeat.py#L60):

```python
play_each = int(self.config_get('play_each', self.PLAY_EACH_DEFAULT))
```

* You can just access and set them like normal attributes (e.g. `self.conf.zoom_level = 3`). Thus the code looks cleaner and your IDE can provide code completion. 
* You don't need to know the exact type when using them. (You don't need to care whether `zoom_level` is a float or int.) There's only a single point where the type of a configuration option is defined.
* You can create your own reusable classes for encapsulating more complex data. One example is `ColorConfProp`, which can transparently load the colors from a string into a RGB tuple, and vice versa. (ConfigParser only supports strings, integers, floats and booleans.)
* It clearly separates the persistent data (e.g. `self.conf.alternate_search_url`) from the temporary data (e.g. `self._view`) of a plugin.

PR #2114, opened 2016-11-22
Update waveformseekbar.py

urielz:
User-friendly color picker for waveform
It uses class PluginConfigMixin pretty much in the same way the synchronized lyrics plugin does it

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -11,8 +11,8 @@ from math import ceil, floor
 
 from quodlibet import _, app
 from quodlibet import print_w
-from quodlibet.plugins import PluginConfigMixin,PluginConfig, BoolConfProp, IntConfProp, \
-    ConfProp
+from quodlibet.plugins import PluginConfig, BoolConfProp, IntConfProp, \
+    ConfProp, PluginConfigMixin
 from quodlibet.plugins.events import EventPlugin
 from quodlibet.qltk import Align
 from quodlibet.qltk import Icons
@@ -285,7 +285,7 @@ class Config(object):
 CONFIG = Config()
 
 
-class WaveformSeekBarPlugin(EventPlugin,PluginConfigMixin):
+class WaveformSeekBarPlugin(EventPlugin, PluginConfigMixin):
     """The plugin class."""
 
     PLUGIN_ID = "WaveformSeekBar"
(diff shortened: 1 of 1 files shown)


declension:
Hmm, whilst I like(d) `PluginConfigMixin` (though am biased I guess), using _both_ systems seems bad, and switching between them doesn't seem related to this change.

Could you not just do it with the config proxies?

urielz:
by _both_ you mean PluginConfigMixin and PluginConfig?

Thanks for the tips. I have zero experience with GTK and close to non with Python so this is mostly a learning experience for me. Might take a while to get it right :)

declension:
> by both you mean PluginConfigMixin and PluginConfig?

Yes, there are I think *four* ways used of "doing config" in plugins. These two are probably the most common and take different approaches. The proxy method (`PluginConfig`) is newer and a bit more abstracted... there's discussion on #1827.

> Thanks for the tips. I have zero experience with GTK and close to non with Python so this is mostly a learning experience for me. Might take a while to get it right :)

No worries. Yes, probably will, but that's fine. Python should be fine but I find GTK+ (in Python) can be a near-undocumented black hole of time once you stray off the well-trodden paths. Luckily @lazka is around :smile:  (read [his excellent GTK+ 3 guide](https://python-gtk-3-tutorial.readthedocs.io/en/latest/) actually).

The most important aspect you already know - _QL is far, far more clever than it first seems_.

urielz:
Excellent. I understand now. Thanks for pointing me in the right direction! I'll give it a try whenever I find some more free time.

lazka:
> read his excellent GTK+ 3 guide actually

(I only worked on the API docs, not the gtk tutorial)

lazka:
With the newest osx bundle it's now possible to run "setup.py quality". See #2117

knockoutMice:
Could we also modify it so the slider doesn't jump, but _slides_? I haven't looked into the issue, but it deeply frustrates me. :)

If nobody jumps in, I'm gonna do it tomorrow if I don't need to modify the behavior of the original seekbar.

urielz:
@knockoutMice too busy at the moment to fiddle with this. feel free to jump in. thanks!

knockoutMice:
#2138

declension:
Should we close this PR now in favour of #2138?

urielz:
@declension yes, fine by me. I'm working on something else atm. I'll open another one if I comeback to this.

PR #2115, opened 2016-11-23
Covergrid: Toggle between horizontal and vertical pane orientations

darthoctopus:


lazka:
Thanks.

Running "./setup.py quality" would safe you the trouble of waiting for travis to do the pep8/pyflakes checks

PR #2119, opened 2016-11-25
Information Widget: code clean up and slight facelift

### LINKED ISSUE #1558: Add "comment" field to information window.
Original [issue 1558](https://code.google.com/p/quodlibet/issues/detail?id=1558) created by hehn@phys.ethz.ch on 2015-03-13T22:19:47.000Z:

In the information window of a single song I would like to have a display showing the content of the &quot;comment&quot; tag of the file.
In many podcasts the comment field contains useful information about its content and I am missing a convenient way of accessing this field.

I wrote a tiny patch which would add this feature to the information window. I would be very happy if you considered applying the patch or adding this feature in another way.

declension:
* For OneSong, lyrics and bookmarks are optional
 * Start cleaning up and extracting common code in advance of proper modularisation
 * Use more modern features where available (`defaultdict`, comprehensions, `if` expressions) to shorten code
 * Lose the bold and underline, go for a lighter approach in keeping with other UIs / web
 * Italicise tracks / album names where relevant, to indicate these are (chosen) names not values.
 * Add `comments` (Fixes #1558)  `website` (mainly for Soundcloud) into a new section.
 * Use regular table layout more for neatness
 * More is possible / changeable, this is just to make sure people are happyish :smile:


quodlibet/quodlibet/qltk/information.py
```diff
@@ -34,18 +36,49 @@ def Label(label=None):
...
+def Table(rows):
+    t = Gtk.Table(n_rows=rows, n_columns=2)
+    t.set_col_spacings(6)
+    t.set_row_spacings(6)
+    t.set_homogeneous(False)
+    return t
+
+
+def get_colors(widget):
+    context = widget.get_style_context()
```
lazka: This needs save()/set_state()/restore(): https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/ext/events/waveformseekbar.py#L232

quodlibet/quodlibet/qltk/information.py
```diff
@@ -34,18 +36,49 @@ def Label(label=None):
...
+    bg_color = context.get_background_color(Gtk.StateFlags.NORMAL)
```
lazka: get_background_color no longer works with newer gtk+ (it only returns the css solid color part, which might not be used when rendering afaik)

lazka:
regarding colors and layout, maybe interesting:

* There is https://lazka.github.io/pgi-docs/#Gtk-3.0/constants.html#Gtk.STYLE_CLASS_DIM_LABEL
* and https://developer.gnome.org/hig/stable/visual-layout.html.en#label-alignment

quodlibet/quodlibet/qltk/information.py
declension: Thanks - was guessing this (hence test failures)

quodlibet/quodlibet/qltk/information.py
declension: Oh, I saw that code originally... applied a few inlinings (assuming referential transparency etc, I guess) and it seemed redundant. Seemed to work without it, so I left the simplified version.

Guess it does something magic on state saving?

lazka: Yeah, it's not obvious and only a problem since gtk 3.20+. The problem is that the getters invalidate things for gtk and when called from the wrong code path you get in an invalidation loop. (new-state -> update css/drawing -> new-state ...).

declension: Madness! Thanks

lazka:
(btw: it's always a good idea to test theme specific things with the Adwaita theme (also the dark variant), check out the theme switcher plugin)

declension:
Yep, might just ditch that colours thing. Wanted to avoid using bold (only) for the frames but increasingly I'm believing that `Gtk.Expander`s might be the nicest experience... especially if they could made to save their state too, one day at least (like the queue I guess)

lazka:
lgtm

declension:
Thanks

PR #2127, opened 2016-11-28
Add register-date filter to lastfmsync plugin

knockoutMice:
If the chart is older than the register date of the user, don't download it. (so the download doesn't start with ~2005 every time.)


lazka:
Thanks

PR #2130, opened 2016-11-29
Fix covergrid preferences: slider release

knockoutMice:
The UI lagged as hell when I tried to change the cover magnification with the slider. Now it only updates the settings / redraws the covers when:
- The user releases the slider
- The new setting is different from the old one

### DIFF
--- a/quodlibet/quodlibet/browsers/covergrid/prefs.py
+++ b/quodlibet/quodlibet/browsers/covergrid/prefs.py
@@ -3,6 +3,7 @@
 #           2009-2010 Steven Robertson
 #      2012,2013,2016 Nick Boultbee
 #           2009-2013 Christoph Reiter
+#                2016 Mice Pápai
 #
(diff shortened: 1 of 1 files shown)


lazka:
We have `util.DeferredSignal` for those situations.

How does `mag_scale.connect('value-changed',  util.DeferredSignal(mag_changed, owner=mag_scale))` perform?

Not using value-changed would break other things changing the slider, like the keyboard (or code)

knockoutMice:
You're right. I'm looking into DefferedSignal. 

By the way my mouse wheel jumps 5.0... 

Maybe an "Update" or a "Preview" button would be the best solution.

knockoutMice:
DeferredSignal behaves the same as the original one and lags heavily.

I introduced a mag_lock boolean to disable updating while sliding with the cursor.
Also fixed the mouse wheel problem (5.0 steps).

PR #2132, opened 2016-11-30
Fix object has no attribute 'registered'

knockoutMice:


lazka:
This didn't help. The problem was due to loading serialized instances form disk. Fixed in ee3754e3d8d7a6, hopefully.

PR #2136, opened 2016-12-01
Discogs Cover Source

knockoutMice:
Downloads covers from Discogs. I set the priority to 0.1 as I'm still testing it, but seems to be working with everything I throw at it.


knockoutMice:
I wrote a DiscogsParser for the Download Album Art plugin (songsmenu/albumart.py), so now it supports Discogs besides Amazon.

PR #2144, opened 2016-12-12
TypeError on song length being not a number

pschwede:
Did you recently change that line? That problem didn't occur since recently.


lazka:
`AudioFile.__setitem__` got stricter recently

PR #2169, opened 2016-12-23
Create da.po (danish translation)

scootergrisen:
I give you danish translation for Quod Libet and Ex Falso.


lazka:
Thanks! And thanks for reporting all those issues

PR #2173, opened 2016-12-26
Some updates and fixes to the Czech translation

mrksu:
Not that many strings updated (there's a lot of work needed) but I will send more later on.

Should I put everything into one big pull request or multiple smaller ones as I progress? I'm kinda new to git so I don't know what works best for you.

Also, will these translations make it into the 3.8 version or does that have to be translated separately?


lazka:
> Not that many strings updated (there's a lot of work needed) but I will send more later on.

Thanks

> Should I put everything into one big pull request or multiple smaller ones as I progress? I'm kinda new to git so I don't know what works best for you.

Whatever is easier for you

> Also, will these translations make it into the 3.8 version or does that have to be translated separately?

I can apply this to the 3.8 branch, but some things might get marked fuzzy since things have diverged a bit.

---

Please post/notify when we should merge this.

mrksu:
Thanks for your reply. You can merge this whenever you like, the changes are reasonably self-contained, I think.

v3.8 or even v3.7 could benefit from it, too, if it's doable – I'm using 3.7.1 now and I see many untranslated or badly translated strings, some of which I've fixed in this pull request. However, I'll continue working on master only if that's all right.

lazka:
> v3.8 or even v3.7 could benefit from it, too, if it's doable – I'm using 3.7.1 now and I see many untranslated or badly translated strings, some of which I've fixed in this pull request. However, I'll continue working on master only if that's all right.

I'll try to get it into 3.8

PR #2180, opened 2017-01-02
Spelling (unsed => used)

eno93:


NoahAndrews:
Pretty sure the intended word was actually "unset". Used doesn't make sense grammatically here.

eno93:
Made a guess, haven't actually tested, either way a correction of some sort is needed.

lazka:
Fixed in master, thanks

PR #2188, opened 2017-01-09
Updated German translation

Mellthas:
Updated German translation which was quite a bit out of date. Also, all the plugins are now translated. I left three unclear strings untranslated for now. I’ll file a separate issue on that.

Up for debate: I changed the translation of "pattern" to "Muster" instead of "Vorlage" in general because that seemed more appropriate to me and the expression was used many times more compared to the time the translation was last updated. Other than that, I tried to stay consistent with the existing translation.

### DIFF
--- a/quodlibet/po/de.po
+++ b/quodlibet/po/de.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: Quod Libet 3.8\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2017-01-06 13:37+0100\n"
-"PO-Revision-Date: 2017-01-09 22:58+0100\n"
+"PO-Revision-Date: 2017-01-10 20:33+0100\n"
 "Last-Translator: Rüdiger Arp <ruediger@gmx.net>\n"
 "Language-Team: GERMAN <LL@li.org>\n"
 "Language: de_DE\n"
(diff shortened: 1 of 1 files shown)


lazka:
Thanks. Some tests fail, but they fail to state the exact reason, pushed a fix for that.

Here is the output for now:

```
E   AssertionError: One or more messages did not pass (ending punctuation missing).
E   "Couldn't find the 'Kanji Kana Simple Inverter' (kakasi)." - "»Kanji Kana Simple Inverter« (kakasi) wurde nicht gefunden" (de.po:1373, ../quodlibet/ext/editing/kakasi.py:63)
E   "Library directory the server connects to." - "Bibliotheksverzeichnis, mit dem sich der Server verbindet" (de.po:2884, ../quodlibet/ext/_shared/squeezebox/base.py:130)
E   "Reload all songs in your library. This can take a long time." - "Alle Titel in der Bibliothek neu laden (dauert möglicherweise lange)" (de.po:5301, ../quodlibet/qltk/prefs.py:660)
E   "_Customize Headers…" - "Listenspalten _anpassen ..." (de.po:5726, ../quodlibet/qltk/songlist.py:1112)
```

> Up for debate: I changed the translation of "pattern" to "Muster" instead of "Vorlage" in general because that seemed more appropriate to me and the expression was used many times more compared to the time the translation was last updated. Other than that, I tried to stay consistent with the existing translation.

Sounds good to me, thanks.

Should we try to address the issues you raised first or merge this once tests pass?

Mellthas:
Fixed the punctuation errors. Can I run these tests locally? Because using `./setup.py test --to-run PO.de` (as recommended in the translation how-to) I get no errors.

I think the issues with the unclear strings are minor and can be addressed later.

lazka:
These tests optionally depend on polib, called "python-polib" or "python2-polib" in most distros. I guess we should depend on it.

PR #2197, opened 2017-01-14
fix popup menu handling

darthoctopus:
GtkIconView doesn't emit popup-menu events for right clicks but DOES when pressing the menu button on the keyboard. This should fix that use case.

### DIFF
--- a/quodlibet/quodlibet/browsers/covergrid/main.py
+++ b/quodlibet/quodlibet/browsers/covergrid/main.py
@@ -470,8 +470,8 @@ class CoverGrid(Browser, util.InstanceTracker, VisibleUpdate,
 
         menu = SongsMenu(library, songs, items=[items])
         menu.show_all()
-        menu.popup(None, None, None, Gdk.BUTTON_SECONDARY,
-            Gtk.get_current_event_time(),
+        menu.popup(None, None, None, None,
+            Gdk.BUTTON_SECONDARY,
             Gtk.get_current_event_time())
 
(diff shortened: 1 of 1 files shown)


CreamyCookie:
Thanks. Sadly, it doesn't seem to work.

Interesting note here from the [docs on `Gtk.Menu.popup`](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Menu.html#Gtk.Menu.popup) which is [used in `covegrid/main.py`](https://github.com/quodlibet/quodlibet/blob/2a94ff3ab66fb2bd323f5903cb22ff8fabec5a85/quodlibet/quodlibet/browsers/covergrid/main.py#L450) to open the context menu: 

> Note that this function does not work very well on GDK backends that do not have global coordinates, such as Wayland or Mir. You should probably use one of the gtk_menu_popup_at_ variants, which do not have this problem.
>
> Deprecated since version 3.22: Please use Gtk.Menu.popup_at_widget(), Gtk.Menu.popup_at_pointer(). or Gtk.Menu.popup_at_rect() instead

CreamyCookie:
@darthoctopus Gave it another try. Sadly still no context menu.

lazka:
What about qltk.popup_menu_at_widget()?

CreamyCookie:
@lazka Nice. That does work!

CreamyCookie:
Last change does work! Thank you both!

lazka:
thanks

Issue #2193, opened 2017-01-11
Add "merge album"/"split album"


lazka:
Select some songs, right click:

1) In case they are all the same album expose a "split album" option which tries to find some differences like different directory, or artist, which splits the thing in two albums.

2) In case there are multiple albums expose "merge albums" which tries to remove any differences, fill missing values.

In both cases a small summary could be shown which shows the proposed changes and what the outcome would be.

CreamyCookie:
Merge sounds great. Personally, I don't remember ever having needed to split an album, but I can see how it could be useful. A summary with proposed changes is a good idea, as it would also double as a confirmation to prevent accidental merges.

lazka:
Split is mostly needed for "Best Of" albums

ambergrant:
thumbs up on this suggestion!

Issue #202, opened 2015-03-14
Feature: Support multiple versions of albums/songs (FLAC/mp3/etc)


lazka:
Original [issue 202](https://code.google.com/p/quodlibet/issues/detail?id=202) created by tyler.cordrey on 2009-05-16T21:07:31.000Z:

I'd love for my media library to be aware of multiple versions of the same
track, and play the highest quality by default. Maybe I currently only have
an mp3 of a particular song, but later get a FLAC copy. Or maybe I'd like
to keep mp3's around for my portable device, but prefer flacs when
listening on the computer. Viewing Info for a track would show all versions
the program knows about.

Ideally Quod Libet would recognize songs with some tolerance for
differences in file/folder or tag names (ie AlbumName [FLAC] vs AlbumName
[MP3]). Personally I would structure my library with root folders for each
version type, but this would have to be up to the user. I've envisioned
accomplishing this through scripts to build symlinks, updating the links to
point to the apropriate place with the highest quality. It seems that this
can be accomplished more gracefully using an existing tool like Quod Libet.

declension:
With 7 years since last update I think it's safe to now close this

PR #2199, opened 2017-01-15
Added '~dirname'-tag to album_key of AudioFile

cwecht:
The album_key of AudioFiles has been expanded by adding the directory name. This is useful if one has multiple copies of the same album in the library. Without this patch these two copies are merged together in one album in albumlist-view; with this patch they are splitted in two albums.

### DIFF
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -146,7 +146,7 @@ class AudioFile(dict, ImageContainer):
                 human(self("albumartistsort", "")),
                 self.get("album_grouping_key") or self.get("labelid") or
                 self.get("musicbrainz_albumid") or "",
-                self.__call__("~dirname"))
+                os.path.dirname(self.get("~filename") or ""))
 
     @util.cached_property
     def sort_key(self):
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -145,7 +145,8 @@ class AudioFile(dict, ImageContainer):
         return (human(self("albumsort", "")),
                 human(self("albumartistsort", "")),
                 self.get("album_grouping_key") or self.get("labelid") or
-                self.get("musicbrainz_albumid") or "")
+                self.get("musicbrainz_albumid") or "",
+                self.__call__("~dirname"))
 
     @util.cached_property
     def sort_key(self):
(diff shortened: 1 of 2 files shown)


lazka:
this breaks having different discs in different directories. The easiest way to split albums is to add an "albumartist" tag

quodlibet/quodlibet/formats/_audio.py
```diff
@@ -145,7 +145,8 @@ def album_key(self):
         return (human(self("albumsort", "")),
                 human(self("albumartistsort", "")),
                 self.get("album_grouping_key") or self.get("labelid") or
-                self.get("musicbrainz_albumid") or "")
+                self.get("musicbrainz_albumid") or "",
+                self.__call__("~dirname"))
```
lazka: this is the same as `self("~dirname")` in python, only slower

lazka:
I've proposed some related functionality in #2193 recently, feel free to comment there

cwecht:
The intention with this was not to keep albums with the same name but different artist apart. This makes it possible to have two identical copies of one album, which are viewed as different albums albumlist-view.

I have basically two versions of my library: one for archival storage and one to sync with a mobile device. I'd like to manage them both in quodlibet at the same time. In the setup #2193 would be not great help as you had to do the distinction manually.

But I see your point. Maybe It would be a viable solution to make this optional.

declension:
+1 to what @lazka says - this is a common use case (for me personally actually), where albums are split over multiple directories e.g. by genre 

Also breaks multi-disc albums with dirs like `CD1`, `CD2` etc

lazka:
> The intention with this was not to keep albums with the same name but different artist apart. This makes it possible to have two identical copies of one album, which are viewed as different albums albumlist-view.

I see, we have some keys (for historical reasons mostly) that have higher priority here: album_grouping_key, labelid. If you add them with some uniquie content for at least one album that would work as well. Of course that's a bit more work if you already have your library organized with one album per directory. But at least it works.

> I have basically two versions of my library: one for archival storage and one to sync with a mobile device. I'd like to manage them both in quodlibet at the same time. In the setup #2193 would be not great help as you had to do the distinction manually.

One hack would be to have two instances where one is started with "QUODLIBET_USERDIR=<somepath> quodlibet"

declension:
> I have basically two versions of my library: one for archival storage and one to sync with a mobile device. I'd like to manage them both in quodlibet at the same time.

Actually, see also related #202 for exactly this

lazka:
I'm not willing to add another option for this -> closing.

I'll keep your use case in mind for #2193

PR #2206, opened 2017-01-24
French translation update

trebmuh:
Hope that helps.

### DIFF
--- a/quodlibet/po/fr.po
+++ b/quodlibet/po/fr.po
@@ -1,12 +1,14 @@
 # translation of fr.po to Français
 # Olivier Gambier <dev@viapanda.com>, 2006.
+# Ludovic DRUETTE <ludovicdruette@gmail.com>, 2016.
+# Olivier Humbert <trebmuh@tuxfamily.org>, 2017.
 msgid ""
 msgstr ""
 "Project-Id-Version: fr\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2016-12-10 11:49+0100\n"
-"PO-Revision-Date: 2016-05-04 15:34+0200\n"
-"Last-Translator: Ludovic DRUETTE <ludovicdruette@gmail.com>\n"
+"PO-Revision-Date: 2017-01-24 01:06+0100\n"
+"Last-Translator: Olivier Humbert <trebmuh@tuxfamily.org>\n"
 "Language-Team: Français\n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
@@ -83,9 +85,8 @@ msgstr "_Notation"
 
 #: ../quodlibet/browsers/albums/main.py:178
 #: ../quodlibet/browsers/covergrid/main.py:63
-#, fuzzy
 msgid "Sort _by…"
-msgstr "_Trier par..."
+msgstr "_Trier par…"
 
 #: ../quodlibet/browsers/albums/main.py:199
 #: ../quodlibet/browsers/covergrid/main.py:84
@@ -133,9 +134,8 @@ msgstr[1] "Recharger les _jaquettes des albums"
 #: ../quodlibet/browsers/iradio.py:852 ../quodlibet/browsers/paned/main.py:292
 #: ../quodlibet/browsers/paned/main.py:298 ../quodlibet/browsers/search.py:125
 #: ../quodlibet/browsers/soundcloud/main.py:312 ../quodlibet/util/library.py:36
-#, fuzzy
 msgid "browsers"
-msgstr "Navigateurs"
+msgstr "navigateurs"
 
 #: ../quodlibet/browsers/albums/prefs.py:28
 #: ../quodlibet/browsers/collection/models.py:17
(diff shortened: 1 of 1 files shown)


quodlibet/po/fr.po
```diff
@@ -769,29 +752,28 @@ msgid "_Eject"
...
 #: ../quodlibet/browsers/playlists/main.py:458
 msgid "_Rename"
 msgstr "Re_nommer"
 
 #: ../quodlibet/browsers/media.py:461
-#, fuzzy, python-format
+#, python-format
 msgid "%(used-size)s used, %(free-size)s available"
-msgstr "<b>%s</b> utilisé(s), <b>%s</b> disponible(s)"
+msgstr "%(taille-utilisée)s utilisé(es), %(taille-libre)s disponible(s)"
```
lazka: The text in `%(...)s` should not be translated

lazka:
Thanks!

quodlibet/po/fr.po
```diff
@@ -455,14 +447,12 @@ msgid "Cover Grid Preferences"
...
-#, fuzzy
 msgid "Show album _text"
-msgstr "Montrer les _jaquettes des albums"
+msgstr "Montrer les _textes des albums"
 
 #: ../quodlibet/browsers/covergrid/prefs.py:71
-#, fuzzy
 msgid "Show \"All Albums\" Item"
-msgstr "Tous les albums"
+msgstr "Monter tous les éléments \"Tous les albums\""
```
pierremoreau: "Monter" -> "Montrer"

I need to have a look at this message in its context, but it does sound weird like that. The message in English does not talk about multiple items, so I would go with "Montrer l'élément \"Tous les albums\"", but…

quodlibet/po/fr.po
```diff
@@ -426,9 +419,8 @@ msgid "Merge"
 msgstr ""
 
 #: ../quodlibet/browsers/collection/prefs.py:188
-#, fuzzy
 msgid "Album Collection Preferences"
-msgstr "Préférences de l'affichage liste d'albums"
+msgstr "Préférences de la collection d'album"
```
pierremoreau: I would write "albums", as "collection de" or "liste de" will always be followed by a plural.

quodlibet/po/fr.po
```diff
@@ -523,9 +513,8 @@ msgid "Internet Radio"
 msgstr "Radio Internet"
 
 #: ../quodlibet/browsers/iradio.py:252
-#, fuzzy
 msgid "Downloading station list"
-msgstr "Télécharger les fichiers"
+msgstr "Télécharger la liste de station"
```
pierremoreau: "[…] liste des stations"

quodlibet/po/fr.po
```diff
@@ -929,28 +908,25 @@ msgstr[1] ""
 "Voulez-vous continuer?"
 
 #: ../quodlibet/browsers/playlists/menu.py:79
-#, fuzzy, python-format
+#, python-format
 msgid "Confirm action for playlist \"%s\""
-msgstr "Confirmez la suppression de la liste de lecture"
+msgstr "Confirmez l'action pour liste de lecture\"%s\""
```
pierremoreau: There is a space missing between "lecture" and "\"%s\"". I would also add a "la": "[…] l'action pour **la** liste de […]"

quodlibet/po/fr.po
trebmuh: Thanks for the hint. Corrected. I'm pushing the fix in a few minutes.

quodlibet/po/fr.po
trebmuh: Thanks for the hint. Fixed. I'm pushing the fix in a few minutes.

quodlibet/po/fr.po
trebmuh: Indeed. Fixed. I'm pushing the fix in a few minutes.

quodlibet/po/fr.po
trebmuh: Not sure about this one since the EN version is not in a plural form. I think you're point of view make sense even if both are fine (referring to the "quantity" or to the "quality" are both fine in French, and I guesstimate that it's the same in English here). If we'd go to changing to plural, it would first need a fix in the original EN version first since I don't like moving out the original when translating.

quodlibet/po/fr.po
trebmuh: Good eyes. Thanks for the hints. Fixed. I'm pushing this in a few seconds.

trebmuh:
@pierremoreau and @lazka : I've pushed the fixes in regards of your comments. Thanks for your help.

quodlibet/po/fr.po
declension: I think English is different though here - we use singular forms for the first noun in compound nouns typically (_car parks_, _boat races_, _meeting notes_ etc)

trebmuh: "stations' list" could have been used as well, isn't ?

declension: @trebmuh - People would work out what it meant, but it'd never be written like that idiomatically.

trebmuh: Oh, strange, I've been seeing that written a numerous time when I was working in a council during 3 years in Australia.

Anyway, fix pushed.

trebmuh:
Last fix pushed.

pierremoreau:
@trebmuh Thank you for your work!

lazka:
Thanks!

trebmuh:
All good, thanks for the quick push and the help people.

Issue #2255, opened 2017-02-21
Allow higher granularity tickrate [enhancement]


Meriipu:
Allow higher granularity tickrate, e.g updating seekbar every 0.2 seconds.

Meriipu:
from some trying around, this replacement in qltk/tracker.py works for the waveform seek bar, but not the other two. I have no idea if it has any negative side effects beyond slightly higher cpu usage when the player is not iconified.
```
                #self.__id = GLib.timeout_add_seconds(1, self.__update)
                self.__id = GLib.timeout_add(250, self.__update)
```

lazka:
Why do you need/want it?

Meriipu:
It was purely for cosmetic purposes. I was looking at the waveform seek bar plugin, and I felt the movement in seconds was a bit jaggy.

PR #2261, opened 2017-02-22
waveformseekbar: remove hires configuration and add hidpi detection

### LINKED ISSUE #2258: Waveform: add a HiDPI mode
On my retina laptop, the waveform looks ugly.

I have code on my local repository that adds a third mode to the lowres/highres setting. This new mode enhances the number of values calculated and disables any averaging during rendering. The result is far more beautiful.

If you're OK with the idea, I'll a PR.

ptitjes:
Fixes #2258.

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -7,11 +7,12 @@
 # published by the Free Software Foundation
 
 from gi.repository import Gtk, Gdk, Gst
+import cairo
 from math import ceil, floor
 
 from quodlibet import _, app
 from quodlibet import print_w
-from quodlibet.plugins import PluginConfig, BoolConfProp, IntConfProp, \
+from quodlibet.plugins import PluginConfig, IntConfProp, \
     ConfProp
 from quodlibet.plugins.events import EventPlugin
 from quodlibet.qltk import Align
@@ -56,7 +57,7 @@ class WaveformSeekBar(Gtk.Box):
         self._tracker.tick()
 
         if player.info:
-            self._create_waveform(player.info, CONFIG.data_size)
+            self._create_waveform(player.info, CONFIG.max_data_points)
 
     def _create_waveform(self, song, points):
         # Close any existing pipelines to avoid warnings
@@ -123,7 +124,7 @@ class WaveformSeekBar(Gtk.Box):
 
     def _on_song_changed(self, library, songs, player):
         if player.info:
-            self._create_waveform(player.info, CONFIG.data_size)
+            self._create_waveform(player.info, CONFIG.max_data_points)
 
         self._waveform_scale.set_placeholder(True)
         self._update(player)
@@ -190,23 +191,29 @@ class WaveformScale(Gtk.EventBox):
         self.queue_draw()
 
     def draw_waveform(self, cr, width, height, elapsed_color, remaining_color):
-        line_width = CONFIG.line_width
+        scale_factor = self.get_scale_factor()
+        hidpi = scale_factor > 1
+        pixel_ratio = float(scale_factor)
+        line_width = 1.0 / pixel_ratio
+
         value_count = len(self._rms_vals)
         max_value = max(self._rms_vals)
-        ratio_width = value_count / float(width)
+        ratio_width = value_count / float(width) / pixel_ratio
         ratio_height = max_value / float(height) * 2
         half_height = height // 2
         cr.set_line_width(line_width)
+        cr.set_line_cap(cairo.LINE_CAP_ROUND)
+        cr.set_line_join(cairo.LINE_JOIN_ROUND)
 
-        if line_width < 2:
+        if not hidpi:
             # Default AA looks bad (and dimmer) for all 1px shapes.
             cr.set_antialias(1)
 
-        position_width = self.position * width
+        position_width = self.position * width * pixel_ratio
         hw = line_width / 2.0
         # Avoiding object lookups is slightly faster
         data = self._rms_vals
-        for x in range(0, width, line_width):
+        for x in range(0, int(width * pixel_ratio), 1):
             fg_color = (elapsed_color if x < position_width
                         else remaining_color)
             cr.set_source_rgba(*list(fg_color))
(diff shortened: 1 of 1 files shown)


declension:
Thanks. This however breaks rendering behaviour for quiet sections on all waveforms:

ql-2261-2017-02-24-quiet-waveform

vs current master:
ql-master-2017-02-24-quiet-waveform

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -218,8 +224,8 @@ def draw_waveform(self, cr, width, height, elapsed_color, remaining_color):
                    if u1 != u2 else 0.5)
 
             # Draw single line, ensuring 1px minimum
-            cr.move_to(x, half_height - val)
```
declension: Losing this 1px behaviour probably causes the problems rendering.

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -190,23 +191,28 @@ def reset(self, rms_vals, player):
         self.queue_draw()
 
     def draw_waveform(self, cr, width, height, elapsed_color, remaining_color):
-        line_width = CONFIG.line_width
+        hidpi = self.get_scale_factor() == 2
```
declension: This should probably be more lenient for future even-higher DPI (or non-OSX hidpi)
 e.g. `> 2` or maybe `> 1` even, not sure

ptitjes: I have no idea how GTK people will make this (integer) property evolve in the future. I'll test for > 1 then.

quodlibet/quodlibet/ext/events/waveformseekbar.py
ptitjes: Indeed. I'll fix that !

ptitjes:
@declension So I made the modification relative to the use of the value returned by `get_scale_factor()`.

However I am not able to reproduce your 1px problem...
Would you mind rechecking with this latest patch ?

Here's what the unstable version from lazca repository draws:
unstable

Here's what this PR draws in non-HiDPI mode:
patched

And here's what this PR draws in HiDPI mode:
patched-hidpi

declension:
Can't see any change here: 
waveform-2261-2017-02-24-2

ptitjes:
@declension Would you mind trying again, please ?

quodlibet/quodlibet/ext/events/waveformseekbar.py
```diff
@@ -215,11 +222,15 @@ def draw_waveform(self, cr, width, height, elapsed_color, remaining_color):
             u1 = max(0, int(floor((x - hw) * ratio_width)))
             u2 = min(int(ceil((x + hw) * ratio_width)), len(data))
             val = (sum(data[u1:u2]) / (ratio_height * (u2 - u1))
-                   if u1 != u2 else 0.5)
+                   if u1 != u2 else 0.0)
 
-            # Draw single line, ensuring 1px minimum
-            cr.move_to(x, half_height - val)
-            cr.line_to(x, ceil(half_height + val))
```
declension: What was wrong with the `ceil`? That works for me

declension:
Also, note: it looks like you're force-pushing rewritten commits. Now that this is public that can make life harder for people (re-)pulling, and loses the history of changes you're doing to the PR (i.e. _Commits: 1_ is a lie). 

Github allows the _merger_ to squash these commits if it suits the project, so perhaps it's best to push normal commits - with commit message to match the intent of the change.

quodlibet/quodlibet/ext/events/waveformseekbar.py
ptitjes: The thing is that doing a 'ceil' draws a line that is *at least* one pixel high. Say val=.25, you will draw a line of 1.25px high.

(BTW I highly suspect the reason of the anti-aliasing disabling in the code has been made for this reason. I think it could be removed once the drawing code has been cleaned.)

Not even to speak that if you have a resolution higher than the coordinate system (HiDPI, printing, ...), then using 'ceil' completely fails !

ptitjes:
I am sorry if that shocks you. I am used to do that for such a tiny PR and as is done in a patch/mail workflow. It enables you to see the difference to the actual code and not to my last diff. If it causes problems to you, I can proceed your way.

I only have HiDPI displays. So I need your help to test my patch on non-HiDPI devices so that it only brings enhancements to the current way. (That is if you will to help me.) Also, I think that adding conditionals to the drawing code would be a shame if we can find a cleaner way to do it whatever the resolution is. So I intend to try some more tweaks, if you will to help me test those.

declension:
@ptitjes ok. I don't have much time left here but have tested your latest code. Perhaps a VM would help here. 
Anyway I see no real change:
waveform-2261-2017-02-24-2

By adding 0.5 and -0.5 on (your) lines 229,230: a "fatter but symmetric zero" approach but probably my favourite, given that it's a seek bar too and there's no additional highlighting:
2261-waveform-2017-02-24 18-55-57

Or just removing the AA disabling:
pr-2261-2017-02-24 18-28-27

As you can see it allows the sub-pixel precision, but it looks pretty bad (which is why I disabled it)

ptitjes:
@declension OK. We happen to have different tastes :) But I think I found information in [Cairo's FAQ](https://www.cairographics.org/FAQ/#sharp_lines) that explains the problem of non-sharpness of the zero line.

I modified my code accordingly. I think it can fit both of us. Please tell me what you feel.

declension:
Thanks - looks better vertically yes but still blurred:
2261-waveform-2017-02-24 20-13-13

Useful Cairo link, thanks;  Adding 0.5 to the horizontal fixed the remaining (horizontal) AA problem... :smile: 

2261-waveform-0 5px-2017-02-24 20-14-27

ptitjes:
@declension On the `line_to` and `move_to` lines ?

ptitjes:
@declension OK, so my last commit should not handle that, whatever the DPI scale factor is.

ptitjes:
If/When this PR gets accepted, I've got those two following commits to PR:

https://github.com/ptitjes/quodlibet/commit/0f26f843dba9e7e773d978c7dc170d7006e08f23:  waveformseeker: enhance redraw performance and shorten interval

* add a mechanism to update TimeTracker's interval
* enhance waveform redraw by using Cairo's clip rectangle
* shorten redraw interval to make the position travel smoothly
 
https://github.com/ptitjes/quodlibet/commit/4756bda4d0c815dbb041cf9f0e671ee789fa8956:  waveformseeker: use separate time trackers for label and waveform

Fixes #2255.

* use separate time trackers for label and waveform
* do less updates depending on the incoming events

declension:
LGTM, though can only test on HiDPI currently:

ql-2261-waveform-2017-03-01

Aside: we should probably scale the height of this in HiDPI actually, looks pretty small relative to the text

declension:
Thanks

ptitjes:
Thank you.

PR #2247, opened 2017-02-18
preview: add a preview player and a simple gui to preview songs

ptitjes:
Added a Backend class that have the responsibility to instantiate
and handle the existing main player and the new preview player.

Please know that it is my first in python. So please comment also on code quality.

PR #2262, opened 2017-02-23
seekbar: add a seek bar provider class

ptitjes:
This is necessary if one wants to have multiple players and seek bars.

### DIFF
--- a/quodlibet/quodlibet/ext/events/seekbar.py
+++ b/quodlibet/quodlibet/ext/events/seekbar.py
@@ -13,6 +13,7 @@ from quodlibet import _
 from quodlibet import app
 from quodlibet.plugins.events import EventPlugin
 from quodlibet.qltk import Icons
+from quodlibet.qltk.quodlibetwindow import SeekBarProvider
 from quodlibet.qltk.seekbutton import TimeLabel
 from quodlibet.qltk.tracker import TimeTracker
 from quodlibet.qltk import Align
@@ -125,11 +126,15 @@ class SeekBarPlugin(EventPlugin):
     PLUGIN_ICON = Icons.GO_JUMP
 
     def enabled(self):
-        self._bar = SeekBar(app.player, app.librarian)
-        self._bar.show()
-        app.window.set_seekbar_widget(self._bar)
+        self._provider = SeekBarProviderImpl()
+        app.window.set_seekbar_provider(self._provider)
 
     def disabled(self):
-        app.window.set_seekbar_widget(None)
-        self._bar.destroy()
-        del self._bar
+        self._provider.destroy_seekbar_widgets()
+        del self._provider
+
+
+class SeekBarProviderImpl(SeekBarProvider):
+
+    def do_create_seekbar_widget(self, player, librarian):
+        return SeekBar(player, librarian)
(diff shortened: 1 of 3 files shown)


ptitjes:
@lazka Can you review this PR ? Do you find the design too contrived ? Or maybe having a second player (currently) falls out of scope of QL ?

lazka:
Will have a look in the evening

ptitjes:
@lazka I just rebased it on master.

declension:
Hmm this feels quite Java-esque, at least in language (`Impl`s, Providers). It could be made more Pythonic definitely - but some larger questions arise first perhaps

1. To me the biggest current weirdness isn't solved still - that as a user you can select both "seek bar plugins" and have only one displayed (depending on order).  This mutual exclusivity would be ideally communicated to the user, e.g. by disallowing selection of both plugins in the UI, or warning on plugin overlap (c.f. Compiz etc)... unless we want to allow multiple widgets at once, but that's a bit weird here - and also not currently supported.
2. For multiple _players_ (not just seekbars) I think there are much bigger questions for QL that need answering first...

ptitjes:
@declension Well, I come from languages, indeed such as Java, where we use names with a lot of emphasis on the functionality. For the `Impl`, I had no choice because the basic seekbar was called `SeekBar`. I would rather have called it `BasicSeekBar` which would have not made its corresponding provider name clash. As for the name `Provider`, that the job it does.

That said... This PR was not made to handle any problem of mutual exclusivity between seek bar plugins. From what I've seen, there is no plugin dependency mechanism, the plugin support and handling are really basic, and plugins can only implement one extension and can't define new extension points. I would be glad if there some discussion somewhere of where you want to go first on these items. I've nothing yet while digging in the issue list.

This PR was really on the ability to instantiate multiple waveform seek bars.

> For multiple players (not just seekbars) I think there are much bigger questions for QL that need answering first...

Then what are those questions and where can we discuss those ?

I've tried to launch some, in addition to this PR, with little success up until now :

- Pull Request #2247: preview: add a preview player and a simple gui to preview songs (which can still be commented on even if I decided to try and put that into some plugins)

- Issue #2273: gui: modularize main window's user interface

ptitjes:
BTW, in my opinion, everything should be provided through extension point APIs and plugins (from Backends to Browsers, ...)

lazka:
Imo that just adds another layer for not much gain.

Now that we have two seek bars we might want to replace set_seekbar_widget() with add/remove_seekbar_widget() and show both in the UI if both are active. While notbody would want that it makes it clear to the user what's going on when both are active.

Regarding extension APIs: As long as there are only one or two plugins needing an API I don't mind if they just use the internal APIs and change internal widgets. If we'd add plugin API for every use case things would get complicated fast. I don't expect there to be any more seek bar implementations.

ptitjes:
> Imo that just adds another layer for not much gain.

My problem is that the plugins instantiate the seek bar widget.
Could we then give the class of the seek bar widget rather than the instance ?

> Now that we have two seek bars we might want to replace set_seekbar_widget() with add/remove_seekbar_widget() and show both in the UI if both are active. While notbody would want that it makes it clear to the user what's going on when both are active.

I agree. This removes the problem that highlighted @declension.

> Regarding extension APIs: As long as there are only one or two plugins needing an API I don't mind if they just use the internal APIs and change internal widgets. If we'd add plugin API for every use case things would get complicated fast. I don't expect there to be any more seek bar implementations.

Well, if you are talking about what I propose in #2273 then look at my last comment there and you'll see it fixes quite a number of use-cases (from different issues). Also it reduces the number of plugin APIs.

declension:
Closing old MRs (#3655)

PR #2263, opened 2017-02-23
plugin: add a plugin to toggle the menubar's visibility

ptitjes:
By pressing the Alt key you can toggle the menubar's visibility (ala Firefox, Atom, ...)

### DIFF
--- a/quodlibet/quodlibet/ext/events/toggle_menu.py
+++ b/quodlibet/quodlibet/ext/events/toggle_menu.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Christoph Reiter
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation
+
+from gi.repository import Gdk
+
+from quodlibet import _
+from quodlibet import app
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.qltk import Icons
+
+
+class ToggleMenuBarPlugin(EventPlugin):
+    PLUGIN_ID = "ToggleMenuBar"
+    PLUGIN_NAME = _("Toggle Menu Bar")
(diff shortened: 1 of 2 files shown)


declension:
Hmm, more relevantly though: this behaviour already works without the plugin for me (Ubuntu 16.10, GTK 3.20.9). 

Is this an OS X only problem?

ptitjes:
I could not tell because I don't use OS X. I run my Mac on Fedora 25, and I don't have the menu bar toggle... Hence my plugin :) Is this only on Ubuntu ? Can you see it on other GTK applications ?

quodlibet/quodlibet/ext/events/toggle_menu.py
```diff
@@ -0,0 +1,38 @@
...
+    PLUGIN_ID = "ToggleMenuBar"
+    PLUGIN_NAME = _("Toggle Menu Bar")
+    PLUGIN_DESC = _("Toggle the menu bar by pressing the Alt key.")
+    PLUGIN_ICON = Icons.EDIT
+
+    def enabled(self):
+        window = app.window
+
+        # Maybe this should be made directly accessible
+        menubar = window.get_children()[0].get_children()[0]
```
lazka: Good enough for a one time thing I guess.

lazka:
> Hmm, more relevantly though: this behaviour already works without the plugin for me (Ubuntu 16.10, GTK 3.20.9).

This is only the case for Ubuntu +  Unity with their patched gtk.

ptitjes:
Thanks!

lazka:
A possible enhancement would be to listen to click events if shown and see if they are not on the menu or any submenu and then hide the menu again. that would more closely follow what for example firefox does. But I'm not sure how hard that is to implement.

ptitjes:
Oh yes indeed. I never did pay attention firefox/atom did that...
I'll think about it.

PR #2264, opened 2017-02-23
plugin: add a tap bpm plugin

ptitjes:
This is a very little plugin that allows the user to tap the BPM for the selected song and save it in its tags.

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/tapbpm.py
+++ b/quodlibet/quodlibet/ext/songsmenu/tapbpm.py
@@ -6,7 +6,7 @@
 # it under the terms of the GNU General Public License version 2 as
 # published by the Free Software Foundation
 
-from gi.repository import Gtk, GObject
+from gi.repository import Gdk, Gtk
 
 from quodlibet import _
 from quodlibet.plugins.songsmenu import SongsMenuPlugin
@@ -27,6 +27,7 @@ class TapBpmPanel(Gtk.VBox):
 
         box = Gtk.HBox()
         box.set_spacing(6)
+        # TRANSLATORS: BPM mean "beats per minute"
         box.pack_start(Gtk.Label(_("BPM:")), False, True, 0)
         self.bpm_label = Gtk.Label(_("n/a"))
         self.bpm_label.set_xalign(0.5)
@@ -40,6 +41,7 @@ class TapBpmPanel(Gtk.VBox):
 
         self.tap_btn = Gtk.Button(label=_("Tap"))
         self.tap_btn.connect('button-press-event', self.tap)
+        self.tap_btn.connect('key-press-event', self.key_tap)
         self.pack_start(self.tap_btn, True, True, 0)
 
         self.init_tap()
@@ -48,9 +50,6 @@ class TapBpmPanel(Gtk.VBox):
         self.show_all()
 
     def update(self):
-        GObject.idle_add(self.do_update)
-
-    def do_update(self):
         has_new_bpm = self.floating_bpm != self.original_bpm
 
         self.dialog.set_response_sensitive(Gtk.ResponseType.OK, has_new_bpm)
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/ext/songsmenu/tapbpm.py
```diff
@@ -0,0 +1,193 @@
...
+        box.pack_start(self.bpm_label, True, True, 0)
+
+        self.reset_btn = Gtk.Button(label=_("Reset"))
+        self.reset_btn.connect('clicked', lambda *x: self.reset())
+        box.pack_end(self.reset_btn, False, True, 0)
+
+        self.pack_start(box, False, True, 0)
+
+        self.tap_btn = Gtk.Button(label=_("Tap"))
+        self.tap_btn.connect('button-press-event', self.tap)
```
lazka: 'clicked' event, otherwise pressing space/enter wont do anything if the button is focused.

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
```diff
@@ -0,0 +1,193 @@
...
+        self.dialog = parent
+        self.song = song
+        self.original_bpm = float(song["bpm"]) if "bpm" in song else 0.0
+
+        self.set_margin_bottom(6)
+        self.set_spacing(6)
+
+        box = Gtk.HBox()
+        box.set_spacing(6)
+        box.pack_start(Gtk.Label(_("BPM:")), False, True, 0)
```
lazka: add a translators comment above:

```
# TRANSLATORS: BPM measn "beats per minute"
```

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
```diff
@@ -0,0 +1,193 @@
...
+        self.tap_btn.connect('button-press-event', self.tap)
+        self.pack_start(self.tap_btn, True, True, 0)
+
+        self.init_tap()
+        self.update()
+
+        self.show_all()
+
+    def update(self):
+        GObject.idle_add(self.do_update)
```
lazka: Why the idle_add?

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
```diff
@@ -0,0 +1,193 @@
...
+    def response(self, win, response):
+        self._window.hide()
+        self._window.destroy()
+        del self._window
+
+        if response != Gtk.ResponseType.OK:
+            return
+
+        # Save metadata
+        self._panel.save()
```
lazka: Not a bug, but try to save before destroying the window. After destroy() widgets shouldn't be used anymore and while it currently doesn't use it in save(), it might in the future.

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
ptitjes: It won't work without it, I think because we are in modal dialog.

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
ptitjes: Sure !

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
ptitjes: I used because when you tap with a mouse or a pad, the clicked comes at release of the button. Whereas the user (well me) expects to tap at press. It is really not nice when tap is triggered by release, because not only you have to tap in rhythm but also kinda press the button for the same duration every time (not simple, I tried :p).

Maybe I should document it so that not one wonders in the future.

I could add an additional trigger for key-presses when the button is focused...

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
ptitjes: Indeed, I'll modify that !

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
lazka: What doesn't work? It seems to work the same here as far as I can see.

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
lazka: I see, let's leave it that way then for now..

quodlibet/quodlibet/ext/songsmenu/tapbpm.py
ptitjes: If call `do_update` directly in `update`, the label is never updated.

ptitjes: OK, forget it you're right ! :)
I'll change that.

ptitjes:
@lazka Changes made as requested. I kept using "button-press-event" but added code to handle space/return key presses.

lazka:
Thanks

ptitjes:
Thank you !

PR #2267, opened 2017-02-25
togglemenubar: add auto-hide mechanism

ptitjes:
That is the best I can do for now.

The only functional difference with what Atom/Firefox do is that they require a first click to tear down a menu. Whereas this PR tears down menus as soon as the pointer enters the menu bar.

Any ideas to do better ?


lazka:
Looks good to me. Do you think it's an improvement over the simple alt toggle?

ptitjes:
Yes, but a little bit strange that the menus tear down as soon as you hover over them.
Did you try it ?

lazka:
Yeah, I tried. I think it's because select_first() activates an event grab for the menu, which would normally happen on the first click on the menu.

lazka:
Thanks!

PR #2270, opened 2017-02-25
songlist: fix moving elements via drag-n-drop (PY3)

### LINKED ISSUE #2268: Queue: Moving elements fails with PY3
While moving elements inside the queue works correctly with python 2, it fails with python 3.

It produces the following error

```
  File "/home/didier/Code/external/quodlibet/quodlibet/quodlibet/qltk/songlist.py", line 329, in __drag_data_received
    model.move_before(self.__drag_iters.pop(0), iter)
AttributeError: 'map' object has no attribute 'pop'
```

or

```
  File "/home/didier/Code/external/quodlibet/quodlibet/quodlibet/qltk/songlist.py", line 332, in __drag_data_received
    model.move_after(self.__drag_iters.pop(), iter)
AttributeError: 'map' object has no attribute 'pop'
```

depending on whether you move the elements before or after themselves.

This boils down from the fact that 'map' results are not lists anymore.
Thus changing line 275 of file `songlist.py` from
```
self.__drag_iters = map(model.get_iter, paths)
```
to
```
self.__drag_iters = list(map(model.get_iter, paths))
```

I could do a PR but I am not sure it is the best fix.

ptitjes:
Fixes #2268.

### DIFF
--- a/quodlibet/quodlibet/qltk/songlist.py
+++ b/quodlibet/quodlibet/qltk/songlist.py
@@ -272,7 +272,7 @@ class SongListDnDMixin(object):
 
             qltk.selection_set_songs(sel, songs)
             if ctx.get_actions() & Gdk.DragAction.MOVE:
-                self.__drag_iters = map(model.get_iter, paths)
+                self.__drag_iters = list(map(model.get_iter, paths))
             else:
                 self.__drag_iters = []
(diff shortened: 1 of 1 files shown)


lazka:
Thanks

PR #2275, opened 2017-02-25
replaygain: save selected replaygain profiles to config

### LINKED ISSUE #2162: Persist Replay Gain mode
The Replay Gain Mode determines whether volume is normalized [per track or per album](https://en.wikipedia.org/wiki/ReplayGain#Track-gain_and_album-gain) (so for the latter the differences in loudness in an album stay intact). It can be set when right clicking the loudspeaker like this:

screenshot from 2016-12-21 23-04-08

I personally prefer the Track Mode, especially since currently the queue seems to use album gain.

_Slightly related:_ It could also be interesting to have a slider there to decide how much you value differences in loudness per track vs. within an album.

ptitjes:
Fixes #2162.

* Add a method to reload profiles from config
* Add a method to update and save a profile
* Add a utility config.unset(...) method

### DIFF
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -203,6 +203,7 @@ setstringlist = _config.setstringlist
 getlist = _config.getlist
 setlist = _config.setlist
 set = _config.set
+unset = _config.unset
 settext = _config.settext
 write = _config.write
 reset = _config.reset
--- a/quodlibet/quodlibet/player/_base.py
+++ b/quodlibet/quodlibet/player/_base.py
@@ -100,6 +100,7 @@ class BasePlayer(GObject.GObject, Equalizer):
 
     def __init__(self, *args, **kwargs):
         super(BasePlayer, self).__init__()
+        self.reload_replaygain_profiles()
 
     def destroy(self):
         """Free resources"""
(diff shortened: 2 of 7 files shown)


lazka:
I think this is the wrong approach. replaygain_profiles contains mostly runtime dependent state and shouldn't be saved.

I'd suggest to save/restore the menu selection instead.

quodlibet/quodlibet/config.py
```diff
@@ -203,6 +203,7 @@ def _config_text(text):
 getlist = _config.getlist
 setlist = _config.setlist
 set = _config.set
+unset = _config.unset
```
lazka: There is a "reset()"

quodlibet/quodlibet/player/_base.py
```diff
@@ -133,15 +149,12 @@ def calc_replaygain_volume(self, volume):
             scale = 1
         return volume * scale
 
-    def _reset_replaygain(self):
```
lazka: this method gets overridden by the implementations

quodlibet/quodlibet/config.py
ptitjes: Well, `reset()` resets to the value in `config.py` whereas `unset()` remove the entry in the configuration file.

Anyway, I won't need it anymore if I don't store the values but just the menu selection...

quodlibet/quodlibet/player/_base.py
ptitjes: Oups, I should not trust my IDE's "Find usage" in Python...

quodlibet/quodlibet/config.py
lazka: Ah, true. I wonder why I implemented it that way..

ptitjes:
> I think this is the wrong approach. replaygain_profiles contains mostly runtime dependent state and shouldn't be saved.
>
> I'd suggest to save/restore the menu selection instead.

Ah OK. Then I did not understand their use :)
So I'll just store the mode name selected in the menu.

ptitjes:
@lazka The issue is labelled "player". However, if I only store the mode of the menu selection then I will do that at the level of the `VolumeMenu`, and one can programmatically change that by updating `player.replaygain_profile[0]`. Is that OK for you ?

lazka:
Sounds good.

(As you may have noticed, replaygain_profile is a bit hacky and all over the place...)

ptitjes:
Well, I volunteer to fix that hackyness if you open a ticket to explain how it should ideally work :)

In the meantime, I'll store the mode from the `VolumeMenu` as said.

ptitjes:
Replaced by PR #2279.

PR #2279, opened 2017-02-27
replaygain: save selected replaygain profiles to config (take 2)

### LINKED ISSUE #2162: Persist Replay Gain mode

ptitjes:
Fixes #2162.

Only the mode name is saved. Profile is set according to that name.

Replaces PR #2275.

### DIFF
--- a/quodlibet/quodlibet/qltk/controls.py
+++ b/quodlibet/quodlibet/qltk/controls.py
@@ -123,6 +123,10 @@ class VolumeMenu(Gtk.Menu):
         self.append(item)
         item.show()
 
+        # Set replaygain mode as saved in configuration
+        replaygain_mode = config.gettext("player", "replaygain_mode", "auto")
+        self.__set_mode(player, replaygain_mode)
+
         rg = Gtk.Menu()
         rg.show()
         item.set_submenu(rg)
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/qltk/controls.py
```diff
@@ -131,15 +135,20 @@ def __init__(self, player):
...
-            item.connect("toggled", self.__changed, player, profile)
-            if player.replaygain_profiles[0] == profile:
+            item.connect("toggled", self.__changed, player, mode)
+            if replaygain_mode == mode:
                 item.set_active(True)
             item.show()
 
-    def __changed(self, item, player, profile):
+    def __set_mode(self, player, mode):
+        player.replaygain_profiles[0] = \
```
lazka: Please handle unknown mode values. Crashing in this case makes it hard to extend them and switching between QL versions.

(In general, code should handle garbage config values in some "sane" way if possible)

ptitjes: Rhaaa, of course !! Will fix that...

ptitjes:
@lazka Done. As stated in the commit message, I chose to not update the configuration but just fallback to the 'auto' mode. Tell me if that fits the bill !

lazka:
Thanks

Issue #1865, opened 2016-02-29
Make infos next to player-icons clickable


timpulver:
Would be great if we could click on an info-item next to the player-icons to browse the album / label / artist.

<img width="342" alt="bildschirmfoto 2016-02-29 um 13 31 49" src="https://cloud.githubusercontent.com/assets/1055819/13394545/670a29c4-dee9-11e5-8d6e-042d88928b3a.png">

lazka:
What would clicking do exactly?

timpulver:
Good question. An ideal outcome would be if:

- Clicking the *artist* – would show all tracks by the artist in the browser
- Clicking the *album* – would show all tracks in the browser
- Clicking the *label* – would show all tracks in the browser

The difficulty here probably is that there are so many different browsers…

timpulver:
Also it would be great if filters would be visible in the paned browser – so e.g. when selecting `artist=Kelela`it would be great if there was a search for a column `artist` or `involved artists` in the paned browser and then the column filter could be set accordingly.

lazka:
OK. Not sure how easy this is (as the display is based on the user defined pattern), but might be worth a try.

CreamyCookie:
I think this could be interesting and cool, as long as it would still be easy to select and copy information from the display into the clipboard, without triggering the filter.

declension:
More food for thought: an incoming URI schema (I started pulling that thread in #1828 a bit actually), in the current trend for (mobile) app deep-linking.

e.g. `quodlibet://search?song=foobar` or `quodlibet://browse/paned/?artist=foobar`
...or some better scheme

declension:
This would also (especially) make sense in the context of #940

PR #2280, opened 2017-02-27
plugins: enable links navigation in plugin descriptions

ptitjes:
* add a `show_uri(label, uri)` function to dispatch a uri
* add a `move_to(plugin_id)` method to `PluginWindow`
* remove automatic escaping of plugin descriptions
* show uri for links inside plugin descriptions
* use this mechanism for `SqueezeboxPlaylistPlugin`
* correctly escape existing plugin descriptions
* add a convenience `clear()` method to `ClearEntryMixin`

The reason for doing that is to ease the configuration of plugins that are dependent one to another. The Squeezebox set of plugins is an example of that. Also, I'd like to release my secondary (aka. preview) player code as two plugins (one that enables/provides the secondary player with configuration for its gst pipeline, and one that adds a "Pre-listen" `SongsMenu` item).

The `show_uri` function has been put in `qltk.__init__.py` so as to enable any part of the user interface to link to plugin preferences. You can do it with the following code:
```
label = Gtk.Label()
label.set_markup("<a href=\"ql:plugins:My Plugin Id\">My plugin</a>")
label.connect("activate-link", show_uri)
```

Ultimately, one can later add an if-branch in `show_uri` to handle `"ql:prefs:"` uri prefix to link to some preferences UI element.

### DIFF
--- a/quodlibet/quodlibet/qltk/__init__.py
+++ b/quodlibet/quodlibet/qltk/__init__.py
@@ -27,7 +27,7 @@ def show_uri(label, uri):
     specific one.
 
     Currently handled quodlibet uris:
-        - ql:plugins:<plugin id>
+        - quodlibet:///prefs/plugins/<plugin id>
 
     Args:
         label (str)
--- a/quodlibet/quodlibet/qltk/__init__.py
+++ b/quodlibet/quodlibet/qltk/__init__.py
@@ -44,8 +44,12 @@ def show_uri(label, uri):
         else:
             return __show_quodlibet_uri(parsed)
     else:
-        from quodlibet import app
-        return Gtk.show_uri_on_window(app.window, uri, 0)
+        # Gtk.show_uri_on_window exists since 3.22
+        if hasattr(Gtk, "show_uri_on_window"):
+            from quodlibet.qltk import get_top_parent
+            return Gtk.show_uri_on_window(get_top_parent(label), uri, 0)
+        else:
+            return Gtk.show_uri(None, uri, 0)
 
 
 def __show_quodlibet_uri(uri):
--- a/quodlibet/quodlibet/qltk/__init__.py
+++ b/quodlibet/quodlibet/qltk/__init__.py
@@ -24,7 +24,7 @@ from quodlibet.util import gdecode, print_d, print_w
 def show_uri(label, uri):
     """Shows a uri. The uri can be anything handled by GIO or a quodlibet
     specific one.
-    
+
     Currently handled quodlibet uris:
         - ql:plugins:<plugin id>
 
@@ -44,6 +44,7 @@ def show_uri(label, uri):
     else:
         from quodlibet import app
         return Gtk.show_uri_on_window(app.window, uri, 0)
+
 
 def get_fg_highlight_color(widget):
     """Returns a color useable for highlighting things on top of the standard
--- a/quodlibet/quodlibet/qltk/__init__.py
+++ b/quodlibet/quodlibet/qltk/__init__.py
@@ -21,6 +21,30 @@ from senf import fsn2bytes, bytes2fsn
 from quodlibet.util import gdecode, print_d, print_w
 
 
+def show_uri(label, uri):
+    """Shows a uri. The uri can be anything handled by GIO or a quodlibet
+    specific one.
+    
+    Currently handled quodlibet uris:
+        - ql:plugins:<plugin id>
+
+    Args:
+        label (str)
+        uri (str) the uri to show
+    Returns:
+        True on success, False on error
+    """
+
+    if uri.startswith("ql:"):
+        if uri.startswith("ql:plugins:"):
+            from .pluginwin import PluginWindow
+            return PluginWindow().move_to(uri[len("ql:plugins:"):])
+        else:
+            return False
+    else:
+        from quodlibet import app
+        return Gtk.show_uri_on_window(app.window, uri, 0)
+
 def get_fg_highlight_color(widget):
     """Returns a color useable for highlighting things on top of the standard
     background color.
--- a/quodlibet/quodlibet/qltk/__init__.py
+++ b/quodlibet/quodlibet/qltk/__init__.py
@@ -19,6 +19,7 @@ from gi.repository import GLib, GObject
 from senf import fsn2bytes, bytes2fsn
 
 from quodlibet.util import gdecode, print_d, print_w
+from quodlibet.compat import urlparse
 
 
 def show_uri(label, uri):
@@ -35,15 +36,24 @@ def show_uri(label, uri):
         True on success, False on error
     """
 
-    if uri.startswith("ql:"):
-        if uri.startswith("ql:plugins:"):
-            from .pluginwin import PluginWindow
-            return PluginWindow().move_to(uri[len("ql:plugins:"):])
-        else:
+    parsed = urlparse(uri)
+    if parsed.scheme == "quodlibet":
+        if parsed.netloc != "":
+            print_w("Unknown QuodLibet URL format (%s)" % uri)
             return False
+        else:
+            return __show_quodlibet_uri(parsed)
     else:
         from quodlibet import app
         return Gtk.show_uri_on_window(app.window, uri, 0)
+
+
+def __show_quodlibet_uri(uri):
+    if uri.path.startswith("/prefs/plugins/"):
+        from .pluginwin import PluginWindow
+        return PluginWindow().move_to(uri.path[len("/prefs/plugins/"):])
+    else:
+        return False
 
 
 def get_fg_highlight_color(widget):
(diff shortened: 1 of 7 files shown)


ptitjes:
Just stumbled upon #1865 which proposes a "quodlibet:" uri prefix...

lazka:
@declension Any thoughts one this? Since it includes squeeze box + plugin window work in #2218

declension:
On holiday now, so in brief:

`quodlibet://` seemed and seems like the best URL protocol to use t o me- and more importantly it's already in use by the Soundcloud auth flows (and there is some progress towards registering QL to handle these on an OS level - _works for me_ etc)

Ideally this change would integrate with that rather than a separate URI handling (showing, here, but maybe not always) process - and one that would allow us eventually to deep-link to specific parts of the app e.g. the plugin preferences pages.

ptitjes:
@declension I'm not sure I understood all of your comment, sorry.

As for the scheme, "quodlibet" makes sense. I'm totally OK with it.

As for the `//`, you made me doubt so I went to check the [URI specification](https://tools.ietf.org/html/rfc3986#appendix-A) again. The `//` should be present *if and only* there is an authority (a.k.a. host) before the hierarchical path part ([explanation](https://tools.ietf.org/html/rfc3986#section-3.3)). I believe that a host makes no sense for QL's URIs and then that the absolute plugins/prefs access URIs should have the following shape:
```
quodlibet:/plugins/MyPluginId
quodlibet:/prefs/MyPrefPageId
```
with the possibility to focus on a particular widget with an additional optional "#WidgetId" part, when it is possible in the future.

Do we agree on that ?

ptitjes:
@declension As for the rest of your comment, I don't understand much the relationship between SoundCloud and QuodLibet.

However, I believe your are saying that we will have the "quodlibet:" scheme registered at the OS level. Are you referring to Gnome ? Also, I think at some point we have to do the dispatch of the URI inside QL, right ? So once the "quodlibet:" scheme is recognized by the OS we get the control and have to process the hierarchical part. So we will be able to reuse that part from my `show_uri` function.

So I propose that I split this function in two:
- A function that is the internal "quodlibet:" URIs dispatch (`__show_quodlibet_uri`)
- A function that is the current `show_uri` that either calls `__show_quodlibet_uri` or fallback to Gtk as it does currently

Do we agree on that too ?

lazka:
> However, I believe your are saying that we will have the "quodlibet:" scheme registered at the OS level. Are you referring to Gnome ?

There is. ``xdg-open quodlibet://foo`` should open QL.

----

OK, so with multiple users of the quodlibet scheme we can have something like push those show_uri() calls to the global Application instance which can emit some "::show-uri" signal and interested parties can react to it. And we should namespace them. And we should be cautious, since that lets other programs/webbrowsers control QL..

@ptitjes You're right that the host part isn't needed, but we currently use "quodlibet://callbacks/soundcloud?code=" for soundcloud (https://github.com/quodlibet/quodlibet.github.io/blob/master/callbacks/soundcloud.html#L118). Is there any downside in just doing "quodlibet://<namespace>/<path>"? If no, please adjust to something like "quodlibet://pluginprefs/foo".

Unifying this can happen later..

ptitjes:
> There is. xdg-open `quodlibet://foo` should open QL.

Really nice !!

---

> emit some "::show-uri" signal and interested parties can react to it

I like that idea. That means we could handle those in plugins too.

---

> You're right that the host part isn't needed

Well the specification says:
> If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//").

The reason explained is a question of parsing rules for absolute vs. relative URIs. That said we could devise our own parsing rules of relative and absolute URIs, but I think it would be error-prone because new-comers and external people would naturally write our URIs the wrong way.

That said I can see from looking at the `cli.py` code and searching the web, that there seem to be no utility to parse URIs in Python. So we could indeed use "quodlibet:///" for absolute URIs and "quodlibet://" for relative URIs. (Even if I really find that sooo ugly... :))

So that gives `quodlibet:///prefs/plugin/PluginId` for plugins preferences ?

> Unifying this can happen later..

It won't be easy to change that later if there are lots of URIs everywhere...

lazka:
>  that there seem to be no utility to parse URIs in Python

``from urlparse import urlparse``

> So that gives quodlibet:///prefs/plugin/PluginId for plugins preferences ?

ok

> It won't be easy to change that later if there are lots of URIs everywhere...

If you think its easier in this PR, go ahead. Either way, please ask if you want a review.

ptitjes:
> Unifying this can happen later..

Did you mean unifying the way the URI is handled by `cli.py` and the current code of this PR ?

If yes, then I'd prefer to plug the dispatch mechanism (for which I created issue #2292) inside the `show_uri()` function later. If you are OK with that, I'll add a commit here to modify the URIs syntax and then ask you to review it.

lazka:
>  @lazka Did you mean unifying how the URI is handled by cli.py and the current code of this PR ?
>
> If yes, then I'd prefer to plug the dispatch mechanism (for which I created issue #2292) inside the show_uri() function later. (And also move the if "prefs/plugin/..." in a signal handler.)

yes

> If you are OK with that, I'll add a commit here to modify the URIs syntax and then ask you to review it.

ok

ptitjes:
@lazka Wasn't sure as how to treat the links inside translations. I think it is ready to review however.

quodlibet/quodlibet/qltk/__init__.py
```diff
@@ -19,6 +19,41 @@
...
+    parsed = urlparse(uri)
+    if parsed.scheme == "quodlibet":
+        if parsed.netloc != "":
+            print_w("Unknown QuodLibet URL format (%s)" % uri)
+            return False
+        else:
+            return __show_quodlibet_uri(parsed)
+    else:
+        from quodlibet import app
+        return Gtk.show_uri_on_window(app.window, uri, 0)
```
lazka: show_uri_on_window is 3.22+

```
if hasattr(Gtk, "show_uri_on_window"): ... else: Gtk.show_uri()
```

the window should be the one containing the label, use ``qltk.get_top_parent(label)``

quodlibet/quodlibet/qltk/entry.py
```diff
@@ -223,9 +223,15 @@ def enable_clear_button(self):
         self.set_icon_from_gicon(Gtk.EntryIconPosition.SECONDARY, gicon)
         self.connect("icon-release", self.__clear)
 
+    def clear(self):
+        self.__do_clear()
+
     def __clear(self, button, *args):
         # TODO: don't change the order.. we connect to clear and remove all
```
lazka: That comment should move to __do_clear as well

quodlibet/quodlibet/qltk/__init__.py
lazka: Alternative you can just make the return value "True if the uri was handled" and return False here. I guess "activate-link" falls back to show_uri() anyway if you return False in the handler (??)

quodlibet/quodlibet/qltk/__init__.py
```diff
@@ -19,6 +19,41 @@
...
 from quodlibet.util import gdecode, print_d, print_w
+from quodlibet.compat import urlparse
+
+
+def show_uri(label, uri):
+    """Shows a uri. The uri can be anything handled by GIO or a quodlibet
+    specific one.
+
+    Currently handled quodlibet uris:
+        - ql:plugins:<plugin id>
```
lazka: needs an update

quodlibet/tests/test_po.py
```diff
@@ -139,7 +139,11 @@ def test_punctuation(self):
         """
         fails = []
         regex = re.compile(r'\s[.,:;!?](?![a-z])|'
-                           r'[a-z](?<!people)[,:;][a-zA-Z]')
+                           r'[a-z]'
+                           r'(?<!people)'
+                           r'(?<!ql)'
+                           r'(?<!plugins)'
```
lazka: Is this change still needed?

quodlibet/quodlibet/qltk/__init__.py
ptitjes: I have to test it again. IIRC, if I return False, it just writes an error in the console.
(Hopefully, Gtk.show_uri() is 3.14 and we just moved to that version ! :))

quodlibet/tests/test_po.py
ptitjes: Well, I'm surprised it passed the CI. because it is still `ql`...

ptitjes: Oh no, you're right! Now there is a `/` after the colon.

ptitjes:
@lazka Thank you. I'll fix all that.

ptitjes:
@lazka Fixed.

lazka:
Thanks

PR #2289, opened 2017-03-02
waveformseekbar: Enhance the redraw performance and frequence

ptitjes:
I carefully enhanced the performance by using Cairo's clipping mechanism. Now, only the minimum required region of the waveform is queued for redraw.

Then the interval time between the redraws is computed depending on the width of the waveform and the length of the song. With a 2'50 song rendered on half of a MacBookPro 13", the redraw happens every ~250 milliseconds (4 times more than currently). On a full screen MacBookPro 13" it happens every ~90 milliseconds. Due to the careful handling of the clip rectangle, the occupation of the CPU shows no real difference.

### DIFF
--- a/quodlibet/quodlibet/qltk/tracker.py
+++ b/quodlibet/quodlibet/qltk/tracker.py
@@ -29,14 +29,22 @@ class TimeTracker(GObject.GObject):
     def __init__(self, player):
         super(TimeTracker, self).__init__()
 
+        self.__interval = 1000
         self.__player = player
         self.__id = None
         self.__stop = False
+        self.__reset = False
         self.__sigs = [
             player.connect("paused", self.__paused, True),
             player.connect("unpaused", self.__paused, False),
         ]
         self.__paused(player, player.paused)
+
+    def set_interval(self, interval):
+        """Update the resolution in milliseconds"""
+
+        self.__interval = interval
+        self.__reset = True
 
     def tick(self):
         """Emit a tick event"""
@@ -58,6 +66,11 @@ class TimeTracker(GObject.GObject):
             self.__source_remove()
             return False
 
+        if self.__reset:
+            self.__reset = False
+            self.__source_remove()
+            self.__paused(self.__player, self.__player.paused)
+
         self.tick()
         return True
 
@@ -69,7 +82,7 @@ class TimeTracker(GObject.GObject):
         else:
             self.__stop = False
             if self.__id is None:
-                self.__id = GLib.timeout_add_seconds(1, self.__update)
+                self.__id = GLib.timeout_add(self.__interval, self.__update)
 
 
 class SongTracker(object):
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/qltk/tracker.py
```diff
@@ -69,7 +82,7 @@ def __paused(self, player, paused):
         else:
             self.__stop = False
             if self.__id is None:
-                self.__id = GLib.timeout_add_seconds(1, self.__update)
+                self.__id = GLib.timeout_add(self.__interval, self.__update)
```
lazka: TimeTracker uses timeout_add_seconds() because that gets synced with all other timeout_add_seconds() calls (also from other apps) and allows us to wake up only once per second. 

So in case it is 1000 it should still call timeout_add_seconds().

ptitjes:
@lazka OK, I fixed the reuse of the existing waking up every seconds. Also, I fixed two bugs in my code.
I think it is good now, if you want to review my additions. Thanks.

lazka:
Thanks. Just one thing (I haven't looked at the drawing part), the invalidated area looks like its width is based on the elapsed time, is that on purpose?

out

ptitjes:
@lazka Thanks, was a big bug ! :)
BTW, how do you highlight the invalidated areas like that ? Can't find that on my search engine...

ptitjes:
OK, finally found it: `GTK_DEBUG=updates` in the environment.
So I confirm it now works as expected.

ptitjes:
There are still some full redraws that could be avoided when seeking through the song.
But still it is far better than what it was before.
I will enhance the redraw behaviours on seeking later.

ptitjes:
Well later is now...

redraws2

lazka:
:+1:

lazka:
If you press "ctrl+shift+i" while running QL, the gtk inspector will start where you can also enable visual updates etc..

ptitjes:
"ctrl+shift+i" doesn't trigger the inspector here !

ptitjes:
OK, to have Ctrl-Shift-D and Ctrl-Shift-I toggle the inspector, I needed that first:
```
gsettings set org.gtk.Settings.Debug enable-inspector-keybinding true
```

ptitjes:
After some more testing it seems to me that maybe one of `_song_started` or `_song_changed` could be ignored. I don't know which one however as I am not sure to fully understand the difference.

Other than that, I'm quite happy with the result.

lazka:
> After some more testing it seems to me that maybe one of _song_started or _song_changed could be ignored. I don't know which one however as I am not sure to fully understand the difference.

song-started is after app.player.song changes, song-changed is after any tags of the provided songs have changed.

ptitjes:
> song-started is after app.player.song changes, song-changed is after any tags of the provided songs have changed.

Yeah! I finally figured out as you can see in my last commit :)

ptitjes:
Thank you!

PR #2290, opened 2017-03-02
Drop Ubuntu 14.04 support Fixes #2288

### LINKED ISSUE #2288: Drop Ubuntu 14.04 support
Questions:

* It's currently the oldest supported, what would be the new "oldest"?
* What does it give us in terms of dependencies?

lazka:
In case we drop Ubuntu 14.04 this is the oldest supported distro


.travis/Dockerfile.debian8.py2
```diff
@@ -0,0 +1,34 @@
...
+    python-setuptools \
+    xvfb \
+    gettext \
+    intltool \
+    libxine2 \
+    dbus-x11 
+
+RUN pip install --upgrade pycodestyle pyflakes xvfbwrapper
+
+RUN useradd -ms /bin/bash user
```
declension: More of a general question: why are we adding a new user in all the Dockerfiles? I'd have thought we can run as root (like most containers do unless they really need to)?

lazka: Some tests failed which used chmod to make a file read only. I thought that since non-root is the usual execution env it can't hurt. Is there any downside?

declension: Cool I see - and no not really, other than extra noise in the `Dockerfile`s

PR #2294, opened 2017-03-03
FilterPluginBox: Remove Gtk.ScrolledWindow in 'More options...' expander

### LINKED ISSUE #2260: Regex Substitution plugin is broken
The regex entry doesn't show up after clicking "More options" in the edit tag windows.

Before:
ql_not_collapsed

After:
ql_collapsed

QL version: 3.8.99+8900~05c6d0e1b-0\~ppa0\~xenial

frestr:
Fixes #2260

The VBox containg the plugin(s) weren't visible when when inside a Gtk.ScrolledWindow. This was fixed by removing the ScrolledWindow and adding the VBox directly to the expander.

Admittedly this removes the ability to scroll the VBox, but it seems like the plugin "Regex Substitution" was the only plugin using this space anyway. So the question is how important this (currently unused) functionality is.

Tested on Ubuntu Xenial and Arch.

### DIFF
--- a/quodlibet/quodlibet/qltk/_editutils.py
+++ b/quodlibet/quodlibet/qltk/_editutils.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2004-2006 Joe Wreschnig, Michael Urman, Iñigo Serna
 #                2014 Nick Boultbee
+#                2017 Fredrik Strupe
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License version 2 as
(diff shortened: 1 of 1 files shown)


ptitjes:
> it seems like the plugin "Regex Substitution" was the only plugin using this space anyway

I guess some people out there may have their own filter plugins in their `.quodlibet` directory, right ?
Will they layout correctly without the ScrolledWindow ?

frestr:
I tested with a duplicate of the Regex Substitution plugin, and it does layout correctly (at least as far as I can see).



I think this only will be an issue if the user has loads of plugins using this space, or some that takes up much vertical space.

Fixing the ScrolledWindow would of course be a better solution, but I was unable to find a fix with it being included that worked on Ubuntu.

lazka:
We only have two plugins which don't take much space, so this looks good to me. If we have a different use case in the future we can have another look. Thanks @frestr

Issue #1118, opened 2015-03-15
Allow resizing of panes in Paned Browser


lazka:
Original [issue 1118](https://code.google.com/p/quodlibet/issues/detail?id=1118) created by kelletim on 2013-02-16T16:30:25.000Z:

The Paned Browser view is awfully nice to browse a collection, but there is so much wasted space in some panes and not enough in others. It'd be awesome if one could slide the panes at the sides and adjust their size. See the pic below for what I'm talking about ...

PR #2301, opened 2017-03-06
Allow resizing of panes in PanedBrowser

frestr:
Implements #1118 

The way it is implemented is to use a nested paned structure instead of the existing vbox. For example, for three panes - p1, p2 and p3 - the structure will look like Paned(p1, Paned(p2, p3)). Gtk doesn't seem to have paneds with more than two panes, so I'm not sure if this can be solved in any other way.

The pane widths are made persistant by extension of the paneds being instances of ConfigRHPaned. ConfigRHPaned only supports writing its width to a single option variable in the config though, so to enable an arbitrary number of paned widths to be saved, each width is saved in the config under the "browser_paned_widths" section, under options "0", "1" etc.

I'm not sure if this is the preferred way to do it, as I couldn't find anything similar already in the code. But as I see it, the alternative will be to either have an arbitrary number of widths saved under the "browsers" section (not good imo), or to save them as a single list option. The disadvantage with the latter will be that ConfigRHPaned needs to be changed, and I also don't think the paneds can constantly write their widths like they do now if they have to all write to the same option (so it may need to be changed to only write widths on shutdown / browser change). Any input on this?

There's also a checkbox in the paned browser preferences - "Equal pane width" - which is checked by default. When checked, all panes will be set to have equal width when the apply button is clicked, so basically like it has been up until now. The reason for having that checkbox is in case people don't want their pane widths to be changed after adding a new pane.

### DIFF
--- a/quodlibet/quodlibet/browsers/paned/main.py
+++ b/quodlibet/quodlibet/browsers/paned/main.py
@@ -24,6 +24,7 @@ from quodlibet.qltk.searchbar import SearchBarBox
 from quodlibet.qltk.x import ScrolledWindow, Align
 from quodlibet.util.library import background_filter
 from quodlibet.util import connect_destroy
+from quodlibet.qltk.paned import ConfigMultiRHPaned
 
 from .prefs import PreferencesButton
 from .util import get_headers
@@ -103,6 +104,8 @@ class PanedBrowser(Browser, util.InstanceTracker):
         self.main_box = qltk.ConfigRPaned("browsers", "panedbrowser_pos", 0.4)
         self.pack_start(self.main_box, True, True, 0)
 
+        self.multi_paned = ConfigMultiRHPaned("browsers",
+                                              "panedbrowser_pane_widths")
         self.refresh_panes()
 
         for child in self.get_children():
@@ -114,14 +117,13 @@ class PanedBrowser(Browser, util.InstanceTracker):
     def set_wide_mode(self, do_wide):
         hor = Gtk.Orientation.HORIZONTAL
         ver = Gtk.Orientation.VERTICAL
-        panes = self.main_box.get_child1()
 
         if do_wide:
             self.main_box.props.orientation = hor
-            panes.props.orientation = ver
+            self.multi_paned.change_orientation(horizontal=False)
         else:
             self.main_box.props.orientation = ver
-            panes.props.orientation = hor
+            self.multi_paned.change_orientation(horizontal=True)
 
     def _get_text(self):
         return self._sb_box.get_text()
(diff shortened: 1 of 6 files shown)


lazka:
> The way it is implemented is to use a nested paned structure instead of the existing vbox. For example, for three panes - p1, p2 and p3 - the structure will look like Paned(p1, Paned(p2, p3)). Gtk doesn't seem to have paneds with more than two panes, so I'm not sure if this can be solved in any other way.

There is a multi paned in C: https://github.com/chergert/panel-gtk/blob/master/src/pnl-multi-paned.c
or one could wrap your strategy in a new container widget, hiding all the ugly details. But all in all it looks good. If we need it in more places we definitely have to look into cleaning it up.

> The pane widths are made persistant by extension of the paneds being instances of ConfigRHPaned. ConfigRHPaned only supports writing its width to a single option variable in the config though, so to enable an arbitrary number of paned widths to be saved, each width is saved in the config under the "browser_paned_widths" section, under options "0", "1" etc.

ok. hacky, but works :)

> I'm not sure if this is the preferred way to do it, as I couldn't find anything similar already in the code. But as I see it, the alternative will be to either have an arbitrary number of widths saved under the "browsers" section (not good imo), or to save them as a single list option. The disadvantage with the latter will be that ConfigRHPaned needs to be changed, and I also don't think the paneds can constantly write their widths like they do now if they have to all write to the same option (so it may need to be changed to only write widths on shutdown / browser change). Any input on this?

Ideally we'd have a new widget (not ConfigRHPaned) which would give you a list of widths which can be saved if any change. Re save on shutdown: we don't have any event to save atm (destroy is too late) so we always save everything on change.

> There's also a checkbox in the paned browser preferences - "Equal pane width" - which is checked by default. When checked, all panes will be set to have equal width when the apply button is clicked, so basically like it has been up until now. The reason for having that checkbox is in case people don't want their pane widths to be changed after adding a new pane.

What about a "Reset pane widths" button instead of the checkbox?

frestr:
> There is a multi paned in C: https://github.com/chergert/panel-gtk/blob/master/src/pnl-multi-paned.c
or one could wrap your strategy in a new container widget, hiding all the ugly details. But all in all it looks good. If we need it in more places we definitely have to look into cleaning it up.

Nice, I wasn't aware of that. It seems like it has not been integrated into Gtk yet though, but is marked as a prototype on the [Gtk roadmap](https://wiki.gnome.org/Projects/GTK%2B/Roadmap).

> Ideally we'd have a new widget (not ConfigRHPaned) which would give you a list of widths which can be saved if any change. Re save on shutdown: we don't have any event to save atm (destroy is too late) so we always save everything on change.

Sounds like a good idea. I'll try to encapsulate the functionality in a new widget.

> What about a "Reset pane widths" button instead of the checkbox?

I actually had that in the beginning, but found it rather tedious to always have to click both apply and that button when changing the panes if I wanted the panes to have the same width. A button may be a little more intuitive though.

zsau:
>I actually had that in the beginning, but found it rather tedious to always have to click both apply and that button when changing the panes if I wanted the panes to have the same width. A button may be a little more intuitive though.

FWIW, the way it works for columns in a TreeView is that any column the user resizes gets its width saved (becoming fixed-width), while other columns' widths remain unspecified/undefined. Then the "extra" view width (total width minus all fixed-width columns) is divided equally among the non-fixed-width columns.

frestr:
I've added two classes, MultiRPaned and ConfigMultiRPaned (plus respective "orientation" versions), mostly mimicking the style of RPaned and ConfigRPaned. ConfigRPaned saves the widths in a string list, so now only a single option field is needed. Thoughts?

urielz:
@frestr this is awesome! I really wanted this in QL. Thank you.

lazka:
Looks good

lazka:
Is there anything missing or should I merge?

frestr:
Everything should be in place, so go ahead.

lazka:
Thanks

PR #2305, opened 2017-03-08
accelerators for "Open Browser" Menu

urielz:
The ability to open a secondary browser while playing music in the main browser is a nice way browse the library while listening to music. This is a feature that I suspect is frequently used by other users as well, yet it lacks shortcuts. This PR adds shortcuts for all the available browsers. It replicates the main shortcuts with the addition of the _alt_ modifier.


declension:
Good idea, and works here - thanks.

PR #2307, opened 2017-03-08
plugins: add a plugin to export a playlist to a folder

ptitjes:
This is a rough cut.
Please review and comment as to make it better.

### DIFF
--- a/quodlibet/quodlibet/ext/playlist/export_to_folder.py
+++ b/quodlibet/quodlibet/ext/playlist/export_to_folder.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Didier Villevalois
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation
+
+from gi.repository import Gtk
+
+from quodlibet import _
+from quodlibet.plugins.playlist import PlaylistPlugin
+from quodlibet.qltk import Icons
+from quodlibet.qltk.notif import Task
+from quodlibet.util import copool
+from quodlibet.util.dprint import print_d
+
+import os
+from shutil import copyfile
+
+
+class ExportToFolder(PlaylistPlugin):
+    PLUGIN_ID = "ExportToFolder"
+    PLUGIN_NAME = _(u"Export Playlist to Folder")
+    PLUGIN_DESC = \
+        _("Exports a playlist by copying files to a folder.")
+    PLUGIN_ICON = Icons.FOLDER
+    ELLIPSIZE_NAME = True
+    _PERSIST_FUDGE = 100
+
+    TEMP_PLAYLIST = "_quodlibet"
+
+    def __copy_songs(self, task, songs, directory):
+        """Generator for copool to copy songs to the folder"""
+        self.__cancel = False
+        total = len(songs)
+        print_d("Copying %d song(s) to directory %s. "
+                "This might take a while..." % (total, directory))
+        for i, song in enumerate(songs):
+            if self.__cancel:
+                print_d("Cancelled export to directory.")
+                self.__cancel = False
+                break
+            # Actually do the copy
+            self._copy_file(song, directory, i)
+            task.update(float(i) / total)
+            yield True
+        print_d("Finished export to directory.")
+        task.finish()
+
+    def __cancel_copy(self):
+        """Tell the copool to stop copying songs"""
+        self.__cancel = True
+
+    def _copy_file(self, song, directory, index):
+        filename = song["~filename"]
+        print_d("Copying %s." % filename)
+        basename = os.path.basename(filename)
+        copyfile(filename, "%s/%04d - %s" % (directory, index, basename))
+
+    def plugin_playlist(self, playlist):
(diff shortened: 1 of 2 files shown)


declension:
@ptitjes I'll try to have a look tonight.

(FYI this kind of feature (albeit a less pretty version) can often be achieved by the Custom Commands plugin)

ptitjes:
@declension Thx.

> FYI this kind of feature (albeit a less pretty version) can often be achieved by the Custom Commands plugin

I see. I did not this Custom Commands feature. But indeed, I'm glad to be able to select the destination folder and change the pattern at export :)

quodlibet/quodlibet/ext/playlist/export_to_folder.py
```diff
@@ -0,0 +1,165 @@
...
+        frame.set_shadow_type(Gtk.ShadowType.IN)
+        frame.set_border_width(0)
+        box.pack_start(frame, True, True, 0)
+
+        pattern_label = Gtk.Label(_("Filename pattern:"))
+        pattern_label.set_line_wrap(True)
+        pattern_label.set_xalign(0.0)
+        box.pack_start(pattern_label, True, True, 0)
+
+        self.pattern_entry = Gtk.Entry(text=pattern)
```
declension: a `qltk.UndoEntry` is slightly nicer IMO, or even `ValidatingEntry` here.

quodlibet/quodlibet/ext/playlist/export_to_folder.py
```diff
@@ -0,0 +1,165 @@
...
+
+        box = Gtk.VBox(spacing=6)
+
+        destination_label = Gtk.Label(_("Destination folder:"))
+        destination_label.set_line_wrap(True)
+        destination_label.set_xalign(0.0)
+        box.pack_start(destination_label, True, True, 0)
+
+        frame = Gtk.Frame()
+        self.directory_chooser = \
```
declension: FYI we generally try to avoid backslash continuations (some discussion on alternatives, though the PEP itself was rejected: https://www.python.org/dev/peps/pep-3125/)

quodlibet/quodlibet/ext/playlist/export_to_folder.py
```diff
@@ -0,0 +1,165 @@
...
+        self.vbox.pack_start(box, True, True, 0)
+
+        self.set_response_sensitive(Gtk.ResponseType.OK, False)
+
+        def changed(*args):
+            has_directory = self.directory_chooser.get_filename() is not None
+            self.set_response_sensitive(Gtk.ResponseType.OK, has_directory)
+
+            pattern_text = self.pattern_entry.get_text()
+            has_pattern = pattern_text != ""
```
declension: FYI it's generally considered more Pythonic to use `has_pattern = bool(pattern_text)`, or in the negative case just `not pattern_text` (also deals with `None`, etc)

quodlibet/quodlibet/ext/playlist/export_to_folder.py
```diff
@@ -0,0 +1,165 @@
...
+        self.directory_chooser.connect("selection-changed", changed)
+        self.pattern_entry.connect("changed", changed)
+
+        self.get_child().show_all()
+
+
+class Config(object):
+    _config = PluginConfig(__name__)
+
+    DEFAULT_PATTERN = "<artist> - <title>"
```
declension: I recommend perhaps using a tied tag "<artist~title>" as it works better if is either is missing (e.g. in Classical music `0004. -  - Sonata K.380.flac`")
...unless those filenames themselves needs parsing by something strict I guess, but this is probably not the _default_ case...

quodlibet/quodlibet/ext/playlist/export_to_folder.py
```diff
@@ -0,0 +1,165 @@
...
+            CONFIG.default_pattern = entry.get_text()
+
+        vbox = Gtk.VBox(spacing=6)
+
+        def create_pattern():
+            hbox = Gtk.HBox(spacing=6)
+            hbox.set_border_width(6)
+            label = Gtk.Label(label=_("Default filename pattern:"))
+            hbox.pack_start(label, False, True, 0)
+            entry = Gtk.Entry()
```
declension: This one more so even perhaps (custom `Entry`)

declension:
A few comments inline but generally LGTM

quodlibet/quodlibet/ext/playlist/export_to_folder.py
ptitjes: Looking at the `PatternCompiler`/`PatternParser`/`PatternLexer` trio, I fail to see where it can fail. They seem to be bullet-proof even with ill-written patterns. What would validate a `ValidatingEntry` then ?

declension: Yeah that logic has changed a lot now and is more resilient to user errors (i.e. assumes more things are just text now). Probably needs some rethought, but there are too many use cases in the wild now for significant changes without breaking lots of people's set ups.

Anyway I found you can usually break it with `#` pretty easily - `<#foo>`

declension:
Thanks

ptitjes:
Thank you.

PR #2310, opened 2017-03-10
properties: add a Playlists tab in the properties panel

### LINKED ISSUE #2259: Classify music in playlists
I like to have my music well organized.
So I have many categories (playlists).  Typically a song belongs to 5 or more playlists of 30 possible.
It is very cumbersome to add a music to several playlists, because the menu disappears after selecting a playlist.

For example, Guayadeque solves this by having a popup dialog where several playlists (labels in Guayadeque terminology) can be easily selected or deselected.

See screenshot.
image

ptitjes:
Fixes #2259.

### DIFF
--- a/quodlibet/quodlibet/browsers/playlists/panel.py
+++ b/quodlibet/quodlibet/browsers/playlists/panel.py
@@ -1,0 +1,179 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Didier Villevalois
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation
+
+from gi.repository import Gtk, Pango
+
+from quodlibet import _
+from quodlibet import qltk
+from quodlibet.browsers.playlists.menu import ConfirmMultipleSongsAction
+from quodlibet.browsers.playlists.util import GetPlaylistName, PLAYLISTS
+from quodlibet.qltk import Icons, get_top_parent
+from quodlibet.qltk.edittags import AudioFileGroup
+from quodlibet.qltk.models import ObjectStore
+from quodlibet.qltk.views import RCMHintedTreeView
+from quodlibet.qltk.x import ScrolledWindow
+from quodlibet.util.collection import Playlist, FileBackedPlaylist
+
+
+class PlaylistPanel(Gtk.VBox):
+    def __init__(self, parent, library):
+        super(PlaylistPanel, self).__init__(spacing=12)
+        self.title = _("Playlists")
+        self.set_border_width(12)
+
+        self.__view = self._create_playlist_view()
+
+        swin = ScrolledWindow()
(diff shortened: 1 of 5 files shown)


quodlibet/quodlibet/browsers/playlists/panel.py
```diff
@@ -0,0 +1,179 @@
...
+        self._toggle_playlist(path)
+
+    def _toggle_playlist(self, path):
+        model = self.__view.get_model()
+        iter_ = model.get_iter(path)
+        playlist = model.get_value(iter_)
+
+        songs = self.__songinfo.songs
+
+        has_some, has_all = playlist.has_songs(songs)
```
ptitjes: I should have factored out all this code from `menu.py` in a shared function!

ptitjes:
Sorry for the code duplication, it is now fixed.

lazka:
@declension any thoughts?

What about moving this to the information window since it doesn't have anything to do with tagging?

screenshot from 2017-03-14 14-09-29

ptitjes:
@lazka Just some remarks on why I think it is better in the "Properties" window :
- the menu is called "Edit Tags" but the window is called "Properties"
- the average user has now idea what is a tag and that playlist association is not stored in tags
- there is no left pane to select files in "Information"
- "Information" sounds more like there is nothing to edit
- btw the "Renames Files" itself has some sense in a "Properties" window bu should not be found by clicking on a "Edit Tags" menu item...

So IMHO the menu items should be called "Edit Properties..." or just "Properties...".

declension:
Yeah, both options have slight UX issues. I've been in favour of tidying up the Information / Edit Tags confusion for a while...  To me the real distinction should be editing versus viewing, I think there's a ticket with this somewhere.

So IMO the nicest option is to keep this for now, change the Edit Tags item name (but separately), then maybe move editing of bookmarks and lyrics around etc

ptitjes:
@lazka @declension I can make a separate PR to change the "Edit Tags" name, if you want.

lazka:
What about renaming it into "Edit Tags" and "Properties"? So it's clear the first one is only for tags (and shared with EF) and the latter is about both information and editing other things.

Regarding "there is no left pane to select files in "Information": at least for the use case in #2259 (adding to many playlists) this shouldn't be a problem.

ptitjes:
Well, I like the idea of the separation of viewing and editing as pointed by @declension. You seem to prefer that the features common to EF and QL be exactly the same. Is that it ?

> Regarding "there is no left pane to select files in "Information": at least for the use case in #2259 (adding to many playlists) this shouldn't be a problem.

OK, the screenshot of the other application in that ticket made me think that it was what the OP was proposing.

lazka:
> Well, I like the idea of the separation of viewing and editing as pointed by @declension. You seem to prefer that the features common to EF and QL be exactly the same. Is that it ?

Yes. That's how it is now at least, and I don't think this issue is a good enough reason to change that.

declension:
Closing old MRs (#3655)

PR #2311, opened 2017-03-10
headerbar: layout player controls in headerbar when running on Gnome

### LINKED ISSUE #2297: Use the header bar to display the player controls
Toyed with this idea for 5 minutes... I mean something like this:

headerbar

Handling full screen could be done ala. Totem, showing the header bar when the pointer is moved.

What do you think ?

ptitjes:
Fixes #2297.

* cleaned and encapsulate player controls' behavior
* introduced an `is_gnome()` helper
* use the header bar to layout player controls when running on Gnome

I was not very sure for the implementation of `is_gnome()`. I guess that a Gnome-specific dbus service to test would be more correct and more efficient, but I was dry on that.

### DIFF
--- a/quodlibet/quodlibet/util/environment.py
+++ b/quodlibet/quodlibet/util/environment.py
@@ -32,6 +32,12 @@ def _dbus_name_owned(name):
         return False
 
 
+def is_gnome():
+    """If we are running under Gnome"""
+
+    return not is_plasma() and not is_unity() and not is_enlightenment()
+
+
 def is_plasma():
     """If we are running under plasma"""
 
--- a/quodlibet/quodlibet/qltk/controls.py
+++ b/quodlibet/quodlibet/qltk/controls.py
@@ -236,6 +236,30 @@ class PlayPauseButton(Gtk.Button):
             player.paused = not self.get_active()
 
 
+class PreviousSongButton(Gtk.Button):
+
+    def __init__(self, player, relief=Gtk.ReliefStyle.NONE):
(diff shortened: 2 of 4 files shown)


ptitjes:
So it seems that I have to do some more tests for the HeaderBar's availability...

quodlibet/quodlibet/util/environment.py
```diff
@@ -32,6 +32,12 @@ def _dbus_name_owned(name):
         return False
 
 
+def is_gnome():
+    """If we are running under Gnome"""
+
+    return not is_plasma() and not is_unity() and not is_enlightenment()
```
bil-elmoussaoui: This is just so wrong , you can use `return os.environ.get("XDG_SESSION_DESKTOP", "") == "gnome"` instead

bil-elmoussaoui: + Checking if the user uses GNOME is not enough. Gtk.HeaderBar was introduced on Gtk 3.14. You will have to check if `Gtk.MAJOR_VERSION >= 3 && Gtk.MINOR_VERSION => 14`

Boiethios:
So, what's up about this? I'm a GNOME fanboy and I'd find cool to have this appearance ;)

declension:
Cleaning up old PRs. Any strong motiviations to keep this open?

@ptitjes (or someone else) can it be updated / fixed (and made optional, ideally)?

declension:
Closing old MRs (#3655)

Issue #1853, opened 2016-02-23
Online crash reporting


lazka:
It would be nice to add a way to allow the user to submit a crash report to us in some way.

For Python exceptions we should display a short stacktrace and remove paths/usernames. In case the user has faulthandler installed we could write the C stacktrace to a file and suggest to report it on the next QL start.

Remaining questions:

1) Where to submit to?
2) How to prevent DOS? Captcha?

lazka:
We could use sentry: https://sentry.io/welcome/

declension:
Yes, Sentry would be cool actually. Didn't think of it (and didn't realise it had a fairly generous free tier) :)

lazka:
They have recently revised their pricing, but the basic one is still free. I feel confident that they wont remove the free option or disappear and that we can depend on it to some extend. In the end most of the work required would also be useful for other services.

My current thinking:

* It would be nice to have this at the switch to Python 3. I assume that there will be quite a bit of fallout and catching that early would be nice.
* We can depend on faulthandler, write Python stack traces to disk on segfault and on the next start raise the error, pushing it to the crash reporter.
* Since python-raven (sentry client) isn't in debian we can make the crash reporting pluggable and put the python package at the right place for our Windows/macOS/PPA builds and not include it in the tarball.

One question remaining is how to present errors to the user. Just showing "submit error" will give us more tracebacks but makes it harder to reproduces some errors because we have no channel to ask questions. Providing both options with a short explanation might do the trick ([2] "Report error to us on github" [1] "Just send error log (better than nothing)")

NoahAndrews:
Perhaps we should also have an in-app text box so they can at least say what they were doing when the app crashed.

lazka:
@NoahAndrews good idea. sentry has a user feedback interface, but that is javascript only. Maybe we can reuse the API locally.

lazka:
I've wrapped the whole thing to get the following API now:

```
sentry = Sentry(SENTRY_DSN)

def foo3():
    a = 6
    try:
        object.fooo5
    except Exception:
        try:
            err = sentry.capture()
            err.set_feedback("my python3 commentöäp")
            err.send(SENTRY_TIMEOUT)
        except SentryError:
            print("something broke, sorry")
```

[...]

PR #2313, opened 2017-03-11
Online crash reporting. See #1853

lazka:
Most of the error reporting is now in quodlibet.errorreport

* Logging the error and printing to stdout
* Writing the error + recent logs to disk
* Showing the error to the user
* Opening a pre-filled github issue
* Sending the error sentry.io (github fallback)
* Logging/reporting segfault stacktraces

Depends on raven, which is unlikely to be installed atm, maybe we'll vendor it.


lazka:
This now includes raven and contextlib2 directly in git and will also install it. They both don't get included in the source tarball and the code falls back to prefilled gihub issues in those cases. Couldn't think of anything better..

raven required quite a bit of monkey patching to get working for our use case, but since we only use the internal version this should be fine I guess.

I consider this mergeable, feedback welcome.

declension:
Good work! All seems to function properly here, though I haven't got the Sentry login to check anything on the other side.

The bit that scares me slightly is the diff: `+14,804 −490 lines`.
Is there a way of pulling `raven` (in a build step maybe), and monkey patching the original one - or is that too tricky (haven't seen the code really)? 

Alternatively is an egg / wheel in Git an option - then there's a cleaner distinction between vendor code and QL code (and devs / IDEs don't see that code so much)? I guess it would still need an explicit build step which is a bit weird for Python (maybe...).

lazka:
> Good work! All seems to function properly here, though I haven't got the Sentry login to check anything on the other side.

Thanks. I'll look into making it possible for more people to see the data.

https://sentry.io/share/issue/3134323431352e323334303630333835/

> The bit that scares me slightly is the diff: `+14,804 −490 lines`.
Is there a way of pulling `raven` (in a build step maybe), and monkey patching the original one - or is that too tricky (haven't seen the code really)? 

Build step would work somehow. The upside now is that we can test it on travis.

> Alternatively is an egg / wheel in Git an option - then there's a cleaner distinction between vendor code and QL code (and devs / IDEs don't see that code so much)? I guess it would still need an explicit build step which is a bit weird for Python (maybe...).

I haven't checked, but zip import might work. Is there any problem with your IDE and the current solution?

See https://github.com/quodlibet/quodlibet/pull/2313/commits/bebd0ef523fe1d6a5df2c6bf958c6cf7d1e8aaee for the code needed to mark quodlibet.optpackages as "not part of the project".

All in all I'm open to rework this in the future if an alternative arises.

lazka:
> Thanks. I'll look into making it possible for more people to see the data.

I've asked if we can get another user.

declension:
> I haven't checked, but zip import might work. Is there any problem with your IDE and the current solution?

Zip import sounds good. No problems as such, more just being in the source tree it gets treated like real source code, e.g. highlighting problems (>1000 warnings). I suppose we can just configure IDEs to also ignore that folder tree.

BTW do we need _any_ of `optpackages/raven/contrib` (support for Django / Celery / Tornado etc)?

> All in all I'm open to rework this in the future if an alternative arises.

OK so maybe let's get it in now and think of something to do with build steps / zip imports separately...

lazka:
> BTW do we need any of optpackages/raven/contrib (support for Django / Celery / Tornado etc)?

Thanks. I've removed that now and automated the process: https://github.com/quodlibet/quodlibet/pull/2313/commits/0c026ea5e03156e85360ed6f60a7ce24ed97dd5f

(and sqashed/reorderd the patches..)

lazka:
> > Thanks. I'll look into making it possible for more people to see the data.
>
> I've asked if we can get another user.

Our account got upgraded now.

PR #2314, opened 2017-03-11
fix title if main window title in advanced preferences is empty

urielz:
This is a (very) silly edge case. If "Main Window Title" in "Advanced Preferences" plugin is empty the the main window title is " - Quod Libet" which looks a bit weird. This PR fixes that by replacing the title with "Quod Libet"


declension:
LGTM

declension:
Thanks

PR #2326, opened 2017-03-18
Fix code tags in the searching docs

frestr:
Missing colons made the examples regarding the new != operator not display properly.

Apparently the file also missed a newline at the end.


declension:
Thanks

PR #2327, opened 2017-03-18
Quodlibet 3.6

redvimo:


redvimo:
This pull request is to fix the Menu Icon being too large. It ignores the bounds of it's containing Button.
My solution was to add more sizes of icon images into the folder "quodlibet/quodlibet/images/hicolor/".
I have only created the 16x16 size, because gnome/gtk+ is automatically choosing to use the smallest size that is available when the application is run. Below is the screenshot. I have also tried removing the "64x64" folder and leaving only the "scalable" folder, but that does not solve the problem. It seems to still show some ".png" file because the image is pixelated.

For now I am sticking to having all the image sizes available (16x16, 24x24, 32x32, 64x64).
If you want me to complete the other sizes and submit another pull, then let me know.

Below are the screenshots:
screenshot from 2017-03-18 04-20-08

screenshot from 2017-03-18 04-20-34

redvimo:
Im closing this pull and redoing it.

PR #2328, opened 2017-03-18
Quodlibet 3.6

redvimo:
This pull request is to fix the Menu Icon being too large. It ignores the bounds of it's containing Button.
My solution was to add more sizes of icon images into the folder "quodlibet/quodlibet/images/hicolor/".
I have only created the 16x16 size, because gnome/gtk+ is automatically choosing to use the smallest size that is available when the application is run. Below is the screenshot. I have also tried removing the "64x64" folder and leaving only the "scalable" folder, but that does not solve the problem. It seems to still show some ".png" file because the image is pixelated.

For now I am sticking to having all the image sizes available (16x16, 24x24, 32x32, 64x64).
If you want me to complete the other sizes and submit another pull, then let me know.

Below are the before and after screenshots:

**_Before:_**
screenshot from 2017-03-17 14-35-16

**_After:_**
screenshot from 2017-03-18 04-59-27

screenshot from 2017-03-18 04-59-33


lazka:
Indeed, thanks. I wonder how gnome-shell figures out our internal icon theme path...

1) We need a pull request for the master branch, this is for the old 3.6 branch
2) Providing all icon sizes sounds like a good idea (for both exfalso and quodlibet)
3) It would be nice if the png generation process was automated. We do something similar for our documentation favicon here: https://github.com/quodlibet/quodlibet/blob/master/quodlibet/docs/favicon/Makefile

If you need help, just ask.

redvimo:
Ok ill a try a Makefile for the images folder. Then ill submit for the
master branch.
I did the pull for 3.6 because I was using ubuntu gnome 16.10, which comes
with quodlibet 3.6.2. I thought if i did it for that version then it would
get an update for the ubuntu repository. Is that how it works ? Or can
ubuntu decide to use any version? Which Distro do url use primarily for
development? I use ubuntu trisquel and debian.
In future should i submit a pull only to master branch or for all the
branches?

lazka:
Ubuntu uses the version which happens to be in Debian when they prepare a release.
We provide two PPAs which allow you to use newer versions for various Ubuntu versions. A stable one, which gets updates every few months when a new release happens and an unstable one which tracks master and gets an update every week or so.

I currently use debian sid.

Pull requests should always be against master. If the change fixes a bug we might backport it to older releases if needed.

PR #2330, opened 2017-03-18
RPaned: ensure get_relative always returns a value between 0 and 1

### LINKED ISSUE #2321: Preserve queue height when queue is hidden and browser is changed
**Steps**:
* Show the queue by clicking the queue visibility button in the lower right corner and expand it
* Set the queue to a preferred height
* Hide the queue by clicking the button again
* Change browser
* Show the queue again

**Expected**: The queue will have the same height as it had when being hidden in step 3. 

**Actual**: The queue height is set to the minimum/default height.

The queue height is preserved if
- toggling visibility in the same browser,
- changing browser with with the queue being visible and
- after a restart (both when visible and not),

but not when the queue is hidden and the user changes browser.

May seem negligible, but can occasionally be a little annoying.

frestr:
Fixes #2321 

The cause of the issue was that RPaned's get_relative() occasionally returned a value greater than 1, because get_max() returned a value of 1 while get_position() returned a value >1. The solution was to not solely trust the _alloced flag, and instead return __relative in cases where the function otherwise would return invalid values.

To make the relative position consistently between 0 and 1, I also added a check for this in set_relative().

### DIFF
--- a/quodlibet/quodlibet/qltk/paned.py
+++ b/quodlibet/quodlibet/qltk/paned.py
@@ -77,6 +77,9 @@ class RPaned(Paned):
     def set_relative(self, v):
         """Set the relative position of the separator, [0..1]."""
 
+        if v < 0 or v > 1:
+            raise ValueError("v must be in [0..1]")
+
         if self.__alloced:
             max_pos = self._get_max()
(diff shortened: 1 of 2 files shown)


lazka:
Thanks

PR #2340, opened 2017-03-20
Add checkbox to stop player after queue is empty

### LINKED ISSUE #43: rfe: stop playing once queue runs out
Original [issue 43](https://code.google.com/p/quodlibet/issues/detail?id=43) created by hhaamu on 2008-11-03T18:33:54.000Z:

QL currently has a &quot;Stop after this song&quot; feature hidden beneath the play
button. I'd like it to be duplicated for the queue feature. I often queue
up stuff and wish QL to stop playing after they are played.

Currently, to accomplish this I must either a) clear the playlist or b)
wait for the last song in the queue and use the aforementioned feature.

frestr:
Fixes #43

Although the "Queue Only" plugin has this as a side effect, the effect of always adding songs to the queue may not be desired. This addition instead adds a checkbox - "Stop After Empty" - next to the "Random" checkbox that has no other side effects than stopping the player after the queue is empty, instead of continuing to play a song from the song pane.

### DIFF
--- a/quodlibet/quodlibet/qltk/queue.py
+++ b/quodlibet/quodlibet/qltk/queue.py
@@ -133,9 +133,9 @@ class QueueExpander(Gtk.Expander):
         left.pack_start(rand_checkbox, False, True, 0)
 
         stop_checkbox = ConfigCheckButton(
-            _("Stop After Empty"), "memory", "queue_stop_after_empty")
+            _("Stop Once Empty"), "memory", "queue_stop_once_empty")
         stop_checkbox.set_active(config.getboolean("memory",
-                                                   "queue_stop_after_empty",
+                                                   "queue_stop_once_empty",
                                                    False))
         stop_checkbox.set_no_show_all(True)
         left.pack_start(stop_checkbox, False, True, 0)
@@ -216,7 +216,7 @@ class QueueExpander(Gtk.Expander):
     def __clear_queue(self, activator):
         self.model.clear()
         stop_queue = config.getboolean("memory",
-                                       "queue_stop_after_empty",
+                                       "queue_stop_once_empty",
                                        False)
         if stop_queue:
             app.player_options.stop_after = True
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/qltk/queue.py
```diff
@@ -121,16 +123,27 @@ def __init__(self, library, player):
...
+                _("_Random"), "memory", "shufflequeue")
+        rand_checkbox.connect('toggled',
+                              self.__queue_shuffle,
+                              self.queue.model)
+        rand_checkbox.set_active(config.getboolean("memory", "shufflequeue"))
+        rand_checkbox.set_no_show_all(True)
+        left.pack_start(rand_checkbox, False, True, 0)
+
+        stop_checkbox = ConfigCheckButton(
+            _("Stop After Empty"), "memory", "queue_stop_after_empty")
```
declension: I think _Stop Once Empty_ reads a bit better than _Stop After Empty_

declension:
Looks reasonable in general, and works for me - thanks.

Part of me still thinks this can / should be solved as a play order plugin / configuration for the queue itself, but that's a much bigger piece so maybe best to get this in for now.

quodlibet/quodlibet/qltk/queue.py
frestr: I agree. I based it on "Stop After This Song" under Controls, but the wording makes more sense in that case.

frestr:
I thought about making it a play order plugin, but that doesn't make much sense the way I see it, as it doesn't really have anything to do with the play order.

I think having a separate configuration window for the queue would be nice, especially if we later want to add even more queue options not relating to the play order. Perhaps we could add a new "Queue" tab in the File->Preferences window?

declension:
Yep I think all it's a bit complicated conceptually and UI-wise both now and maybe forever. Essentially I wanted to provide a way to abstract the `PlaylistMux` better, and also unify the "Random" with Shuffle plugins (why can't the queue be shuffled in a weighted way, or whatever).

That way, this and the Queue only plugin could live somewhere a bit more meaningful. Though of course that then makes it _more_ complicated, if slightly neater / more powerful.

...but again this is a whole separate piece really.

@frestr if you can make that text change, then assuming @lazka doesn't have anything I think it's good to go.

declension:
Thanks

lazka:
Thanks

PR #2351, opened 2017-04-05
Make sure that menu items are selected on song change

### LINKED ISSUE #2350: Tray Icon doesn't show current rating in Gnome 3
If you enable the Tray Icon in QL under Gnome 3 then the icon shows, but no values are ever selected in the Ratings sub-menu. From a quick bit of debugging, this appears to be because [set_songs()](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/qltk/ratingsmenu.py#L63) sets the song but does nothing else, and the menu was already built and values selected in the `__init__` method.

Clicking a rating applies the rating to a song, it is just showing the current rating that doesn't work.

Running QL 3.8.1 from OBS on openSUSE 42.2 (Gnome 3.20.2)

IBBoard:
Fixes #2350.

Code change extracts common function (because DRY) and stores ratings items in array so that we don't have to check if sub-menu item is a rating item, a separator or a remove item.

Based on 3.8 branch, because I'm running Stable.

### DIFF
--- a/quodlibet/quodlibet/qltk/ratingsmenu.py
+++ b/quodlibet/quodlibet/qltk/ratingsmenu.py
@@ -46,15 +46,17 @@ class RatingsMenuItem(Gtk.ImageMenuItem):
 
         submenu = Gtk.Menu()
         self.set_submenu(submenu)
+        self._rating_menu_items = []
         for i in RATINGS.all:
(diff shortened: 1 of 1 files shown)


IBBoard:
Found a bug when multiple tracks were selected.

PR #2359, opened 2017-04-17
Quick fix for MPRIS notifications

### LINKED ISSUE #2358: MPRIS metadata is not updated when song rating is changed.
>What did you try to do?

Rate a song.

>What did you expect to happen?

That quodlibet would update it's MPRIS metadata so that the new rating would show up in MPRIS clients that show ratings.

>What did happen instead?

MPRIS metadata is not updated when song rating is changed.

>Which version of Quod Libet?

3.8.1

>Which operating system including version (Ubuntu 14.04, Win7, Debian sid, ...)?

Fedora 25 GNOME

I am a contributor to https://github.com/eonpatapon/gnome-shell-extensions-mediaplayer

A user recently added support for setting ratings in quodlibet which works as expected, but quodlibet doesn't update it's metadata as it should when song ratings change. This results in the new rating not showing up in the extension if set outside of the extension.(the player or from the command line)

The fix for this is simple. Anytime anything in the metadata is changed it should be updated.

IBBoard:
Seems to fix #2358 - `song` is actually a set of songs that were updated and `app.player.info` is the currently playing song, so we skip notifying if no songs were changed or if the playing song is not one of the changed songs

Presumably `song` was once a single song and the check worked

### DIFF
--- a/quodlibet/quodlibet/ext/events/mpris/mpris2.py
+++ b/quodlibet/quodlibet/ext/events/mpris/mpris2.py
@@ -145,8 +145,8 @@ value="false"/>
     def __seeked(self, player, song, ms):
         self.Seeked(ms * 1000)
 
-    def __library_changed(self, library, song):
-        if song and song is not app.player.info:
+    def __library_changed(self, library, songs):
+        if not songs or app.player.info not in songs:
             return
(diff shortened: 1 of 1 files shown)


lazka:
thanks!

PR #2360, opened 2017-04-18
Waveform user friendly color pick

urielz:
Adds the ability to change the elapsed and remaining colors of the waveform via a user-friendly color palette


lazka:
Two observations, not sure if important:

1) First opening the plugin preferences, the color shown in the button and the color used don't match
2) This removes the ability to select the default theme colors

urielz:
Thanks for the comments @lazka. I'll look into that. I'll also add a button to reset the colors to default.

urielz:
Default theme colors is now an option. 

On my end it seems that colors match at first opening the plugin preferences. @lazka Would you mind checking again?

lazka:
Default switches it to red/black here and not the theme colors. Something is wrong with the preferences, they increase the preference window width a lot now for no apparent reason.

urielz:
Ah I see. I didn't know that different themes have different default colors. I'll take a look at that and the width problem as well. Thanks for the feedback @lazka

lazka:
Maybe a checkbox with "Use theme colors" which makes the color picker insensitive would be a solution.

lazka:
Feel free to reopen / create a new one.

Issue #2357, opened 2017-04-17
Updated waveform disappears on repeat track


Meriipu:
I recently updated to head rather than using my local similar change of increase tickrate to milliseconds, though now when a track finishes and repeats, the waveform is a flat line without any visible progress. 

latest version from github on Linux

urielz:
Thanks for the report. I can reproduce as well.

ptitjes:
I can reproduce it too. I'll look into it during next week.

ptitjes:
Sorry for the delay...
I hope it is not too late to make it for 3.9 !

Meriipu:
The waveform persists between repeats now, but did a recent change break #2255 ? 
I went through the commits but could not find any that made sense to me for breaking it

The waveform is at least for me back to updating every second or so.

Meriipu:
Thanks.

I thought it broke the interval at first but I just installed the wrong branch.

PR #2365, opened 2017-04-21
Attempted fix of waveform redraw issue on repeat track

Meriipu:
Fixes issue https://github.com/quodlibet/quodlibet/issues/2357 for me, though I do not know whether anything else breaks.


Meriipu:
yeah it seems it uses the previous tracks waveform for a short moment before changing to the new one

I prefer that behaviour over the new waveform not drawing at all, however, so I will be using this personally until a cleaner fix is implemented.

Meriipu:
@ptitjes

PR #2366, opened 2017-04-25
Added support for custom DateColumn timestamp formats

Meriipu:
Personally I dislike the random "Yesterday",  "23:55:12",  "30/01/2015", and so on timestamps scattered through the datecolumn

This adds an option in advanced preferences to specify the format string to be used.

### DIFF
--- a/quodlibet/quodlibet/qltk/songlistcolumns.py
+++ b/quodlibet/quodlibet/qltk/songlistcolumns.py
@@ -279,14 +279,23 @@ class DateColumn(WideTextColumn):
             except (OverflowError, ValueError, OSError):
                 text = u""
             else:
-                today = datetime.datetime.now().date()
-                days = (today - date).days
-                if days == 0:
-                    format_ = "%X"
-                elif days < 7:
-                    format_ = "%A"
+                format_setting = config.gettext("settings",
+                                      "datecolumn_timestamp_format")
+
+                # use default behaviour-format
+                if not format_setting:
+                    today = datetime.datetime.now().date()
+                    days = (today - date).days
+                    if days == 0:
+                        format_ = "%X"
+                    elif days < 7:
+                        format_ = "%A"
+                    else:
+                        format_ = "%x"
+                # use format from Advanced Preferences
                 else:
-                    format_ = "%x"
+                    format_ = format_setting
+
                 stamp = time.localtime(stamp)
                 text = time.strftime(format_, stamp)
             cell.set_property('text', text)
--- a/quodlibet/quodlibet/qltk/songlistcolumns.py
+++ b/quodlibet/quodlibet/qltk/songlistcolumns.py
@@ -282,8 +282,11 @@ class DateColumn(WideTextColumn):
                 format_setting = config.gettext("settings",
                                       "datecolumn_timestamp_format")
 
+                # use format configured in Advanced Preferences
+                if format_setting:
+                    format_ = format_setting
                 # use default behaviour-format
-                if not format_setting:
+                else:
                     today = datetime.datetime.now().date()
                     days = (today - date).days
                     if days == 0:
@@ -292,9 +295,6 @@ class DateColumn(WideTextColumn):
                         format_ = "%A"
                     else:
                         format_ = "%x"
-                # use format from Advanced Preferences
-                else:
-                    format_ = format_setting
 
                 stamp = time.localtime(stamp)
                 text = time.strftime(format_, stamp)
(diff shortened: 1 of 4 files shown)


Meriipu:
I can not say I know what the failed check is for but ok

Meriipu:
weird

lazka:
What is the format string or behavior that you prefer?

Regarding the failed check, this is an old gtk+ bug which sometimes leads to crashes. I've opened #2371

Meriipu:
I prefer having a date (year, month, day) and optionally a timestamp

Meriipu:
trying to add in the new changes to the file from head to the branch being merged seems to have done more harm than not

Meriipu:
hopefully it rebased now. 

but yeah this is just something I did to solve one of my gripes with gtk3 (applications? it is hard to not put too much of the blame for anything on gtk itself at times) tending to use a bit too mixed/unprecise labels for timestamps for my liking.

declension:
Looks reasonable to me, though would prefer if this had some tests around the now-uncovered code, i.e. testing what happens when the new config entry is set.

quodlibet/quodlibet/qltk/songlistcolumns.py
```diff
@@ -279,14 +279,23 @@ def _apply_value(self, model, iter_, cell, stamp):
...
-                days = (today - date).days
-                if days == 0:
-                    format_ = "%X"
-                elif days < 7:
-                    format_ = "%A"
+                format_setting = config.gettext("settings",
+                                      "datecolumn_timestamp_format")
+
+                # use default behaviour-format
+                if not format_setting:
```
declension: Why not swap the order of the if-else and avoid the `not` (always harder to read)?

Meriipu: I suppose it could be swapped. I started with `== ""` and changed it to `not s` in an attempt to fix some failed check (mentioned earlier), and at that point the new order maybe makes more sense. 

I can try to look into how tests work to add some.

Meriipu: Would something like setting it to a format string and running test_qltk_songlistcolumns.py again work? I have no idea how one would test it, other than maybe setting the current time to a specific value and checking whether a date-column formats to what was expected with some entry.

Meriipu:
Should I include a test for the standard behaviour (Say making "~#added" a date that was yesterday and comparing with "Yesterday"), or is testing for the new format enough?

quodlibet/tests/test_qltk_songlistcolumns.py
```diff
@@ -85,3 +88,20 @@ def test_artist(self):
...
+        format = "%Y%m%d %H:%M:%S PLAINTEXT"
+        quodlibet.config.settext("settings", "datecolumn_timestamp_format",
+                                 format)
+
+        d = datetime.datetime(year=1999, month=5, day=1,
+                              hour=23, minute=11, second=59)
+        stamp = int(time.mktime(d.timetuple()))
+        column = create_songlist_column("~#added")
+        text = self._render_column(column, **{"~#added": stamp})
+        self.assertEqual(text, "19990501 23:11:59 PLAINTEXT")
```
declension: Good stuff :+1:

quodlibet/tests/test_qltk_songlistcolumns.py
```diff
@@ -85,3 +88,20 @@ def test_artist(self):
...
+
+    def test_nonconfigured_datecol_format(self):
+        text = quodlibet.config.gettext("settings",
+                                        "datecolumn_timestamp_format")
+        self.assertEqual(text, "")
```
declension: Perhaps this test (or a separate) should do a quick check to see that the default behaviour _hasn't_ been changed. Bit harder as you said because of the way the existing one works but you could set the date as above and at the very least check that it isn't the above...

quodlibet/quodlibet/qltk/songlistcolumns.py
```diff
@@ -279,14 +279,23 @@ def _apply_value(self, model, iter_, cell, stamp):
...
-                days = (today - date).days
-                if days == 0:
-                    format_ = "%X"
-                elif days < 7:
-                    format_ = "%A"
+                format_setting = config.gettext("settings",
+                                      "datecolumn_timestamp_format")
+
+                # use format configured in Advanced Preferences
+                if format_setting:
```
declension: :+1:

Meriipu:
Maybe it is possible to set it to more than 7 days ago, and under the assumptions tests run under the US locale, expect it to be formatted as "%x", though I am not sure about that and it seems unreliable.

declension:
Sorry about the delay. Thanks @Meriipu

PR #2383, opened 2017-05-15
Update OSX modules

virtuald:
yelp-xsl fails to build if intltool isn't present

```
...
checking for intltool >= 0.40.0...  found
configure: error: Your intltool is too old.  You need intltool 0.40.0 or later.
```


virtuald:
Nevermind, I missed the meta-bootstrap piece.

lazka:
Yeah, my goals was to get rid of the bootstrap and explicitly depend on everything, but I gave up after some time.

PR #2388, opened 2017-05-18
Update German translation

Mellthas:
Translations for all the new strings added since the last update.


lazka:
Thanks!

lazka:
Pushed to 3.9 as well

PR #2402, opened 2017-05-28
Add some css hackery to get a drag handle for the pane separator.

lazka:
So it's easier to see that these things can be moved.
Limited to gtk+ 3.22 for now as I haven't tested witholder versions.

### DIFF
--- a/quodlibet/quodlibet/_init.py
+++ b/quodlibet/quodlibet/_init.py
@@ -383,6 +383,22 @@ def _init_gtk():
         css_override.register_provider("Adwaita", style_provider)
         css_override.register_provider("HighContrast", style_provider)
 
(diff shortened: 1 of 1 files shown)


lazka:
Any opinions on this?

screenshot from 2017-05-28 13-45-09

frestr:
I think it looks alright. Can possibly be a little distracting in the presence of many panes, but that's just nitpicking. I also tested with gtk+ 3.18 (just removed the conditional), and it didn't seem to work there.

lazka:
OK, a bit less contrast maybe?

screenshot from 2017-05-28 17-28-05

frestr:
Yeah, I think it looks better now!

urielz:
Looks good. It's a nice touch.

Issue #2426, opened 2017-06-09
PY3: import playlist throws TypeError


declension:
Looks like the usual `filter` problems. Uncovered by tests...
```
Traceback (most recent call last):

  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/browsers/playlists/main.py", line 580, in __import
    playlist = parse_m3u(filename, library=library)

  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/browsers/playlists/util.py", line 70, in parse_m3u
    return __create_playlist(pl_name, filename, filenames, library)

  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/browsers/playlists/util.py", line 118, in __create_playlist
    playlist.extend(filter(None, songs))

  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/collection.py", line 347, in extend
    self._emit_changed(songs, msg="extend")

  File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/collection.py", line 483, in _emit_changed
    % (msg, len(songs), self))

TypeError: object of type 'filter' has no len()
```

declension:
This is fixed now

PR #2427, opened 2017-06-09
Playlist import DnD: fix import name

### LINKED ISSUE #2424: Playlists: importing a playlist via DnD throws if name is taken
### Steps
 * Drag and drop a valid playlist (m3u / pls) e.g. `foobar.m3u`, where that name is _not_ taken by another playlist to the Playlist browser window
 * Do the same again

### Expected
 * A new playlist is created with non-conflicting name like `foobar (1)`

### Actual
 * A `ValueError`:

```
      File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/browsers/playlists/main.py", line 407, in __drag_data_received
        playlist.rename(name)
      File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/collection.py", line 414, in rename
        self.name = self._validated_name(new_name)
      File "/home/nick/workspace/quodlibet/quodlibet/quodlibet/util/collection.py", line 607, in _validated_name
        _("A playlist named %s already exists.") % new_name)
    ValueError: A playlist named foobar.m3u already exists.
```

declension:
* Ditch the PL temp file on DnD import, and therefore the call
 to `rename` which has no retry logic (unlike `new`)
 * Instead, pass filelike objects.
 * Extract import logic to method too
 * Add tests for this
 * Fix `filter` for PY3 (#2426)
 * Fix tests for new calling
 * Use text not bytes for reading.
 * Make tests use `test_filename()`` - neater.

Fixes #2424,#2426

### DIFF
--- a/quodlibet/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/quodlibet/browsers/playlists/main.py
@@ -7,7 +7,6 @@
 # published by the Free Software Foundation
 
 import os
-from tempfile import NamedTemporaryFile
 
 from gi.repository import Gtk, GLib, Pango, Gdk
 
@@ -393,18 +392,13 @@ class PlaylistsBrowser(Browser, DisplayPatternMixin):
             uri = uri.encode('utf-8')
             try:
                 sock = urlopen(uri)
-                f = NamedTemporaryFile()
-                f.write(sock.read())
-                f.flush()
                 if uri.lower().endswith('.pls'):
-                    playlist = parse_pls(f.name, library=library)
+                    playlist = parse_pls(sock, name, library=library)
                 elif uri.lower().endswith('.m3u'):
-                    playlist = parse_m3u(f.name, library=library)
+                    playlist = parse_m3u(sock, name, library=library)
                 else:
                     raise IOError
                 library.add(playlist.songs)
-                if name:
-                    playlist.rename(name)
                 self.changed(playlist)
                 Gtk.drag_finish(ctx, True, False, etime)
             except IOError:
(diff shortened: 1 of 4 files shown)


declension:
Hi @lazka...  couldn't work out the problem i caused here in Windows, even with increasing debug logging 🙄

What's the best way to run or at least debug the Windows tests locally on Linux? I tried Dockerising in a similar way to Travis, with Wine, but apparently this isn't advised. Do you use a VM directly now or is local wine ok too?

Thanks!

lazka:
clippy

lazka:
I'll have a look. I use an msys2 environment as described here: https://github.com/quodlibet/quodlibet/blob/master/win_installer/README.rst

Either on a Windows machine or in a VM..

lazka:
Some random fixes which makes things pass here: https://bpaste.net/show/6f00f6a2a94e

declension:
Thanks @lazka  :). Spent far too long dealing with Windows VM / msys2 issues (32bit Windows was a problem, pip2 was a problem, some ming32/64 paths were a problem, eventually ran out of time).

Anyway have pushed those changes verbatim, want to check they're OK on py2 etc.
Still not sure _why_ it was failing (nothing added on Windows) but I can experiment on this branch and see what it was that breaks things (and then squash or close and re-PR)

Thanks

lazka:
uh, one was wrong

```
diff --git a/quodlibet/quodlibet/browsers/playlists/util.py b/quodlibet/quodlibet/browsers/playlists/util.py
index 88a832916..aeafdac5f 100644
--- a/quodlibet/quodlibet/browsers/playlists/util.py
+++ b/quodlibet/quodlibet/browsers/playlists/util.py
@@ -130,7 +130,7 @@ def _name_for(filename):
 
 def _dir_for(filelike):
     try:
-        return os.path.dirname(text2fsn(filelike.name))
+        return os.path.dirname(path2fsn(filelike.name))
     except AttributeError:
         # Probably a URL
         return text2fsn(u'')
```

type mess..

lazka:
> Spent far too long dealing with Windows VM / msys2 issues (32bit Windows was a problem, pip2 was a problem, some ming32/64 paths were a problem, eventually ran out of time).

I've went through the bootstrap myself yesterday because my old environ is in a mix of py35/36 atm. Pushed some small fixes then. Never used a 32 bit Windows, so not sure if there are any additional problems.

declension:
Not had much more time to look again, but pushed that fix too, thanks

lazka:
On master you can now do:

```
./docker.sh win32.py2
python setup.py test
py.test tests/test_util.py
...
```

declension:
Thanks. Not sure how Travis on _this_ branch has changed, must be globally configured somehow. Anyway looks like using text-based processing [now fails on Win + PY3 (only)](https://travis-ci.org/quodlibet/quodlibet/jobs/244661617#L2466), and only with the difficult (unicode-heavy) entry I added.

I guess reverting to bytes everywhere is the only way ahead here, this PR has taken way too long already (and thanks again)

Issue #2432, opened 2017-06-12
Plugin: Seek points


Meriipu:
Would a plugin that stores points in a field for each track, and whenever that track is played, it starts playing from and continues to the next songs at the timestamps A,B be feasible?

Some tracks have intros I would like to skip, but without the whole ordeal of reencoding it to a new track.

declension:
Well this sounds like a use for bookmarks for a start.

And based on that - yes.

Meriipu:
I will try looking into it though I might not get very far.

Meriipu:
I have a pretty trashy implementation that kind of works
Apparently seeking immediately on song_started does not work?

https://github.com/Meriipu/quodlibet/blob/seekpoints/quodlibet/quodlibet/ext/events/seekpoints.py

I use a short callback timeout for point A (start_from).
I check for whether the current position is past point b (end_at) each tick (second, I guess?)

It reads the bookmarks SeekPoint_A and SeekPoint_B if they exist, otherwise do nothing. 

This is probably very not clean at all, and I have not tested it much.

lazka:
>  Apparently seeking immediately on song_started does not work?

Try now

Meriipu:
Seems to work, thanks.

Meriipu:
There is likely room for improvement like a cleaner way to avoid unnecessary on_tick-checks when there is no B-point set to compare the current position to (than the method that was used in the implementation), though if it looks ok I suppose this can be closed.

lazka:
The once a second check isn't that bad as long as it doesn't block the main loop too much. We wake up once second anyway for various reasons..

Meriipu:
I had not reverted it to using the once a second check yet.
So at the moment when a track starts, it starts the tracker, checks for point A and seeks to it if it exists. Then on the first tick it checks whether point B exists, and if it does not, it destroys the tracker until next track-start. 

If a check "if not self._seekpoint_B: return" in _on_tick is a really minor impact when B is None, then it would probably simplify the code a bit to just use enabled and disabled to start/stop the tracker as suggested.

lazka:
> If a check "if not self._seekpoint_B: return" in _on_tick is a really minor impact when B is None, then it would probably simplify the code a bit to just use enabled and disabled to start/stop the tracker as suggested.

I think doing the work in every tick shouldn't be a problem. Feel free to change it if you want.

PR #2437, opened 2017-06-15
Rough implementation of automatic seekpoint using bookmarks

Meriipu:
https://github.com/quodlibet/quodlibet/issues/2432

This could probably be tidier in ways but it seems to work well enough for me.
Set bookmark names to use for points (A,B) in plugin config if desired (defaults are A,B). 
Create bookmarks A and /or B for a track, whenever plugin is enabled and track is played, it skips to point A if it exists and stops after point B if it exists.

Are there cleaner ways to detect whether a time tracker has been destroyed? 
I destroy it if point B does not exist, as there is no point I can see checking every tick if the current position is past point B if point B does not exist.

### DIFF
--- a/quodlibet/quodlibet/ext/events/seekpoints.py
+++ b/quodlibet/quodlibet/ext/events/seekpoints.py
@@ -1,0 +1,141 @@
+# -*- coding: utf-8 -*-
+
+#This is free and unencumbered software released into the public domain.
+#
+#Anyone is free to copy, modify, publish, use, compile, sell, or
+#distribute this software, either in source code form or as a compiled
+#binary, for any purpose, commercial or non-commercial, and by any
+#means.
+#
+#For more information, please refer to <http://unlicense.org>
+
+from gi.repository import GLib, Gtk
+from quodlibet import _, app
+from quodlibet.qltk.entry import UndoEntry
+from quodlibet.plugins import PluginConfigMixin
+from quodlibet import qltk
+from quodlibet.plugins.songshelpers import has_bookmark
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.qltk import Icons
+from quodlibet.qltk.tracker import TimeTracker
+
+class SeekPointsPlugin(EventPlugin, PluginConfigMixin):
+    """The plugin class."""
+
+    PLUGIN_ID = "Seekpoints"
+    PLUGIN_NAME = _("Seekpoint bookmarks")
+    PLUGIN_ICON = Icons.GO_JUMP
+    PLUGIN_CONFIG_SECTION = __name__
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/ext/events/seekpoints.py
```diff
@@ -0,0 +1,139 @@
...
+       "bookmark names the plugin looks for when deciding whether to seek.")
+
+    CFG_SEEKPOINT_A_TEXT = "A"
+    CFG_SEEKPOINT_B_TEXT = "B"
+    DEFAULT_A_TEXT = "A"
+    DEFAULT_B_TEXT = "B"
+
+    def enabled(self):
+        self._SeekPoint_A, self._SeekPoint_B = self._get_SeekPoints()
+        self._try_create_tracker(force=True)
```
lazka: You should be able to just use enabled() and disabled() for creating/removing TimeTracker.

quodlibet/quodlibet/ext/events/seekpoints.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
+#
+# For more information, please refer to <http://unlicense.org>
```
lazka: Please use the following license to fix the tests:
```
# This program is free software; you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation; either version 2 of the License, or
# (at your option) any later version.
```

quodlibet/quodlibet/ext/events/seekpoints.py
```diff
@@ -0,0 +1,139 @@
...
+    def _on_tick(self, tracker):
+        if not self._SeekPoint_B:
+            self._try_destroy_tracker()
+            return
+
+        time = app.player.get_position() // 1000
+        if self._SeekPoint_B <= time:
+            self._seek(app.player.info("~#length"))
+
+    def _get_SeekPoints(self):
```
lazka: Please use lowercase function names and variable names. https://www.python.org/dev/peps/pep-0008/#function-names

quodlibet/quodlibet/ext/events/seekpoints.py
Meriipu: Can the tracker be stopped and started otherwise so to not make on_tick callbacks for tracks that do not have a B-seekpoint, or is the overhead from that minimal?

lazka:
Thanks

Issue #2446, opened 2017-06-21
Clarify purpose of browsers/playlists/main.py: __check_current


Meriipu:
In browsers/playlists/main.py, what does the method __check_current do (or rather, what is its purpose) ?
https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/browsers/playlists/main.py#L304
```
    def __check_current(self, model, path, iter):
        model, citer = self.__selected_playlists()
        if citer and model.get_path(citer) == path:
            songlist = qltk.get_top_parent(self).songlist
            self.activate(resort=not songlist.is_sorted())
```

Not running it seems to fix #2253 for me, but I noticed at least one thing that breaks:
I am in the playlist browser (ctrl+2), and have a playlist selected. If I drag and drop a track to the playlist I am in from the tracklist in the playlist (so to add a duplicate track), the playlist does not update with the new track until I select a different playlist. 

Other than that I have not noticed anything that breaks. 

__check_current (only?) seems to be called as a result of the signal row-changed.

lazka:
For such cases it's always good to have a look at the blame output: https://github.com/quodlibet/quodlibet/blame/master/quodlibet/quodlibet/browsers/playlists/main.py

row-changed is not the best place for this as it's usually just meant for updating the view. +1 for getting rid of it.

Meriipu:
I have no idea how it could be changed though.
If the only thing that breaks is that the playlist is not reloaded when it grows (or shrinks?) then maybe an extra check for whether the playlist actually grew or shrunk, or a separate callback for that and removing this.

I am not very into gtk concepts like signals, eventsl and updates at all though.

Meriipu:
since it was removed with that commit I suppose this can be closed unless maybe the reason it was there in the first place resurfaces.

Issue #2253, opened 2017-02-19
Playlists browser clears selection on track change if playlist is selected


Meriipu:
I do not know if it is intended, but in the playlist mode (ctrl+2) having disabled the behaviour of jumping to the playing song automatically seems to have no effect.

Regardless of if a different track further down in the list is selected or not, the currently playing track is in view, or track changes are automatic or manual - It jumps (shifts the view) to the next song.

Meriipu:
I found a pretty silly workaround but it works

In the main view, add the ~Playlists column. Then just search for the playlist name.
It is still a bug though I think unless I have misunderstood something.

Meriipu:
Is this intended behaviour? I have not used the playlist tab because of this.

declension:
For me, the track _selection_ disappears on track change - though the currently playing highlight (in light grey on my theme) correctly moves to the new playing track. This _doesn't_ happen on the search browser, and is definitely a bug (possibly long-standing).

@Meriipu if that's _not_ what you meant please describe more exactly what's happening and what you're expecting (maybe with screenshots) and I'll update this

Meriipu:
@declension it is kind of similar but not what I initially reported

So despite the option 
preferences>song list>jump to playing song automatically 
being disabled:
1) Go to playlist view
2) Select a playlist with enough songs to be able to scroll
3) Start playback of a song 
4) Scroll down so the playing song is not centered
5) Select "next track", let a song finish, or just click another track that is not in the center of the current view
6) the view should jump so that the next song is centered (despite this option being disabled)

Meriipu:
I have been trying to look into this but it seems the list of songs in the playlists-browser are cleared and recreated each time a track finishes playing ?

So if this is the case, would a workaround be to store the last position (percentage of scroll-wise) in the list and then scroll to it?

I do not really understand the playlist browser. If you start playing a track by selecting it from the main browser, it seems a temporary playlist is constructed or something like that when you swap to the playlist view, but once a track finishes playing after having swapped to the playlist view, it gets removed from this temporary list
So the number of songs in the list gradually dwindle down to none if you remain in the playlist-browser after having started playing tracks from the main browser but without selecting an actual playlist after having swapped to the playlist browser (or something like that).

Meriipu:
```
diff --git a/quodlibet/quodlibet/qltk/songlist.py b/quodlibet/quodlibet/qltk/songlist.py
index b7cc85758..bdb3ea066 100644
[...]

PR #2449, opened 2017-06-25
Attempt at fixing various playlist-updating issues

Meriipu:
related to issues #2446 #2253

This appears to fix (in the playlist browser):
* A track change causing the current selection to be cleared
* A track change causing the playlist to scroll to the new track regardless of whether Jump on track change in config is set or not
* At least for me, when a playlist had duplicate tracks and one started playing, the last one would start playing, potentially skipping far into the playlist. 

now I have no idea if anything breaks so I suppose this pull request would serve better as some discussion of some sort (or if any of these changes are horribly messy).

I would like to add that I have no idea what I am doing and I worry something fairly unobvious breaks as a result of this.

as a sidenote I have not been able to run tests locally, the tests stop after a short while at 
```
tests/test_browsers_paned.py .Trace/breakpoint trap
#(with this in my /var/log/messages, which I could not make sense of:)
#Jun 25 10:22:40 Meowth kernel: traps: python[17475] trap int3 ip:7fb763cb5101 sp:7fffc9ecb140 error:0
```

### DIFF
--- a/quodlibet/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/quodlibet/browsers/playlists/main.py
@@ -259,8 +259,8 @@ class PlaylistsBrowser(Browser, DisplayPatternMixin):
         view.connect('popup-menu', self.__popup_menu, library)
         view.get_selection().connect('changed', self.activate)
         model = view.get_model()
-        s = model.connect('row-changed', self.__check_current)
-        connect_obj(self, 'destroy', model.disconnect, s)
+        #s = model.connect('row-changed', self.__check_current)
+        #connect_obj(self, 'destroy', model.disconnect, s)
         self.connect('key-press-event', self.__key_pressed)
 
     def __create_cell_renderer(self):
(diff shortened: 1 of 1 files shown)


Meriipu:
One thing that this breaks is that if a track in one playlist is activated and the user swaps to another playlist (that does not necessarily also contain the track), the next track will be played from that new selected playlist once the previous track ends. 

uuuu

Meriipu:
It does not work at all in head, but some interactions have the same result in this patch too:

* Make a playlist with duplicate tracks
* Start one track, it will start playing the one you clicked (as opposed to the last instance)
* Click a different playlist then click the original, all while the track is still playing
* The last track is now selected; the original playlist "forgot" which instance of the track was playing.

lazka:
I'm OK with DnDing something onto the playlist not updating the song list (that's how it happens elsewhere anyway).

Just removing the code without replacement (no commenting) is fine.

For the other issues, best to tackle them one by one in a separate issue/PR.

Meriipu:
So should I keep or remove the added lines for updating the songlist (is this the list of tracks that are seen in the playlist browser when a playlist is selected?) 

The original behaviour in the playlist browser is that it updates the songlist with the new DnD track. 
The behaviour in this commit is that it does not do the scrolling/clearing selection-stuff, but it still updates the current songlist with the DnD track.

I have not extensively tested combinations of browsers like from the menubar Browse>Open Browser>Playlists, in combination with a second browser in the main window, though some of those interactions already seem to have issues in the current version too (#2454)

lazka:
> So should I keep or remove the added lines for updating the songlist (is this the list of tracks that are seen in the playlist browser when a playlist is selected?)

it seems to work fine here.

thanks!

Issue #2251, opened 2017-02-19
Allow custom ellipsizing for FS songlist columns


Meriipu:
Is there some way to change the way the path/fullname is split in the column?

When the column is too narrow to contain the entire string, I prefer it to display the last part of the path rather than both the beginning and the end. 

As it is now I only see the directory all my music is in (mostly identical for all tracks) in the left part,
and the filetype and name of the track but not the album/artist directory on the right. 

Would an option to change it to display as ".../artistalbum/01 - trackname.flac" be too specific?

lazka:
Hm, this seems a bit too edge case for me to get its own option. Any thoughts from someone else?

@Meriipu: Given ~dirname, ~filename, ~basename. Assume you could use multiple columns, would changing of the default ellipsizing for any of the tags improve the situation?

Meriipu:
The current ellipsizing in the middle of the full path looks cleaner, so it makes sense as a default.

I have it set to ellipse at the beginning of the path in my current build, so that I usually see the directory the file resides in and maybe one directory up, rather than just the filename + top level directories. This looks pretty unclean though.

I think what I would prefer is having the filename take a constant width, ellipsized in the middle if appropriate. That way the trailing slashes of the leaf-directories should line up -- and then also ellipsize the beginning of the path so that the directories closest to the leaf are likely to be visible.

This sounds pretty edge case to me though yes.

CreamyCookie:
As the associated pull request was closest, I'm also closing this.

It definitely feels a bit too edge case though I fully understand the interest in improving the situation. Maybe there's even a way to make all parties happy without a setting. Anyway, thanks for the report!

PR #2450, opened 2017-06-25
Added support for custom ellipsizing in filepath column

Meriipu:
The configuration feels a tiny bit unclean. 

Implements https://github.com/quodlibet/quodlibet/issues/2251

### DIFF
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -214,8 +214,6 @@ INITIAL = {
         # the format of the timestamps in DateColumn
         "datecolumn_timestamp_format": "",
 
-        # the ellipsizing mode to use for SongColumn filepaths
-        # "1" Start, "2" Middle, "3" End
         "ellipsizing_mode": "2",
     },
 
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -9,12 +9,14 @@
 # (at your option) any later version.
 
 import shutil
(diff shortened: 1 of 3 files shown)


Meriipu:
oh I think I messed up the direction of my merge

Meriipu:
I can look into these when I have time (it might be a while). The suggested improvements all make sense.

Meriipu:
so I tried with an enum.Enum at first but it turned out a bit too complex for me at least. 
Putting the dict (pretend-enum) of modes in config.py does not seem like the wisest decision, as it requires importing pango and stuff.

is defining the text version ("MIDDLE", hardcoded) in config.py a problem?
the example lines seem tricky/too specific to me to include differently. 

other than that there should be fewer magic constants floating around now...

Meriipu:
I ran into a KeyError when trying to run the last commit with a config file based on the earlier commit (that used numbers rather than the string versions). 

I do not think that is something that will occur if the value was set for the first time to be one of the string ones but I suppose it might not hurt to check

Perhaps this is also a bit too niche so that it might be better to just close it unless somebody else really hates the default ellipsizing behaviour of middle. Applying this patch personally for my own use to my builds works fine for me either way.

declension:
Cleaning up old PRs 

> Perhaps this is also a bit too niche so that it might be better to just close it unless somebody else really hates the default ellipsizing behaviour of middle. Applying this patch personally for my own use to my builds works fine for me either way.

Thanks, will close for now then.

Issue #2434, opened 2017-06-14
New feature: play audio from video .mp4 .flv .mkv


legluondunet:
Hello,

I have a lot of videos music on my hard disc, mixed with my mp3 files. 
Several are ripped from Youtube, there are mp4 and flv files video format, sometimes mkv. 
It will be great if Quad Libet could play at least audio , better video, from this files.

LGDN

lazka:
mp4 should work.

mkv/flv is currently not implemented.

legluondunet:
I just ripped a youtube video in mp4 format, it works.
But I choose the mp4 format:

```
$ youtube-dl -F https://www.youtube.com/watch?v=_ePlKJfWtaM
[youtube] _ePlKJfWtaM: Downloading webpage
[youtube] _ePlKJfWtaM: Downloading video info webpage
[youtube] _ePlKJfWtaM: Extracting video information
[youtube] _ePlKJfWtaM: Downloading MPD manifest
[info] Available formats for _ePlKJfWtaM:
format code  extension  resolution note
139          m4a        audio only DASH audio   48k , m4a_dash container, mp4a.40.5@ 48k (22050Hz), 1.93MiB
249          webm       audio only DASH audio   55k , opus @ 50k, 2.13MiB
250          webm       audio only DASH audio   74k , opus @ 70k, 2.84MiB
140          m4a        audio only DASH audio  128k , m4a_dash container, mp4a.40.2@128k (44100Hz), 5.16MiB
171          webm       audio only DASH audio  136k , vorbis@128k, 5.27MiB
251          webm       audio only DASH audio  145k , opus @160k, 5.64MiB
278          webm       192x144    144p   80k , webm container, vp9, 30fps, video only, 2.96MiB
160          mp4        192x144    DASH video  115k , avc1.4d400c, 15fps, video only, 4.48MiB
242          webm       320x240    240p  187k , vp9, 30fps, video only, 6.82MiB
133          mp4        320x240    DASH video  247k , avc1.4d400d, 30fps, video only, 9.81MiB
243          webm       480x360    360p  342k , vp9, 30fps, video only, 12.55MiB
134          mp4        480x360    DASH video  612k , avc1.4d401e, 30fps, video only, 15.85MiB
244          webm       640x480    480p  655k , vp9, 30fps, video only, 23.52MiB
135          mp4        640x480    DASH video 1126k , avc1.4d401e, 30fps, video only, 38.14MiB
[...]

PR #2458, opened 2017-06-28
tests: test decoding h264/h265 videos on macOS/Windows. See #2434

lazka:
And make the test fail if decoding fails.

It's a common usecase to play videos downloaded from youtube etc
so make sure we ship the right codecs.


lazka:
macOS is missing gmedec/opusdec

Issue #2352, opened 2017-04-06
Build Docker images in advance


declension:
It's great to have Dockerised build containers after #2222. However, quite a lot of time is spent in CI _building_ the Docker images before using them, yet they change very rarely.

It would be more efficient to push these to our own Docjer Hub space, say, and pull them (caching at Travis should work here too).

To keep them up-to-date the first step would be manual pushes (authenticated). Then we could probably work out some CD process for this itself so that any changes to those `Dockerfile` files would be published to the hub (before any real build, if still in the QL codebase, or these could live in separate repo).

lazka:
I'm not sure if it's worth it with the added complexity.

lazka:
I've moved all tests to CircleCI now, and all tests take 5.5 minutes there, including building the docker images. Might just be luck re user count/time of day etc, but I don't remember travis being that fast ever.

lazka:
docker images are now cached @ circleci

lazka:
It doesn't seem to help that much in the overall CI time (circleci might have more load atm, so not sure), but the first build finishes/fails 2 minutes earlier.

Any thoughts? Can this be closed?

declension:
Sorry - yeah that's fine I. If I get some time I might try pushing a couple of images based on those `Dockerfile`s and experiment in a branch, but... :)

PR #2459, opened 2017-06-30
circleci: retry docker build

lazka:
it fails sometimes due to network errors


declension:
Cool. However I still think #2352 is a good idea (massively reduces the networking load too)

Issue #138, opened 2015-03-14
Quod Libet doesn't match FreeDesktop directories specs


lazka:
Original [issue 138](https://code.google.com/p/quodlibet/issues/detail?id=138) created by thibaut.bethune on 2009-02-24T23:52:29.000Z:

I've tried Quod Libet 2.0 on Ubuntu 9.04 alpha 4
and it seems that Quod Libet places its configuration files in
/home/.quodlibet directory which doesn't match FreeDesktop directories specs :

The default for $XDG_CONFIG_HOME is $HOME/.config, the default for
$XDG_DATA_HOME is $HOME/.local/share. So all applications should look for
those environment variables and use those default values if the variables
are not set.

This is quite important since it's not possible to easily backup config
&amp; data files if these files are not stored at the right place

See http://www.freedesktop.org/wiki/Specifications/basedir-spec
See also
http://ploum.frimouvy.org/?184-cleaning-user-preferences-keeping-user-data
(main post and comment#8)

jrobeson:
seems like it should be like so:

covers: `$XDG_CACHE_DIR/quodlibet`
control: `$XDG_RUNTIME_DIR/quodlibet`
playlist: `$XDG_DATA_DIR/quodlibet`
lists: `$XDG_DATA_DIR/quodlibet`
config: `$XDG_CONFIG_DIR/quodlibet`

jcrben:
Any movement around this? Would you be happy to take a pull request?

lazka:
No.

I'm for simply moving "~/.quodlibet" to "~/.config/quodlibet" and adding a symlink.
And if possible do that as atomically as possible. We'd need similar functionality for osx, see #1950.

And for cache stuff we can just switch to "~/.cache/quodlibet" without migrating data; we already use that for some cover caching atm.

jcrben:
I guess setting QUODLIBET_USERDIR functions as a workaround, right?

lazka:
> I guess setting QUODLIBET_USERDIR functions as a workaround for now, right?

yeah

ptitjes:
@jcrben Are you working on this ?

jcrben:
No, not using quod libet right now

Socialdarwinist:
[...]

PR #2466, opened 2017-07-05
Extremely lazy fix for #138

sauyon:
This just uses $XDG_CONFIG_HOME/quodlibet as the default directory unless it doesn't exist and ~/.quodlibet does.

I'm willing to go through and actually put files where they should be if that's necessary.

EDIT: For #138.

### DIFF
--- a/quodlibet/quodlibet/_main.py
+++ b/quodlibet/quodlibet/_main.py
@@ -117,7 +117,9 @@ def get_user_dir():
 
     if os.name == "nt":
         USERDIR = os.path.join(windows.get_appdate_dir(), "Quod Libet")
-    elif not is_osx():
+    elif is_osx():
+        USERDIR = os.path.join(os.path.expanduser("~"), ".quodlibet")
+    else:
         USERDIR = os.path.join(xdg_get_config_home(), "quodlibet")
 
         if not os.path.exists(USERDIR):
--- a/quodlibet/quodlibet/_main.py
+++ b/quodlibet/quodlibet/_main.py
@@ -118,7 +118,16 @@ def get_user_dir():
     if os.name == "nt":
         USERDIR = os.path.join(windows.get_appdate_dir(), "Quod Libet")
     else:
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/_main.py
```diff
@@ -118,7 +118,16 @@ def get_user_dir():
     if os.name == "nt":
         USERDIR = os.path.join(windows.get_appdate_dir(), "Quod Libet")
     else:
-        USERDIR = os.path.join(os.path.expanduser("~"), ".quodlibet")
+        if 'XDG_CONFIG_DIR' in environ:
```
lazka: Use quodlibet.util.path.xdg_get_config_home

quodlibet/quodlibet/_main.py
```diff
@@ -118,7 +118,16 @@ def get_user_dir():
     if os.name == "nt":
         USERDIR = os.path.join(windows.get_appdate_dir(), "Quod Libet")
     else:
```
lazka: We don't want this change on macOS (quodlibet.util.is_osx), needs a new case using the old path for this.

lazka:
Notice the code quality tests failing. You can run them locally using either "setup.py quality" or "py.test tests/quality"

sauyon:
I'm seeing a bunch of quality errors from stuff in optpackages but I'm assuming I'm to ignore those; can I turn them off?

lazka:
uh, optpackages should get skipped for this (in tests.quality.util.iter_project_py_files())

sauyon:
It might be worth moving the check for QUODLIBET_USERDIR/build type above to speed up launches (slightly).

sauyon:
Ah - it was finding them in `build/lib/quodlibet/optpackages`. Getting rid of that worked fine (maybe it should ignore `root/build` as well?).

lazka:
Ah, right, thanks. I'll have a look.

lazka:
Thanks!

Issue #2479, opened 2017-07-14
Selecting arrays in lists with Ctrl+Shift


Socialdarwinist:
Desired behavior:
I select selectively with Ctrl, and put rows into my selection by Ctrl+Shift. Like when I have an array {1,2,3,4,5,6,7,8,9,10}, I click 2 to select it, then I ctrl+click 4 to have selected 2 and 4, then ctrl+shift+click 8 to have 2 and 4–8 selected. This works in all good file managers and bookmark folders in browsers.

Actual behavior: 2 and 4 and 8 is selected, not also 5, 6 and 7.

Implementing this behavior would substantially increase efficiency of usage. Right now one has, for instance, to remove pieces of an album from a playlist to select them one by one or to search them first in another browser to remove them alltogether.

I am using Quod Libet on Cinnamon on Arch Linux.

CreamyCookie:
Closing this as a duplicate of #3227 

It's a bit newer but uses the new template and I found it a bit more clear. I added part of your description though. Thanks for the report!

PR #2480, opened 2017-07-15
ctrl+shift TreeView range selection

Meriipu:
issue #2479 

So I tried a bit and managed to get something working but it is an awful fix and I would hope there are better ways to do it, hopefully something native to pygtk that just requires changing an option. I am starting to doubt that native functionality exists, though this pull request as it stands now is a bit too awkward to be considered a solution in my opinion. 

On finding out whether it is a native gtk thing or not - firefox has the behaviour described in the issue with ctrl+shift for me in the bookmarks window (bookmarks->show all bookmarks) and qBittorrent has it in the torrent-list, although as far as I know qBittorrent is qt and not gtk.

Thunar does not seem to have it (it is gtk2 yes?), Firefox download list does not seem to have it, nor does my build of gimp in the open-files dialog.  

I looked at qltk/views.py, where the MultiDragTreeView seems to be the closest to defining something relevant (__block_selection), and otherwise inherits from native gtk things (treeview). After looking at the pygtk documentation I could not find any other way than manually setting the range. 

so I store the last selected treeview element, and when a ctrl+shift selection is made try to select the range from the stored element to the clicked element. 

awkward temporary flags to avoid endless recursion (self.__ctrl_shift in progress), as set_select_function apparently can not be changed (back to lambda *args: True) in the middle of a selection without segfaults? 

The actual selection logic might have some awkward elements too.


Meriipu:
I tried to put together a test case as that might be useful even on its own but I can not get the helper _send_button_click_event by passing it coordinates of the rows of the widget as x,y and a shift modifier to actually select anything from what I can tell.

quodlibet/tests/helper.py
```diff
@@ -143,6 +145,28 @@ def send_button_click(widget, button, primary=False, shift=False,
...
+        state |= Gdk.ModifierType.SHIFT_MASK
+
+    assert isinstance(widget, Gtk.Widget)
+    box = widget.get_cell_area(path, widget.get_column(0))
+    wx, wy = widget.convert_tree_to_widget_coords(box.x, box.y)
+    #width, height = widget.convert_tree_to_widget_coords(box.width, box.height)
+    wx += box.width // 2
+    wy += box.height // 2
+    handled = _send_button_click_event(widget, x=wx, y=wy, button=button, state=state)
+    # I did not get the parameter to enable stdout to work...
```
lazka: `py.test -s tests/test_qltk_views.py`

or `assert 0, "foo"` is also helpful

lazka:
I don't know why it doesn't work. One would need to step through the gtk C code. My guess would be that we are missing many more event related events here. At least focus events, enter and leave events etc.

Maybe try to emulate the gtk behavior roughly by doing something like

```
    handled = _send_button_click_event(widget, x=wx, y=wy, button=button, state=state)
    if not handled:
        widget.get_selection().select_path(path)
```

Meriipu:
Though if the behaviour that was to be tested was selecting elements with modifiers active (and then check whether the intended range was selected), emulating it that way would not really test it?

lazka:
Yeah, you'd have to emulate that as well I guess. Event related code in gtk is sadly hard to test :(

Meriipu:
this was pretty much a disaster so closing for now

PR #2481, opened 2017-07-16
undo covergrid hacks

darthoctopus:
this is *considerably* faster for me (tested on Fedora 26).


lazka:
Not the case here, about twice as slow. hmm..

darthoctopus:
I should be more specific: I believe you are referring to performance on resize, and that does appear to be the case. however (at least for me) this also causes scrolling to be heavily CPU-bound (100% CPU utilisation for about half a minute whenever viewport is moved, vs 60 fps scrolling without the hack). I suspect but cannot prove that something is preventing hardware acceleration when these methods are overwritten.

lazka:
I can't reproduce. Any change with the commit 84d95f92a2f on master? That should improve scrolling when covers get loaded.

darthoctopus:
no change to scrolling performance, unfortunately.

It seems that commenting out the redefinition of `do_get_preferred_height_for_width` is sufficient to almost entirely get usable scrolling behaviour back.

lazka:
ok, thanks, I'll have another look.

declension:
Any updates on this PR?

darthoctopus:
update: performance issues seem gone with current git master; must have been a graphics driver or GTK issue.

lazka:
Thanks for testing again

declension:
Closing old / broken PRs - please re-open or resubmit based on current `master` if this is still relevant. Thanks!

PR #2486, opened 2017-07-18
Created new confirmation dialog; cleared some redundancy in dialog usage

Meriipu:
New confirmation dialog in qltk/msg.py for cancel/accept-dialogs that takes title, description, and button label. 
This also changes the plugin prompts (for activating plugins on X many items) in songsmenu to use this for a tiny cleanup in code reuse. 

I have a second branch that adds a confirmation prompt using this new dialog for removing tracks from the library that I could include in this request if that is desirable, or maybe in a second request (though it depends on this). 

I noticed tests/test_qltk_chooser.py which I think could be used to put together some tests too

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
+++ b/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
@@ -11,7 +11,7 @@ import os
 
 from quodlibet.util.songwrapper import SongWrapper
 
-from quodlibet.qltk.songsmenu import ConfirmMultiSongInvoke
+from quodlibet.qltk.songsmenu import confirm_multi_song_invoke
 
 import quodlibet
 from quodlibet import _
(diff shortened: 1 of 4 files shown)


Meriipu:
no idea how I missed it the first time, pretty sure I searched. I can not find any other ConfirmMulti-usages now except ConfirmMultipleSongsAction and ConfirmMultiPlaylistInvoke now though and those are defined elsewhere.

Meriipu:
although how do you run a plugin on multiple playlists?

lazka:
Not sure about the playlists. There is a plan to support selecting multiple playlists, so that might have been a forward-looking implementation.

Looks good. Although in the long run we'll have to move away from blocking dialog methods again as GTK4 wont have a run method for dialogs (no nested main loops).

lazka:
ok to merge?

Meriipu:
should be good from what I can tell unless replacing other dialogs too is desirable

* ConfirmMultipleSongsAction is the only one with the same ConfirmMulti-naming style left I can find, and it seems to add three buttons so I decided to leave that as is. 

* A few others use the delete icon, so maybe if icon to use is passed as argument

* maybe qltk/ratingsmenu.py:ConfirmRateMultipleDialog

but yes other than that.

Issue #2200, opened 2017-01-15
Filter function (e.g. max, min) doesn't work correctly with lastplayed


CreamyCookie:
![screenshot from 2017-01-15 14-29-23](https://cloud.githubusercontent.com/assets/3063858/21962945/a9bac7f8-db30-11e6-9076-90bdaba8db33.png)

Note that "Mittwoch" is German for "Wednesday".

The part of the filter / search that doesn't seem to work is `#(lastplayed:min > 9 month ago)`. Unless I'm missing something, the selected album should not be there, since all songs have been played recently. 

Ubuntu 16.04
Quod Libet 3.8.0
Python: 2.7.12
Mutagen: 1.36
GTK+: 3.18.9 (X11)
PyGObject: 3.20.0
GStreamer: 1.8.2.0

declension:
Quite a query there :)

I first thought that's actually correct, as the minimum _time_ for that album (some time on Wednesday) is indeed larger (newer) than nine months ago but now I see that it seems to be the duration after all `#(lastplayed < 7 days ago)` is the standard _recently played_ filter.

It does feel like there's a bug with the aggregates. Testing here is proving hard either way though; remember that a min for the timestamps is often 0 (i.e. 1970) which can be confusing.

@CreamyCookie perhaps you can try experimenting with the Python Console plugin: select the album you think is buggy and use the new(ish) `col` variable in the console, e.g. `col('~#lastplayed:min')` to see what the results for the _collection_ are coming out as. It can sometimes be quite helpful.

CreamyCookie:
> it seems to be the duration after all #(lastplayed < 7 days ago) is the standard recently played filter.

It's a bit unintuitive that `~#lastplayed` returns a Unix timestamp of when the song was last played, but in a query it's the duration since it was last played. 

Maybe `#(lastplayed in 7 days)` would be more intuitive, as then to you'd immediately know that in this context it's a duration. Then when you want to find only the songs that haven't been played in the last 7 days (i.e. `#(lastplayed > 7 days ago)`), you'd just negate the query.

> perhaps you can try experimenting with the Python Console plugin

Good idea! Haven't used it before, but that's a pretty useful plugin for debugging. Here's the output:

```python
>>> col('~#lastplayed:max')
1484149887
>>> col('~#lastplayed:min')
1484148684
>>> import datetime
>>> datetime.datetime.utcfromtimestamp(int('1484148684'))
[...]

PR #2504, opened 2017-07-25
Fix results for queries with time tags and aggregate functions

slosd:
This fixes issues as reported in #2200. 

Tags related to times weren't correctly identified when combined with an aggregate function which produced wrong query results.

### DIFF
--- a/quodlibet/quodlibet/query/_match.py
+++ b/quodlibet/quodlibet/query/_match.py
@@ -269,7 +269,10 @@ class NumexprTag(Numexpr):
         else:
             num = data(self._ftag, None)
         if num is not None:
-            if self._ftag in TIME_TAGS:
+            # Strip aggregate function from tag
+            func_start = self._ftag.find(":")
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/query/_match.py
```diff
@@ -269,7 +269,10 @@ def evaluate(self, data, time, use_date):
         else:
             num = data(self._ftag, None)
         if num is not None:
-            if self._ftag in TIME_TAGS:
+            # Strip aggregate function from tag
+            func_start = self._ftag.find(":")
+            tag = self._ftag[:func_start] if func_start >= 0 else self._ftag
```
lazka: `tag = self._ftag.split(":", 1)[0]` is handy for such things, but this is ok too.

lazka:
Please add a test for this

lazka:
Thanks

slosd:
👍 
Just as a reminder, this fixes #2200

declension:
Thanks @slosd

PR #2507, opened 2017-07-28
Soundcloud: my tracks and more

declension:
* Add new category (filter) for my (soundcloud) tracks
 * Gather and persist to config the `user_id` from API
 * Move config handling entirely into the API class rather than in
 two places which was a bit weird
 * Add a new `soundcloud_user_id` QL tag with mapping from API
 * Use the known value to template a QL filter for my tracks whilst
 actually doing the query in the background for all API results
 (there should be 1:1 here though, but the browser works by
 overlaying the QL filter on all the cached results so it kinda has
 to do this)
 * BUGFIX: ordering of filters (categories)
 * BUGFIX: async updating issue by connecting library properly to
 browser (assumes everything API is a change)
 * BUGFIX: actually disable invalid items (favourites, my tracks)
 when user is logged out and re-enable on login.

### DIFF
--- a/quodlibet/quodlibet/browsers/soundcloud/api.py
+++ b/quodlibet/quodlibet/browsers/soundcloud/api.py
@@ -13,7 +13,7 @@ from quodlibet import util, config
 from quodlibet.util import website
 from quodlibet.util.dprint import print_w, print_d
 from quodlibet.util.http import download_json, download
-from quodlibet.compat import urlencode, iteritems
+from quodlibet.compat import urlencode, iteritems, text_type
 
 from .library import SoundcloudFile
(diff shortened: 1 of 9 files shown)


lazka:
:+1:

lazka:
only minor thing is I got a "you are connected" dialog when first activating the browser

declension:
> only minor thing is I got a "you are connected" dialog when first activating the browser

Yes, I thought about that... in the end because it's a one-off when you "upgrade" to this version (it goes to query for your `user_id`, then persists it to config) I figured it's not a big deal...

lazka:
Yeah, not a big deal, I just thought I'd mention it.

PR #2509, opened 2017-07-30
 Update song order in song list when modified

slosd:
With these changes the song list order is updated if a song is modified. I personally would like this behaviour so I can for example sort by laststarted or lastplayed during randomized playback and have the list reflect the history of plays without having to manually toggle the sort column.  

Generally this works well, occasionally though when a song is double-clicked and consequently modified/sorted (e.g. laststarted) the selection messes up. In addition to the clicked song the song originally following it is selected too. I was unable to find the cause of this. The only thing I know is that the selection changes after `SongList.__song_updated` (qltk/songlist.py).

Incidentally these changes also allowed me to improve `SongList.add_songs` so it no longer has to re-sort all songs when new ones are added.

### DIFF
--- a/quodlibet/config.py
+++ b/quodlibet/config.py
@@ -101,6 +101,11 @@ INITIAL: Dict[str, Dict[str, str]] = {
         "column_expands": "",
     },
 
+    "song_list": {
+        # Automatically re-sort song list when tags are modified
+        "auto_sort": "true",
+    },
+
     "browsers": {
 
         # search bar text
--- a/quodlibet/qltk/prefs.py
+++ b/quodlibet/qltk/prefs.py
@@ -61,11 +61,16 @@ class PreferencesWindow(UniqueWindow):
         def __init__(self):
             def create_behaviour_frame():
                 vbox = Gtk.VBox(spacing=6)
-                c = CCB(_("_Jump to playing song automatically"),
+                jump_button = CCB(_("_Jump to playing song automatically"),
                         'settings', 'jump', populate=True,
                         tooltip=_("When the playing song changes, "
                                   "scroll to it in the song list"))
-                vbox.pack_start(c, False, True, 0)
+                autosort_button = CCB(_("Sort songs when tags are modified"),
+                        'song_list', 'auto_sort', populate=True,
+                        tooltip=_("Automatically re-sort songs in the song list when "
+                                  "tags are modified"))
+                vbox.pack_start(jump_button, False, True, 0)
+                vbox.pack_start(autosort_button, False, True, 0)
                 return qltk.Frame(_("Behavior"), child=vbox)
 
             def create_visible_columns_frame():
--- a/quodlibet/qltk/songlist.py
+++ b/quodlibet/qltk/songlist.py
@@ -750,16 +750,24 @@ class SongList(AllTreeView, SongListDnDMixin, DragScroll,
     def _sort_songs(self, songs):
         """Sort passed songs in place based on the column sort orders"""
 
+        order = self.get_sort_orders()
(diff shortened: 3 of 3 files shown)


lazka:
Generally a good idea, but moving from a comparison key to a comparison function looks like it would have an impact on the sorting performance and that part is one of the most performance critical ones in QL.

slosd:
Took me a while to get back to this, but here it is. 

The performance hit from the comparison function was quite noticeable. This new implementation has virtually no impact on regular sorting while providing the same advantages. 

These are the results from cProfile:

version | ncalls | tottime | percall | cumtime | percall | filename:lineno(function)
-- | - | - | - | - | - | - 
master |	7 |	0.000 |	0.000 |	1.239 |	0.177 |	songlist.py:738(_sort_songs)
first |	7 |	0.000 |	0.000 |	7.340 |	1.049 |	songlist.py:738(_sort_songs)
second |	7 |	0.000 |	0.000 |	1.226 |	0.175 |	songlist.py:739(_sort_songs)

_master_ is the version this branch is based on. _first_ is the implementation using a comparison function. _second_ is the current version in this branch.

slosd:
I really would like to see this merged. Any feedback on the changes?

bassstorm:
@lazka any chance to review this one please

slosd:
@lazka @declension 
Please, can we move forward with this pull request in any way? It has just been sitting here for more than 2 years now.

Meriipu:
I am slightly leaning towards preferring to have an option to turn it off to be honest
though I have not tried it out enough to tell whether my intuitive feeling about how it would work is right (and consequently whether there are parts of the list updating itself after it has been sorted that I would dislike).

slosd:
For the most part an update would be immediately following an action by the user (e.g. after changing title, artist or other tags). Rows may have moved such that they are no longer visible. Maybe the rows which were moved should scroll back into view, but that only helps when they get moved to positions close to one another.

Updates can occur without an action of the user when lastplayed, laststarted etc. or rating (by the auto-rating plugin) are modified, but obviously only if the user has sorted the song list by one of those properties. In that case auto-sorting to reflect the changes seems like the desired behavior to me.

Meriipu:
> In that case auto-sorting to reflect the changes seems like the desired behavior to me.

not necessary if you view sorting by a column as a one-time transformation to bring the playlist from one state into a new state, and then wanting to have it stay in that new state until explicitly sorted again.

When autoupdating, sorting by some increasing value like last-updated would make the tracks earlier in the list never play because the tracks below (in the worst case, two tracks) take turns jumping to the end of the list, so it would be stuck playing two tracks. 

it is a pretty big change in behaviour that a lot of users might have gotten used to.

slosd:
> When autoupdating, sorting by some increasing value like last-updated would make the tracks earlier in the list never play because the tracks below (in the worst case, two tracks) take turns jumping to the end of the list, so it would be stuck playing two tracks.
It's correct that when sorted by "lastplayed" the last two tracks are played alternately. I don't get why tracks earlier in the list would be played instead without this change. Do you mean when playback is in repeat mode?

I'll gladly add an option for this but first I'd like to have some indication of whether this actually has a chance to be merged or not.

declension:
Cleaning up old PRs. Sorry about delays. I don't have strong opinions but it seems like a good change for _most_ circumstances especially given the performance worry seems fixed. 

But yes, safest to have a preferences option then I think we can merge (CC @lazka though)?

slosd:
@declension Added a checkbox to the preferences. Let me know what you think.

declension:
Thanks @slosd !

Issue #237, opened 2015-03-14
Copy album art (folder.jpg\cover.jpg) when copying\renaming files


lazka:
Original [issue 237](https://code.google.com/p/quodlibet/issues/detail?id=237) created by still.ryan on 2009-06-23T18:56:28.000Z:

It would save a lot of time for people who use one album art image instead
of embedded album art if you were able to copy the relevant (or
user-defined) image file to the destination directory alonb with the music
files.

CreamyCookie:
Closed as a duplicate of the clearer and more general (not just album art) #2751.

PR #2511, opened 2017-08-02
Album art mods

elbeardmorez:
Hi,

**Here I aim to improve the Album Art Downloading experience**

This is achieved (certainly it is/was for me ..but hopefully this will affect others usage positively too) via this patch set which i split into two parts. Please read the notes having looked at the pictures for context in each instance.

**1. Custom filename for output**
-commit: allow configurable custom files names for downloaded art

image: [pre custom filename](http://picpaste.com/9781e0a66574e7f404efed35cd33ca59.png)
image: [post custom filename](http://picpaste.com/6cba6319ae0e5327188cfebc80b2efb2.png)
_**note:**_ the main quodlibet prefs now allows the playbar's cover image to be selected based on a custom filename ..great ..but somewhat useless to me if i name all my 'covers' 'front', and yet i cannot download via the downloader to that chosen name, for me 'front.jpg'.

So to be clear.. **with** the changes, I **can** set custom filenames for the download target, **without**, I **cannot**. That is why I wish this to be included in quodlibet.

**2. Search string clarity and custom search sizes**
-commit: expose album art downloader's blackbox
-commit: expose album art downloader query results limit 

image: [search changes](http://picpaste.com/432f981dbcb4393ecab2bace0c1a4588.png)
_**note:**_ the changes are in red. clearly, we would now have the ability to switch between 'raw' input search string, and the 'clean' (..ed via blackbox) string as per the original, pre-commit setup

image: [original search example](http://picpaste.com/65e0b228818c55590d1e97cabae05765.png)
_**note:**_ here is the original search results for a 'American Pie: The Wedding' search.
_**note2:**_ do try to ignore the **correct** top match in this pic as at the time I did this work the quodlibet Amazon service was down, do you (Lazka) remember the poke from declension about this last month? ..that was me poking declension ..and that's where this all started! thus ONLY Discogs results were returned, they were terrible/useless and limited in number and I wanted to know why!

image: [new 'raw' search](http://picpaste.com/0626e1259519f0eea57237cd69bf1b23.png)
_**note:**_ the original 'blackbox' / 'cleaned' search is below this 'raw' search. it is now totally obvious as to why the original search failed ..the cleaning routine for the search (the blackbox) was overzelous. now we recover better matches from Discogs, and more of them if we want!

So to be clear.. **with** the changes, I **can** get the search results I want from both cover providers, **without**, I **cannot**. That is why I wish this to be included in quodlibet. Given quodlibet is (IMO) pitched at a slightly more advanced user, I don't think they'd baulk at exposure to this extra detail either.

**Further implementation notes:**
-Everywhere possible I've left defaults as they were and disabled the additional features.

**Configs**
-My current config mods, not required in any way, but as an example..

[plugins]
cover_filenames = front.jpg,back.jpg,cd.jpg,\<artist\>.jpg
cover_resultsmax = 10
cover_searchraw = False

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/albumart.py
+++ b/quodlibet/quodlibet/ext/songsmenu/albumart.py
@@ -693,21 +693,57 @@ class AlbumArtWindow(qltk.Window, PluginConfigMixin):
         sw_list.set_shadow_type(Gtk.ShadowType.IN)
         sw_list.add(treeview)
 
-        self.search_field = Gtk.Entry()
+        search_labelraw = Gtk.Label('raw')
+        search_labelraw.set_alignment(xalign=1.0, yalign=0.5)
+        self.search_fieldraw = Gtk.Entry()
+        self.search_fieldraw.connect('activate', self.start_search)
+        self.search_fieldraw.connect('changed', self.__searchfieldchanged)
+        search_labelclean = Gtk.Label('clean')
+        search_labelclean.set_alignment(xalign=1.0, yalign=0.5)
+        self.search_fieldclean = Gtk.Label()
+        self.search_fieldclean.set_can_focus(False)
+        self.search_fieldclean.set_alignment(xalign=0.0, yalign=0.5)
+
+        self.search_radioraw = Gtk.RadioButton(group=None, label=None)
+        self.search_radioraw.connect("toggled", self.__searchtypetoggled,
+                                     "raw")
+        self.search_radioclean = Gtk.RadioButton(group=self.search_radioraw,
+                                                 label=None)
+        self.search_radioclean.connect("toggled", self.__searchtypetoggled,
+                                       "clean")
+        #note: set_active(False) appears to have no effect
+        #self.search_radioraw.set_active(
+        #    self.config_get_bool('searchraw', False))
+        if self.config_get_bool('searchraw', False):
+            self.search_radioraw.set_active(True)
+        else:
+            self.search_radioclean.set_active(True)
+
         self.search_button = Button(_("_Search"), Icons.EDIT_FIND)
         self.search_button.connect('clicked', self.start_search)
-        self.search_field.connect('activate', self.start_search)
+        search_button_box = Gtk.Alignment()
+        search_button_box.set(1, 0, 0, 0)
+        search_button_box.add(self.search_button)
+
+        search_table = Gtk.Table(rows=3, columns=4, homogeneous=False)
+        search_table.attach(search_labelraw, 0, 1, 0, 1,
+                            xoptions=Gtk.AttachOptions.FILL, xpadding=6)
+        search_table.attach(self.search_radioraw, 1, 2, 0, 1,
+                            xoptions=0, xpadding=0)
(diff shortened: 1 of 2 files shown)


elbeardmorez:
Well after struggling with a 'Gtk.init_check' issue for a while, which (miraculously) disappeared all of a sudden (I was initially blindly calling newly installed py.test with '-s' param and as it turns out 's' is not for 'script', so maybe it was that) ..the failing test ran without issue. 28 others failed but that's mainly due to my env (e.g. no pyflakes etc.)

The failing test only fails on the 'debian8-py2' bot so if anyone with that env available could shed more light that would be great.

elbeardmorez:
Further, the error in question (caused by Gtk.init_check() failure)..

> 'py.test tests/test_util_copool.py'
..
..
E       AssertionError: None != True

..occurs for me at random if I rapidly repeat the test.. sometimes pass, sometimes fail. Could the same be occurring on the build bot?

lazka:
yeah, a flaky test, ignore

lazka:
What are the additional features and why do you think they should be added?

elbeardmorez:
Apologies that this has taken me so long to come back on. Sadly, my initial reaction was that I'd given sufficient detail, specifically with commit names like:

-allow configurable custom files names for downloaded art
-move album art when renaming [#237]

but perhaps the final feature, implemented in:
-expose album art downloader's blackbox
-expose album art downloader query results limit
was a little less descriptive ..but for those there were/are commit messages too so.. ..and there was also picture?!

So I've re-written the initial message and added many more pictures ..it does read much better than the initial attempt, and now, hopefully, even for someone who hasn't necessarily seen/used the 'Album Art Downloader' it should be enough to gauge exactly what I'm doing. The problem is, that this write up has taken me about as long it did to change the code ..and personally, that isn't a balance that I find workable.

Given that you (Lazka), are effectively a deity to me (even given my limited knowledge of what you do / have achieved in the OS community) could you humour me sufficiently to elaborate on what wasn't there originally? (lol ..difficult now I've deleted it, but maybe you recall?)

Cheers

declension:
Excuse the jumping in here... I've re-read the changes and understand better now (thanks for the updates / pictures). Whilst I get reducing the number of PRs, AIUI there are _definitely_ two separate things going on:

 * Album art plugin improvements (1 and 2 in the description ). These seem pretty reasonable to me, though a combobox and a few more defaults for the cover art name (e.g. `front.jpg`) might be easier actually...
 * Change of a core UI piece to allow moving art with tag renames - in both EF and QL. (item 3 in the description). There are a few UI suggestions I have and it also needs some tests.

Can I be annoying and ask you to split the second out to a separate PR? Then the first one could be merged easier, notwithstanding any comments from @lazka

elbeardmorez:
thanks for looking. i'll split this PR as suggested, not annoying at all, i can't argue against sound reasoning.

elbeardmorez:
pretty sure that that current failure is nothing to do with me! i haven't changed anything since the last successful (passing) build, i only rebased the work over master HEAD thus i imagine it'll disappear the time around

declension:
Sorry about the delay. Thanks @elbeardmorez

Issue #2166, opened 2016-12-23
Text that is not translateable and suggestions for string changes


scootergrisen:
Bottom right corner it might say "X songs (X minutter, X sekunder)".
In about dialog "Copyright © 2004-2016 ...".
In about dialog is says "...Remote File...". Maybe that can be removed?
"Comment" in the song information dialog. Not translatable.

scootergrisen:
In the About menu it says keyboard shortcut but the dialog just says shortcut and might come from GTK. Maybe they should say the same.
The Wikipedia pluings have a string with "website" in plugins dialog.
Plugin with swiss army have strings with "Browse folders (Thunar)" and "Output playlist to stdout" and more.
Plugin with smiley have "Pattern", "Accounts" and other strings.
Settings dialog have "folder.jpg"

scootergrisen:
Should
"ReplayGain Analyzer"
Be
"Replay Gain Analyzer"
?

declension:
Thanks. Agreed on nearly all. Originally plugins weren't translatable hence a fair bit missing.

ReplayGain is a proper noun IIRC so should probably stay as is.

declension:
Yes - https://en.m.wikipedia.org/wiki/ReplayGain?wprov=sfla1

scootergrisen:
Then maybe change "Replay Gain" to "ReplayGain" in others strings.

Mellthas:
I want to confirm that (at least with the German translation) the status bar display "X songs […]" doesn’t appear translated even though it is in the PO file and translated there. Same with "X stations" in the radio browser and "different across X songs" / "missing from X songs" in the tag editor (both master and 3.8 branch).

"Comment" in the song information window doesn’t seem to be marked as translatable.

CreamyCookie:
Unless I'm missing something this was fixed with #2576. Closing.

scootergrisen:
What does #2576 fix that is mentioned here?

CreamyCookie:
From #2576: 

> Fix non-translatable strings mentioned in #2166 and also other strings.

scootergrisen:
Ok but i also suggested changes regarding "Replay Gain" and other things.

CreamyCookie:
Sorry, reopening.

Could you check which ones haven't been fixed yet?

CreamyCookie:
@scootergrisen As I'm not sure which changes are not fixed yet, please create a new issue and add the unfixed parts there. Closing this.

PR #2521, opened 2017-08-08
i18n: Fix numeric_phrase(). See #2166

krisse7:
Fix the text in bottom right corner ("X songs") so that it is translated and other cases that use numeric_phrase().


lazka:
Looks good. Thanks!

PR #2530, opened 2017-08-13
Acoustic fingerprint plugin fixes

Mellthas:
Here are several fixes to the acoustic fingerprint plugins:

1. Acoustic fingerprint submissions were missing the track title. The first commits also adds the track title to these. I don’t know if it was left out intentionally; but the web service accepts the track title as `track` (see https://acoustid.org/webservice).
2. Fingerprint lookup and submission were not working with Python 3. The second commit fixes both.
3. In the fingerprint submit dialog, opening the expander made the dialog very wide, and closing it did retain the large size. The third commit tries to smooth that behavior by making the expander affect the dialog size and making the contained label wrap.
4. The button order in the dialog was inconsistent with most other dialogs (affirmative action normally on the right side).


lazka:
Thanks

Issue #2419, opened 2017-06-08
Waveform seek plugin: add time indication on hover 


ClarkeCC:
It would be nice when I hover my mouse over the wave bar it shows in real time the track time, instead of clicking it. Like SoundCloud's wave bar does.

lazka:
![out](https://user-images.githubusercontent.com/991986/26914610-9206c7a2-4c22-11e7-85db-53a9cf38ae1e.gif)

Issue #2470, opened 2017-07-08
Add waveform seek for Soundcloud browser 


declension:
It would be nice if the Soundcloud browser displayed a waveform much like the Waveform seek bar plugin.

This could work using Soundcloud's own pre-rendered image, probably, rather than downloading the whole stream.

lazka:
+1

(assuming soundcloud survives..)

declension:
(I have a mostly-working proof of concept of this, just needs some work getting it finished and much faster)

PR #2550, opened 2017-09-01
Waveform Seekbar time indication (#2419)

Muges:
This pull request fixes issue #2419:

- It adds a soundcloud-like visual effect when the mouse hovers the waveform seekbar, and shows the time pointed by the mouse instead of the current position.
peek 01-09-2017 20-17
- It allows to seek by dragging (pressing the mouse button, moving it, then releasing it), in the same way as a Gtk.Scale.

Here is what it looks like with various popular gtk themes (from top to bottom: Adwaita-dark, Adwaita, Arc-Dark, Arc, Ambiance, Radiance):

waveformseekbar-hover

As you can see the default "hover color" (the one to the left of the mouse cursor in the screenshots above) is not always great. It is computed by blending the two other colors, and I could not find any other way that works well for all the themes I tested. I hope it's good enough, considering that the three colors used in the waveform are configurable in the plugin's options. If not, I could remove the visual effect, and only change the time labels when hovering.

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -353,37 +353,28 @@ class WaveformScale(Gtk.EventBox):
         # Avoiding object lookups is slightly faster
         data = self._rms_vals
 
-        # There can't be more than one clip rectangle, due to the draws queued
-        # But handle the other case anyway
-        rectangle_list = cr.copy_clip_rectangle_list()
-        (cx, cy, cw, ch) = rectangle_list[0]
-        if len(rectangle_list) > 1:
-            for i in range(1, len(rectangle_list), 1):
-                (ox, oy, ow, oh) = rectangle_list[i]
-                (cx, cy, cw, ch) = \
-                    (min(cx, ox), min(cy, oy), max(cw, ow), max(ch, oh))
+        # Use the clip rectangles to redraw only what is necessary
+        for (cx, cy, cw, ch) in cr.copy_clip_rectangle_list():
+            for x in range(int(floor(cx * pixel_ratio)),
+                           int(ceil((cx + cw) * pixel_ratio)), 1):
 
-        # Use that clip rectangle to redraw only what is necessary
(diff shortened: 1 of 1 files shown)


declension:
Hmm, I can't get it to work as intended here (with Ubuntu 17.04 + HiDPI). The hovering seems to un-highlight the played section when going back in time but not at all when going forwards, and can't ever go further than half way. 

It can also result in highlighting that looks like dark-light-dark-none (screenshots force a redraw which fixes this issue, though, so can't show this)

Muges:
I did not take into account the gtk scale factor for HiDPI displays. This should be fixed.

ptitjes:
I did not test it yet, but the code looks good to me.

declension:
@Muges thanks that fixes the DPI bugs.

Generally looks great to me. The only thing is the colours: I find it confusing that if you:
 * hover to the left (before) the current position, a light colour means currently played, dark means (tentative) new position.
 * hover to the right (after), dark means currently played, light means (tentative) new position

Perhaps the _whole_ bar should be lighter on dragging (yes - this means you can't see the current position I guess, but it feels less strange than colours swapping meaning IMO)

Muges:
The last commits I pushed makes the whole bar lighter as suggested, and shows the current position with a vertical line. It looks like this:
peek 06-09-2017 20-55

If you prefer I can remove the vertical line.

declension:
@Muges good work - I think it's much clearer now. Personally I think it looks nicer without the line at all, but if other people feel strongly that's fine

Muges:
> Personally I think it looks nicer without the line at all, but if other people feel strongly that's fine

Okay, I removed it. I think I liked it a bit better with a cursor to show the current position, but it's fine either way.

declension:
Thanks @Muges. Perhaps this could be an prefs option then longer term. 

Seems good to get this merged now (I'll then rebase the patch for #2470 if I get some time).

Meriipu:
So if I understand this it is intended that hovering the waveform seekbar while a track is playing will hide the current position in the track and only show where you would seek to?

I miss having the current position show even when hovering

Issue #2553, opened 2017-09-05
Convert lyrics window plugin to UI (sidebar) plugin


declension:
This plugin would probably be nicer as an sidebar, and we can do this now we have the plugin hooks.
See https://github.com/quodlibet/quodlibet/issues/273#issuecomment-80742615.

declension:
@lazka happy with that?

lazka:
sure

PR #2555, opened 2017-09-06
LyricsWindow: convert to sidebar (#2553)

declension:
* Convert to `UserInterfacePlugin` with main UI class becoming a `ScrolledWindow` not a `Window`
 * Rename to `weblyrics` for clarity (but preserve existing config) and update description.
 * Remove x,y, w, h config options (no longer relevant)
 * Remove title-setting code


declension:
Seems to work fine for me, but would be good to make sure before merging...

PR #2586, opened 2017-10-12
data_editors: Fix capitalization and tied tags not translating. Fixes #2585

### LINKED ISSUE #2585: Edit Columns, Edit Custom Commands: Incorrect title casing and tied tags not translated
When using Quod Libet in a language that doesn't use title casing, in some places the text is title cased anyway. I found at least the following places:

- File -> Preferences -> Song List tab -> Visible Columns -> Edit...
    - Here the Description column is always title cased (e.g. Original Artist in Finnish should be "Alkuperäinen artisti", not "Alkuperäinen Artisti")
    - Additionally, the description is not translated at all if the tag expression is a tied tag like `~title~version`
- File -> Plugins -> Custom Commands -> Edit Custom Commands...
    - "Max Args" and "New Command" (default name after pressing "New" button) are always title cased

I'm using the latest version from master.

krisse7:
Fix for #2585

* Edit Columns: Use the same method to format column names that is used in the actual view.
* Edit Custom Commands: Use capitalization instead of title case for parameters.

I left "New Command" title-cased as it is now. It is constructed dynamically ("New %s") and the second part ("Command") is already capitalized.

### DIFF
--- a/quodlibet/quodlibet/qltk/data_editors.py
+++ b/quodlibet/quodlibet/qltk/data_editors.py
@@ -7,7 +7,6 @@
 
 from gi.repository import Gtk
 from gi.repository import Pango
-import re
 
 from quodlibet import _
 from quodlibet import qltk, util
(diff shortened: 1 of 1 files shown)


lazka:
Thanks, looks good!

Issue #1590, opened 2015-04-21
Tag mapping versioning


lazka:
For #48, #1179 and #323 we need to add support for reading more or switching the mapping between QL tags and the tag format tags.

As a result reloading a file with new code might result in changes to tags (all performer tags change to albumartist) or making changes will delete all tags which QL didn't read before.

One solution is to version the read/write code and save the version in the DB which was used to last read and write the file.

Example code here: https://gist.github.com/lazka/588165fe1c5a8c179a66

Any thoughts?

declension:
Whilst definitely adding complexity, it sounds like a promising direction to me.

Without versioning, one or all of those features aren't going to happen without people getting very upset...

otyu:
As a user, I think this would help the interoperability of QL with other players (particularly, the TPE2 ID3 tag troubles me quite a lot). Hope this happens sometime in the foreseeable future! Cheers.

zsau:
Another option would be to create config values for any tag mapping behavior that changes, such as "map TPE2 to Album Artist". Then the default value for the setting can be changed, affecting only new libraries and not existing ones. Users with existing libraries could then flip the setting if they like, when they're ready.

zsau:
@lazka, what are your current thoughts on this? Also, possible dupe of #1215.

Issue #48, opened 2015-03-14
ID3: TXXX frames from other software are not displayed in QL


lazka:
Original [issue 48](https://code.google.com/p/quodlibet/issues/detail?id=48) created by mahikeulbody on 2008-11-10T20:20:00.000Z:

Only TXXX frames prefixed by &quot;QuodLibet::&quot; are displayed by QL into the
properties editor. TXXX such as &quot;FOOBAR::style&quot; or &quot;style&quot; are not displayed.
On the other hand, Foobar2000 and Kid3 display TXXX frames produced by QL.

Please see the attached file which is the id3 part of a mp3 file tagged
with Foobar2000 (open it with an hexa editor).

By the way, it would be nice to have an option in order QL not prefix the
user descriptor by &quot;QuodLibet::&quot;.

timpulver:
I think using a prefix for TXXX-frames is super messy and makes it hard to combine Quod Libet with other applications. I personally don’t have a »train wreck« in the public namespace, my tags are well organized. For me having `QuodLibet::MY_TAG`-tags in there would make it a train wreck and keep me from using it for writing tags. I think prefixing custom tags is a really bad decision and only because Foobar uses prefixes does not make it a good thing.

So my two cents:

- Do not write prefix on custom tags
- Write custom tags in uppercase, e.g. `CATALOGNUMBER` (this seems to be a standard)
- Use tags internally in lowercase, e.g. `<catalognumber>`

Moonbase59:
Generally, I’d be a little careful about re-casing existent tags, because there are many applications out there that actually EXPECT the casing to be correct. I know the spec says "ignore case" but applications out there in the wild sadly don’t always care. Sigh.

Just to name two from memory:

1. Older versions of SAM Broadcaster stored BPM (and other data) in a TXXX called "XFade" (exactly and ONLY this case combination).

2. MusicBrainz Picard has an option not to overwrite/change a list of tags, in which the case is respected. It would be a nightmare putting things like "replaygain_album_gain,REPLAYGAIN_ALBUM_GAIN,…" in there.

But—in any case—please DO NOT prefix TXXX frames.

sandersantema:
Has there been any progress regarding this issue?

Moonbase59:
[...]

PR #2592, opened 2017-10-15
ID3: Use TPE2 for Album Artist and a TXXX tag for Performer

### LINKED ISSUE #1179: ID3: Use TPE2 for Album Artist and a TXXX tag for Performer
Original [issue 1179](https://code.google.com/p/quodlibet/issues/detail?id=1179) created by derVedro on 2013-06-11T21:48:01.000Z:

Hi, I saw that Album Artist taged mp3 files doesn't showed correctly in Clementine and EasyTag. May be it is better to use TPE2?

muusik:
fixes #1179

### DIFF
--- a/quodlibet/quodlibet/formats/_id3.py
+++ b/quodlibet/quodlibet/formats/_id3.py
@@ -39,7 +39,7 @@ class ID3File(AudioFile):
            "TIT2": "title",
            "TIT3": "version",
            "TPE1": "artist",
-           "TPE2": "performer",
+           "TPE2": "albumartist",
            "TPE3": "conductor",
            "TPE4": "arranger",
            "TEXT": "lyricist",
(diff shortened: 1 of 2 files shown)


declension:
Hmm, no problems with the idea, but some kind of migration script to move existing libraries across might be good.

...or at least a warning to users about this change and the impact it will have on their libraries.

lazka:
Yeah, we need a strategy for migrating things... :/

zsau:
#48 needs a migration strategy as well. Tag mapping versioning (#1590) is an option, but sounds complex. I'd favor a user-activated migration script (plugin?) and a message about the change.

fdw:
Has there been any progress regarding the migration strategy? Would it be enough to move the content of the "albumartist" tag to TPE2 programmatically, or am I missing something?

declension:
Unfortunately no progress on the tag migration. 

Given this PR no longer merges I propose we close until we have a way of doing this unless anyone strongly objects...

PR #2605, opened 2017-10-28
console: Use plugin name instead of description as window title

krisse7:
Use plugin name instead of description as window title for Python Console plugin.

Currently the window title looks like this: "Interactive Python console. Opens a new window for 1 song (Quod Libet)". After the change it looks like "Python Console for 1 song (Quod Libet)".


declension:
Thanks!

Issue #2615, opened 2017-11-01
Waveform seekbar: Option to show current pos


Meriipu:
#2550 implemented a nice way that shows the position that would be seeked to in the style of the waveform bar

I miss having a way to show the current position as well, though.

PR #2616, opened 2017-11-01
Waveform plugin shows current pos (if configured to)

Meriipu:
Implements #2615 and shows the current position as a small rectangle (vertical line), which was removed from the last version of #2550 

I hope this is a clean way of doing it with configs and all...


declension:
Works well, thanks.

PR #2621, opened 2017-11-05
Update Norwegian (bokmål) translation

akrosikam:
Complete translation to Norwegian bokmål (nb) as of current source.


lazka:
Thanks

Issue #1877, opened 2016-03-05
Proposal: Second "wide" mode for paned browser


slckr:
It would be cool if the filters of the paned browser could be configured totally vertically instead of being stacked one on the other, in order to fully use the width of the screen!
See my rapid mockup attached!
![mockup](https://cloud.githubusercontent.com/assets/17674151/13548124/ae83606a-e2e8-11e5-8dca-2a66f1f3e9df.png)

Cheers,
Negativland.

lazka:
Seems reasonable, thanks.

declension:
Sounds good. If anything I'd say that's a _tall_  mode (though I prefer something more like drill-down mode, or _explorer mode_ perhaps... or maybe _columnar_) - but really all the names are confusing to me, including the current _wide mode_.

Once there are three or more options, the ideal UI might be a clickable mock-up (icon) in the preferences, almost like a "layout wizard" or something equally horrible sounding.

Negativland:
In the same spirit, the queue could be displayed vertically on the far right of the window, instead of below the song list.

**[library filters x | y | z] [song list] [queue]**

_Tall_ mode sounds good! ;)

Anyway, thanks for this fine piece of software!

declension:
> It would be awesome if there were a mode for the browser which allowed draging and dropping 'found' tracks into a playlist.

In the meantime you can open the Playlists Browser in a separate window and do exactly this.

PR #2630, opened 2017-11-15
Adds a second 'wide' mode for paned browser #1877

Tamriel:
Described in #1877.

screenshot_20171115_155354


quodlibet/quodlibet/browsers/paned/main.py
```diff
@@ -227,7 +230,11 @@ def refresh_panes(self):
             tags = [t for t in p.tags if not t.startswith("~#")]
             self.__star.update(dict.fromkeys(tags))
 
-        self.set_wide_mode(config.getboolean("browsers", "pane_wide_mode"))
+        try:
+            mode = config.gettext("browsers", "pane_mode")
+        except:
+            mode = ColumnModes.SMALL
```
frestr: You can add a default argument to gettext (the third argument), which will be returned if the value is not found

quodlibet/quodlibet/browsers/paned/prefs.py
```diff
@@ -18,12 +20,34 @@
...
+        super(ColumnModes, self).__init__(spacing=6)
+
+        group = None
+        for mode in self.modes:
+            group = Gtk.RadioButton(group=group, label=mode)
+            try:
+                if mode == config.gettext("browsers", "pane_mode"):
+                    group.set_active(True)
+            except NoOptionError:
+                pass
```
frestr: Instead of using this try-except, add a default value in the quodlibet/config.py file. This is a bit nicer, and also fixes the failing test. The old pane_wide_mode entry in config.py can also be removed.

quodlibet/quodlibet/browsers/paned/prefs.py
```diff
@@ -159,14 +183,9 @@ def __init__(self, browser):
...
 
-        wide_mode = ConfigCheckMenuItem(
-            _("_Wide Mode"), "browsers", "pane_wide_mode", True)
-        wide_mode.connect("toggled", self.__wide_mode_changed, browser)
-        menu.append(wide_mode)
-
         pref_item = MenuItem(_("_Preferences"), Icons.PREFERENCES_SYSTEM)
 
-        def preferences_cb(menu_item):
+        def preferences_cb(_):
```
frestr: Even though the parameter is not used, I think it's still a bit cleaner to make it have a descriptive name. This is nitpicking though

declension: Plus `_` is used throughout as an alias to `gettext`

frestr:
@Tamriel Would you be able to finish this, or do you want someone else do it?

Tamriel:
If it's no problem for you to finish it, I would be fine with that, and thankful :-)
Otherwise I will be able to finish it next month.

frestr:
Ok, thanks for the update. I'll wait then, as I would say it's best if you do it yourself.

Tamriel:
Ah, that merge into master is ugly.
After I fixed the tests I will do a fresh PR.

Tamriel:
> The "Equal pane width" option doesn't get applied if the column layout is changed. The apply button must consequently be clicked twice to both change layout and make the pane widths equal.
A potential workaround is to make the column layout option take effect immediately, without having to click apply.

I was not able to create this workaround. I moved my logic from the method `__apply` to the class `ColumnModes` and connected it to the click signal of each radio button. But somehow that click signal is raised two times or on creation.

frestr:
> But somehow that click signal is raised two times or on creation.

You will need to connect the radio buttons after they have all been created and one of them has been set to active.

You can see an example of that in the following snippet:

https://github.com/quodlibet/quodlibet/blob/ce92ed096fed297fa627e304ca7f83a4c71368dc/quodlibet/quodlibet/ext/events/animosd/prefs.py#L162-L176

Tamriel:
It is ready to be merged now, I think :-)

quodlibet/quodlibet/browsers/paned/prefs.py
```diff
@@ -18,13 +19,45 @@
...
 from quodlibet.util import connect_obj, escape
+from quodlibet.const import COLUMN_MODE_SMALL, COLUMN_MODE_WIDE, \
+    COLUMN_MODE_COLUMNAR
 from .util import get_headers, save_headers
 
 
-class PatternEditor(Gtk.VBox):
+class ColumnModes(Gtk.VBox):
+    modes = [COLUMN_MODE_SMALL, COLUMN_MODE_WIDE, COLUMN_MODE_COLUMNAR]
+    buttons = []
```
frestr: Move this into __init__ and rename to self.buttons, otherwise the buttons won't work the second time the pref window is opened

frestr:
Thanks :)

PR #2634, opened 2017-11-22
allow cover image pop-up to scale up to maximum size

pfps:
If the intermediate stuff I did will show up in the git log, then it might be better for someone to just  make the two tiny changes directly.


lazka:
Thanks

Issue #480, opened 2015-03-14
Sync playlist to folder/device


lazka:
Original [issue 480](https://code.google.com/p/quodlibet/issues/detail?id=480) created by cdmoore1972 on 2010-04-08T20:19:35.000Z:

I need to copy (or sync) the selected songs to a folder (or device).  The
plugin/feature would allow me to set the folder/file naming structure for
the sync-ed folder (EX:  %artist/%album/%track--%title).

In the simplest form, right click on a list of selected songs and click
&quot;sync&quot;.  From the &quot;sync&quot; dialogue, pick the destination folder and the file
naming structure.  

For reference, Songbird has the Folder Sync plugin that will do this
(interface is ugly, but you get the idea) -- 
http://addons.songbirdnest.com/addon/1535

Foobar2000 does something similar to this with their &quot;File Operations&quot;
interface (instead of sync-ing, you get copy, move, delete files instead) --
http://wiki.hydrogenaudio.org/index.php?title=Foobar2000:File_operations

Tamriel:
I wrote a plugin which syncs saved searches to a device: https://github.com/Tamriel/quod_libet_sync_to_device

Tamriel:
I think this ticket is solved by the plugin "Export Playlist to Folder".

PR #2636, opened 2017-11-24
Adds plugin 'synchronize to device' #480

Tamriel:
Described in #480.
screenshot_20171125_005019

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -65,6 +65,7 @@ quodlibet/ext/events/searchprovider.py
 quodlibet/ext/events/seekbar.py
 quodlibet/ext/events/seekpoints.py
 quodlibet/ext/events/squeezebox_sync.py
+quodlibet/ext/events/synchronize_to_device.py
 quodlibet/ext/events/synchronizedlyrics.py
 quodlibet/ext/events/telepathy_status.py
 quodlibet/ext/events/themeswitcher.py
--- a/quodlibet/ext/events/synchronize_to_device.py
+++ b/quodlibet/ext/events/synchronize_to_device.py
@@ -1,0 +1,1154 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Jan Korte
+#           2020 Daniel Petrescu
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+import os
+import shutil
+import unicodedata
+from pathlib import Path
+
+from gi.repository import Gtk, Pango
+from senf import fsn2text
+
+from quodlibet import _
+from quodlibet import app
+from quodlibet import config
+from quodlibet import get_user_dir
+from quodlibet import ngettext as ngt
+from quodlibet import qltk
+from quodlibet import util
+from quodlibet.pattern import FileFromPattern
+from quodlibet.plugins import PM
+from quodlibet.plugins import PluginConfigMixin
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.qltk import Icons
+from quodlibet.qltk.cbes import ComboBoxEntrySave
+from quodlibet.qltk.ccb import ConfigCheckButton
+from quodlibet.qltk.views import HintedTreeView
+from quodlibet.query import Query
+from quodlibet.util import print_d, print_e, print_exc
+from quodlibet.util.enum import enum
+from quodlibet.util.path import strip_win32_incompat_from_path
+from quodlibet.util.string.titlecase import human_title
+
+PLUGIN_CONFIG_SECTION = 'synchronize_to_device'
+
+
+class Entry:
+    """
+    An entry in the tree of previewed export paths.
+    """
+
+    @enum
+    class Tags(str):
+        """
+        Various tags that will be used in the output.
+        """
+        EMPTY = ''
+        PENDING_COPY = _('Pending copy')
+        PENDING_DELETE = _('Pending delete')
+        DELETE = _('delete')
+        SKIP = _('Skip')
+        SKIP_DUPLICATE = _('DUPLICATE')
+        IN_PROGRESS_SYNC = _('Synchronizing')
+        IN_PROGRESS_DELETE = _('Deleting')
+        RESULT_SUCCESS = _('Success')
+        RESULT_FAILURE = _('FAILURE')
+        RESULT_SKIP_EXISTING = _('Skipped existing file')
+
+    def __init__(self, song, export_path=None):
+        self._song = song
+        self.export_path = export_path or ''
+        self.tag = self.Tags.EMPTY
+        self._filename = None
+
+    @property
+    def filename(self):
+        if self._song is not None:
+            return fsn2text(self._song('~filename'))
+        else:
+            return self._filename
+
+    @filename.setter
+    def filename(self, name):
+        if self._song is None:
+            self._filename = name
+        else:
+            raise ValueError(_('Cannot set the filename of a song.'))
+
+
+class SyncToDevice(EventPlugin, PluginConfigMixin):
+    PLUGIN_ICON = Icons.NETWORK_TRANSMIT
+    PLUGIN_ID = PLUGIN_CONFIG_SECTION
+    PLUGIN_NAME = human_title(PLUGIN_CONFIG_SECTION.replace('_', ' '))
+    PLUGIN_DESC = _('Synchronizes all songs from the selected saved searches '
+                    'with the specified folder.')
+
+    CONFIG_SECTION = PLUGIN_CONFIG_SECTION
+    CONFIG_QUERY_PREFIX = 'query_'
+    CONFIG_PATH_KEY = '{}_{}'.format(PLUGIN_CONFIG_SECTION, 'path')
+    CONFIG_PATTERN_KEY = '{}_{}'.format(PLUGIN_CONFIG_SECTION, 'pattern')
+
+    path_query = os.path.join(get_user_dir(), 'lists', 'queries.saved')
+    path_pattern = os.path.join(get_user_dir(), 'lists', 'renamepatterns')
+
+    spacing_main = 20
+    spacing_large = 6
+    spacing_small = 3
+    summary_sep = ' ' * 2
+    summary_sep_list = ',' + summary_sep
+
+    default_export_pattern = os.path.join(
+        _('<artist>'), _('<album>'), _('<title>'))
+
+    model_cols = {'entry': (0, object),
+                  'tag': (1, str),
+                  'filename': (2, str),
+                  'export': (3, str)}
+
+    def PluginPreferences(self, parent):
+        # Check if the queries file exists
+        if not os.path.exists(self.path_query):
+            return self._no_queries_frame()
+
+        # Read saved searches from file
+        self.queries = {}
+        with open(self.path_query, 'r', encoding='utf-8') as query_file:
+            for query_string in query_file:
+                name = next(query_file).strip()
+                self.queries[name] = Query(query_string.strip())
+        if not self.queries:
+            # query_file is empty
+            return self._no_queries_frame()
+
+        main_vbox = Gtk.VBox(spacing=self.spacing_main)
+        self.main_vbox = main_vbox
+
+        # Saved search selection frame
+        saved_search_vbox = Gtk.VBox(spacing=self.spacing_large)
+        self.saved_search_vbox = saved_search_vbox
+        for query_name, query in self.queries.items():
+            query_config = self.CONFIG_QUERY_PREFIX + query_name
+            check_button = ConfigCheckButton(query_name, PM.CONFIG_SECTION,
+                                             self._config_key(query_config))
+            check_button.set_active(self.config_get_bool(query_config))
+            saved_search_vbox.pack_start(check_button, False, False, 0)
+        saved_search_scroll = self._expandable_scroll(min_h=0, max_h=300)
+        saved_search_scroll.add(saved_search_vbox)
+        frame = qltk.Frame(label=_('Synchronize the following saved searches:'),
+                           child=saved_search_scroll)
+        main_vbox.pack_start(frame, False, False, 0)
+
+        # Destination path entry field
+        destination_entry = Gtk.Entry(
+            placeholder_text=_('The absolute path to your export location'),
+            text=config.get(PM.CONFIG_SECTION, self.CONFIG_PATH_KEY, '')
+        )
+        destination_entry.connect('changed', self._destination_path_changed)
+        self.destination_entry = destination_entry
+
+        # Destination path selection button
+        destination_button = qltk.Button(label='', icon_name=Icons.FOLDER_OPEN)
+        destination_button.connect('clicked', self._select_destination_path)
+
+        # Destination path hbox
+        destination_path_hbox = Gtk.HBox(spacing=self.spacing_small)
+        destination_path_hbox.pack_start(destination_entry, True, True, 0)
+        destination_path_hbox.pack_start(destination_button, False, False, 0)
+
+        # Destination path information
+        destination_warn_label = self._label_with_icon(
+            _("All pre-existing files in the destination folder that aren't in "
+              "the saved searches will be deleted."),
+            Icons.DIALOG_WARNING)
+        destination_info_label = self._label_with_icon(
+            _('For devices mounted with MTP, export to a local destination '
+              'folder, then transfer it to your device with rsync. '
+              'Or, when syncing many files to an Android Device, use adb-sync, '
+              'which is much faster.'),
+            Icons.DIALOG_INFORMATION)
+
+        # Destination path frame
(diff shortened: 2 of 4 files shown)


quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,144 @@
...
+        "Needs the python package 'pydub'.")
+    c_path = __name__ + '_path'
+
+    def PluginPreferences(self, parent):
+        vbox = Gtk.VBox(spacing=6)
+
+        queries = {}
+        query_path = os.path.join(get_user_dir(), 'lists', 'queries.saved')
+        try:
+            with open(query_path, 'rU') as query_file:
```
lazka: open(query_path, 'r', encoding="utf-8")

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,144 @@
...
+                log = Gtk.TextView()
+                log.set_left_margin(5)
+                log.set_right_margin(5)
+                log.editable = False
+                buffer = log.get_buffer()
+                scroll = Gtk.ScrolledWindow()
+                scroll.set_size_request(-1, 100)
+                scroll.add(log)
+
+                def append(text):
```
lazka: This is racy, the buffer and log could be destroyed one idle_add calls the callback.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,144 @@
...
+                for query_string in query_file:
+                    name = next(query_file).strip()
+                    queries[name] = Query(query_string.strip())
+
+                for query_name, query in queries.items():
+                    check_button = ConfigCheckButton(query_name, "plugins", self._config_key(query_name))
+                    check_button.set_active(self.config_get_bool(query_name))
+                    vbox.pack_start(check_button, False, True, 0)
+
+                def synchronize():
```
lazka: All the quodlibet API isn't thread safe (except if explicitly stated), so threads wont work here.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,144 @@
...
+                def stop(_):
+                    append("Stopping...")
+                    self.running = False
+
+                def path_changed(entry):
+                    self.path = entry.get_text()
+                    config.set('plugins', self.c_path, self.path)
+
+                try:
+                    self.pattern = config.get('plugins', self.c_path)
```
lazka: get() can be passed a default value

quodlibet/docs/packaging.rst
```diff
@@ -117,6 +117,8 @@ plugin from loading.
 **libappindicator-gtk3** + **typelibs**:
     * For the Tray Icon plugin under Ubuntu Unity and KDE Plasma
 
+**pydub**:
```
lazka: (It's a bit unfortunate that pydub is needed, since theoretically we already ship everything for transcoding with GStreamer, but we can always change it at some later point)

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,144 @@
...
+
+                vbox.pack_start(destination_path_box, True, True, 0)
+                vbox.pack_start(start_button, True, True, 0)
+                vbox.pack_start(stop_button, True, True, 0)
+                vbox.pack_start(Gtk.Label(label=_("Output:")), True, True, 0)
+                vbox.pack_start(scroll, True, True, 0)
+                return qltk.Frame(_("The following saved searches shall be synchronized:"), child=vbox)
+        except FileNotFoundError:
+            return qltk.Frame(_("No saved searches yet, create some and come back!"))
+        except Exception as e:
```
lazka: Let uncaught exception through, so we get a crash report and can fix them.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
Tamriel: Thanks for the excellent code review!
Without using threads the UI will become unresponsive until the synchronisation has finished.
What could happen due to the not-thread-safe API? I think the worst scenario are wrong songs on the device.
The user is aware of the ongoing sync and won't edit the library.
So I propose to keep the threads.

Tamriel: So what shall we do here?

Tamriel:
> All the quodlibet API isn't thread safe (except if explicitly stated), so threads wont work here.

But without threads, we can't show a process bar and the UI will become unresponsive, so we should keep threads.
We have two options now:

a) Use a modal dialog and show the sync process there. The dialog blocks user interaction with the library.

b) Allow using and editing the library while syncing, because the sync may take an hour and the user still wants to control what music he listens to.
 If the user adds songs to the library while syncing, then he won't have the new songs on his device. I think this is natural behaviour, the user is aware of this and will just re-sync after editing the library.

As soon as we have a decision I will polish this PR.

@frestr What do you think?

frestr:
> But without threads, we can't show a process bar and the UI will become unresponsive, so we should keep threads.

I haven't tested or looked thoroughly through this PR yet, but a common way to do this in gtk applications is to have the following snippet run regularly:

```python
while Gtk.events_pending():
    Gtk.main_iteration()
```

This ensures gtk can process user events and prevents the application from becoming unreponsive. It might be possible to have someting like that between each song synced?

> We have two options now:

The second option makes the most sense to me, given that the above suggestion works.

frestr:
You might also want to take a look at how the "Export Playlist to Folder" plugin is implemented. It uses the copool to do copying and regularly yields to let the UI update. It makes the UI pretty slow, but it's still usable to some extent.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
Tamriel: How could the buffer and log be destroyed?

Tamriel:
I have some problems with your language tests. How can I fix them?

Tamriel:
The test complains my plugin name 'Synchronize To Device' should be in title case like  'Last.fm Cover Source', 'Last.fm Sync', 'Send to iFP', 'This is a test'.
I think it is, so the test may be wrong?

frestr:
That test is perhaps a bit too strict (with not very helpful error messages). The reason it fails here is that you have written "To" with a capital T. "Synchronize to Device" works.

Tamriel:
Thanks for your help.
The only thing left is the maybe racy stuff [above](https://github.com/quodlibet/quodlibet/pull/2636#discussion-diff-159198577R40).
Is that important?

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,168 @@
...
+
+                            # prevent the application from becoming unreponsive
+                            while Gtk.events_pending():
+                                Gtk.main_iteration()
+
+                            song_path = song['~filename']
+                            song_file_name = "{} - {} - {} - {}.mp3".format(
+                                song("genre"), song("artist"),
+                                song("album"), song("title"))
+                            valid_chars = "-_.() %s%s" % (
```
frestr: Why is this here? Unicode characters should be allowed. This also generates duplicate filenames if the only difference in the filename is unicode characters.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
frestr: Do you need to use idle_add here?

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,168 @@
...
+                        append("Synchronization finished.")
+                    except FileNotFoundError as e:
+                        append(str(e))
+
+                def start(_):
+                    append("Starting...")
+                    self.running = True
+                    synchronize()
+
+                def stop(_):
```
frestr: Use `button` or something similar as the first argument instead, as this overrides `_` which we use for translation.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,168 @@
...
+                vbox.pack_start(start_button, True, True, 0)
+                vbox.pack_start(stop_button, True, True, 0)
+                vbox.pack_start(Gtk.Label(label=_("Progress:")), True, True, 0)
+                vbox.pack_start(scroll, True, True, 0)
+                return qltk.Frame(
+                    _("The following saved searches shall be synchronized:"),
+                    child=vbox)
+        except FileNotFoundError:
+            return qltk.Frame(
+                _("No saved searches yet, create some and come back!"))
```
frestr: Also check if the file is empty.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,168 @@
...
+                            enabled_queries.append(query)
+
+                    selected_songs = []
+                    for song in app.library.itervalues():
+                        if any(query.search(song) for query in
+                               enabled_queries):
+                            selected_songs.append(song)
+
+                    filename_list = []
+                    destination_path = config.get('plugins', self.c_path)
```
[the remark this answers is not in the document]
frestr: By default value I mean any value, such that no exception is thrown. The empty string `''` would be good here.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,168 @@
...
+                    self.running = False
+
+                def path_changed(entry):
+                    self.path = entry.get_text()
+                    config.set('plugins', self.c_path, self.path)
+
+                self.pattern = config.get('plugins', self.c_path,
+                                          'the path of your device, e.g. '
+                                          '/run/media/my-username/device-id/'
+                                          'Music')
```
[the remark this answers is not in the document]
frestr: You can set a tooltip on the entry with `destination_path.set_tooltip_text()`.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
frestr: That definitely exists if you think about non-latin scripts like the ones used in Japanese, Chinese etc. Ideally there would be some fallback or option for this, but it should be okay for now.

frestr:
> Is something wrong with the tests?

Yeah, you need to rebase to include 2926f2422924f0ebd30cd52c87e859df6b2680df.

> Regarding idle_add: It's not a method of the scolled object

What do you mean by this?

Tamriel:
Is the test config broken?

frestr:
> One test fails.
Is the test config of travis broken?

It just timed out. Fixed by a rerun.

frestr:
So it looks alright now, but is there a reason you can't have append like this:

```python
def append(text):
    buffer.insert(buffer.get_end_iter(), text + "\n")
    log.scroll_to_mark(buffer.get_insert(), 0.0,
                       True, 0.5, 0.5)
```

without the idle_add?

Tamriel:
> So it looks alright now, but is there a reason you can't have append like this
It works your way, so I added it. I guess the test timed out again?

frestr:
> I guess the test timed out again?

You have syntax errors in your file.

Tamriel:
I have no idea why that test fails.

Tamriel:
It is ready to be merged :-)

declension:
@lazka are you happy to merge this?

dpet23:
I was thinking of writing a similar plugin when I came across this pull request, but I'm wondering why this plugin syncs only saved searches and not playlists? (I currently have several large playlists that I'd like to sync, made up of manually-curated songs with many different tags; these playlists would be very hard to replicate using saved searches.)

gevhaz:
@daniel, there's an easy solution! One of the great things with Quod Libet is the advanced search functionality and you can do e.g. `~playlists=<playlist name>`.

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,207 @@
...
+                                        "exists.".format(dest_file))
+                                else:
+                                    append(
+                                        "Writing '{}'...".format(dest_file))
+                                    copyfile(song_path, dest_file)
+
+                            # delete files which are not
+                            # in the saved searches anymore
+                            for existing_file in Path(destination) \
+                                    .rglob('*.' + file_ext):
```
dpet23, line 130: If `selected_songs` is empty, `file_ext` will not be defined at this point:

```

  File "quodlibet/quodlibet/ext/events/synchronize_to_device.py", line 168, in start
    synchronize()

  File "quodlibet/quodlibet/ext/events/synchronize_to_device.py", line 134, in synchronize
    .rglob('*.' + file_ext):

UnboundLocalError: local variable 'file_ext' referenced before assignment
```

quodlibet/quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -0,0 +1,207 @@
...
+    config_path_key = __name__ + '_path'
+
+    def PluginPreferences(self, parent):
+        vbox = Gtk.VBox(spacing=6)
+
+        query_path = os.path.join(get_user_dir(), 'lists', 'queries.saved')
+        try:
+            with open(query_path, 'r', encoding="utf-8") as query_file:
+                if not query_file.read(1):
+                    raise FileNotFoundError
```
dpet23, line 42: This check prevents reading the first character of the `query_file`, and the query at the top of the file will be parsed incorrectly.

Example: with the current code, a query of `title=Sun` at the top of the file will be parsed as `itle=Sun`.

To show the "No saved searches yet" message when the file is empty, I suggest the following:

1. Remove this `if` statement:
    ```suggestion
    ```

1. _(Optional)_ Create a custom exception class to clarify why we're raising an exception:

    ```python
    class NoSavedQueriesError(Exception):
        pass
    ```

1. Add an `else` clause after reading the `query_file`:

    ```python
    queries = {}
    for query_string in query_file:
        name = next(query_file).strip()
        queries[name] = Query(query_string.strip())
    else:
        # query_file is empty
        raise NoSavedQueriesError
    ```

    or add a new `if` statement after the `for` loop:

    ```python
    if not queries:
        raise NoSavedQueriesError
    ```

1. _(Optional)_ If using the custom exception class, update the `except` clause at the bottom of the file:

    ```python
    except NoSavedQueriesError:
    ```

Tamriel:
I fixed dpet23s bug and pulled from master to be able to merge.
Somehow the test `tests.test_util_copool.Tcopool testMethod=test_pause_restart_pause` fails. It does not fail on my machine and did not fail without the upstream changes.
Any ideas?

Tamriel:
I think it is ready to be merged :-)

Meriipu:
This is not really a review, but maybe consider rebasing/squashing some of the commits to make the git log cleaner and more useful (93 commits spread over time is a lot to deal with when viewing the log).

dpet23:
> consider rebasing/squashing some of the commits to make the git log cleaner and more useful

Is it acceptable to force push after changing history? It's usually discouraged.

Also, I'd recommend reviewing the overall changes instead of commit-by-commit, since the code has been refactored several times.

Meriipu:
> > consider rebasing/squashing some of the commits to make the git log cleaner and more useful
> 
> Is it acceptable to force push after changing history? It's usually discouraged.
>
It should not matter as long as it is an unmerged branch that is having its history edited (before being merged), since nobody is building on top of those commits yet. 
It is not really for the benefit of the reviewer as much as it is for the benefit of anybody reading the commit log in the future after the changes have been merged. 

> Also, I'd recommend reviewing the overall changes instead of commit-by-commit, since the code has been refactored several times.
>
Sure, the overall changes are easy to see when they are presented in a pull request in a combined manner.
But after they are merged, all the commits are loose again. If somebody were to view the log of quodlibet:master it would be full of these `fix merge error`-commits that do not really change anything (and there seems to be a lot of them here)

I see the value in preserving authorship in case a pull-request has several authors, and it should be possible to preserve authorship when squashing/combining commits from the same author. 
So I do not think squashing it all into a single commit is appropriate, but at least squashing the related commits together might be beneficial for future readability. 

These are just my personal opinions on it though, and I think quodlibet is/has been fairly lax on tidiness in the git log (at least for smaller pull requests with <5 commits). 
93 where a majority of them are very small changes like formatting are a bit on the excessive side in my opinion though.

declension:
@Tamriel @Meriipu cleaning up old PRs. Is this one rescuable? If so let's try to fix the branch and any urgent review comments and get it merged (I can help perhaps)

If not let's close

dpet23:
@declension @Meriipu, I'll squash some of the commits to make the log cleaner and (force-)push my changes in the next few days.

I think the review comments above have been addressed, but it would be useful for this PR to be reviewed again, since I've changed most of the initial implementation.

Otherwise, I think this is ready to be merged.

Tamriel:
I squashed the commits into one. I would be happy for a final review and a merge then :-)

declension:
Thanks!

Issue #2641, opened 2017-11-29
Could not add RSS podcast feed


rohshall:
What did you try to do?
Added the following podcast: 
http://cbsradionewsfeed.com/rss.php?id=126&ud=12

* What did you expect to happen?
podcast being added

* What did happen instead?
I get an "Unable to add feed" error dialog

* Which version of Quod Libet?
master branch on 11/28

* Which operating system including version (Ubuntu 14.04, Win7, Debian sid, ...)?
Python: 2.7.14 (default, Nov  2 2017, 19:02:24) 
[GCC 7.2.1 20170829 (Red Hat 7.2.1-1)]
Platform: Linux-4.15.0-0.rc0.git7.1.fc28.x86_64-x86_64-with-fedora-28-Rawhide

lamacchiacosta:
I've got the same, trying to add the following feed:
http://rss.earwolf.com/yo-is-this-racist

declension:
This seems to now work, probably the upgrade to feedparser 6.0 in d218a9daea181394eddbbd3e02b17d1182260f43

PR #2642, opened 2017-11-29
#2641 - fix for being unable to add feeds whose content-type is HTMLish

rohshall:
Python's feedparser handles feeds that are HTMLish, that is, whose content type is text/html.
I have tested that after accepting these feeds, the podcast plays just like other feeds.

### DIFF
--- a/quodlibet/quodlibet/browsers/audiofeeds.py
+++ b/quodlibet/quodlibet/browsers/audiofeeds.py
@@ -50,6 +50,7 @@ class InvalidFeed(ValueError):
 
 
 class Feed(list):
+
     def __init__(self, uri):
         self.name = _("Unknown")
         self.uri = uri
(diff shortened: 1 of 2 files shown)


declension:
Thanks.

Could you add a test for this (with some fake or real data)? It should be relatively easy now, see my recent addition for a pattern.

declension:
BTW @rohshall I recommend not using live URLs in the tests - too much maintenance and we can't rely on networking always. Suggest copying (a portion of) the data to a local file in `tests/data` and using that, e.g. [this test data](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/tests/data/valid_feed.xml).

rohshall:
If it was about just using test-data, I would have created mock data. But,
it needs the mock HTTP response, with Content-Type header as text/html. I
don't know enough about Feedparser to write that test. Any pointers?

declension:
@rohshall Sorry I started looking at that but run out of time. I see what you mean. I'll have a look...

declension:
This code is now gone, but underlying problem fixed by d218a9daea181394eddbbd3e02b17d1182260f43

PR #2656, opened 2017-12-14
add explanation of double-clicking on main song list to overview; fix problems in cli enqueue and enqueue-files

pfps:
The overview didn't have a description of what double-clicking did in the main song list except to compare it to what happens in other song lists.

### DIFF
--- a/quodlibet/quodlibet/qltk/edittags.py
+++ b/quodlibet/quodlibet/qltk/edittags.py
@@ -819,7 +819,7 @@ class EditTags(Gtk.VBox):
 
     def __edit_tag(self, renderer, path, new_value, model):
         new_value = gdecode(new_value)
-        new_value = ', '.join(new_value.splitlines())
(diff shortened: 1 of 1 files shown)


pfps:
There were two problems.  enqueue was passing relative file names.  enqueue-files wasn't unpacking its argument (and thus it was also passing relative file names)

pfps:
also format code to meet guidelines

declension:
LGTM generally. Would you mind adding some tests around this? This has been [started in `test_comands.py`](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/tests/test_commands.py#L88) actually...

pfps:
I'll give it a try, but I'm not sure how to proceed.  The real test is to have Quodlibet running in one directory and enqueue one or more files using the CLI running from a different directory.  Is there a test harness to test the remote CLI interface?

pfps:
Closed to allow for fresh synchronization in my fork.

Issue #1724, opened 2015-10-25
Setting volume command-line anomaly


ruuter:
**What did you try to do?**
Set exact volume from command-line with `quodlibet --volume=10`

**What did you expect to happen?**
Volume to be set at 10%

**What did happen instead?**
46%. Volume 1=22, 2=27, 3=31, 4=34, ... 80=93, .. 85=95 ...

**Which version?**
Quod Libet 3.5.1, Ubuntu 14.04.3 LTS

I can't be sure, but I think it came with latest update.
I can be sure though it worked as expected few months ago.

lazka:
The volume widget in QL now uses the pulseaudio display scale for volume (linear ^ (1/3) = PA volume, 0.1 ^ (1/3) = 0.46415) to match the display widgets in pavucontrol and the gnome sound settings. The command line interface still exposes the linear volume scale.

ruuter:
That explains it.

jplatte:
Okay, so the command line interface still exposes the linear volume scale... But why? The volume widget in the interface feels more intuitive when it comes to actual noise level, and it is just inconsistent like this.

Could this issue be reopened / would you take a PR changing the command line interface (`--volume`, `--volume-up` and `--volume-down`) to match the GUI?

jplatte:
Seriously, I can't even work around this by converting the numbers myself because the volume flag only accepts integers.

@lazka Would you accept a PR that changes the command line to operate in PA display scale and/or a PR that makes the volume flag take a float argument?

lazka:
float sounds fine.

lazka:
``quodlibet --volume=$(python3 -c "import sys; print(int(float(sys.argv[1])**3*100), end='')" 0.5)``

jplatte:
If somebody else was having issues with this, here is my updated i3blocks script that converts uses the same display scale as PulseAudio and allows increasing / decreasing the volume by 5% in that scale by scrolling: https://github.com/jplatte/i3blocks-scripts/blob/31511b8a5da678eca94401344be483083bc83bbc/quodlibet-volume

PR #2661, opened 2017-12-21
cli, commands: Allow floating point arguments for volume

jplatte:


lazka:
Thanks!

lazka:
For context: #1724

PR #2667, opened 2017-12-26
show confirmation dialog when removing songs

### LINKED ISSUE #2248: Add option to confirm removals from playlists
I miss an option to turn on a confirmation for removing tracks from playlists, or confirmations for any removals even.

zsau:
Fixes #2248. Not sure I did the testing right.

### DIFF
--- a/quodlibet/quodlibet/qltk/songsmenu.py
+++ b/quodlibet/quodlibet/qltk/songsmenu.py
@@ -31,11 +31,16 @@ from quodlibet.util.songwrapper import ListWrapper, check_wrapper_changed
 
 def confirm_song_removal_invoke(parent, songs):
     songs = set(songs)
-    if len(songs) == 1:
-        title = _("Remove track: \"%s\" from library?"
-                  % next(iter(songs))['title'])
-    else:
-        title = _("Remove %s tracks from library?" % len(songs))
+    if not songs:
+        return True
+
+    count = len(songs)
+    song = next(iter(songs))
+    title = ngettext("Remove track: \"%(title)s\" from library?",
+                     "Remove %(count)d tracks from library?",
+                     count) % {'title': song('title'),
+                               'count': count}
+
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/qltk/songsmenu.py
```diff
@@ -29,6 +29,17 @@
 from quodlibet.util.songwrapper import ListWrapper, check_wrapper_changed
 
 
+def confirm_song_removal_invoke(parent, songs):
+    songs = set(songs)
+    if len(songs) == 1:
+        title = _("Remove track: \"%s\" from library?"
+                  % next(iter(songs))['title'])
+    else:
+        title = _("Remove %s tracks from library?" % len(songs))
```
declension: This needs to use the `ngettext` (plural) form - I know you check for n=1, but languages behave quite differently with their pluralisation (e.g. some have several plural forms, and n=0 varies (though I guess this should never happen)).  

Also that should really be a `%d`...

quodlibet/quodlibet/qltk/songsmenu.py
```diff
@@ -29,6 +29,17 @@
 from quodlibet.util.songwrapper import ListWrapper, check_wrapper_changed
 
 
+def confirm_song_removal_invoke(parent, songs):
+    songs = set(songs)
+    if len(songs) == 1:
+        title = _("Remove track: \"%s\" from library?"
+                  % next(iter(songs))['title'])
```
declension: Perhaps we should consider the possible-though-rare case where tracks don't have titles. We could default to the `~filename` in this case, perhaps, rather than display `""`

zsau:
I should also credit @Meriipu for this, since it's based on their gist from #2667 (https://gist.github.com/Meriipu/f59d3a0871b528d58fc9f2ad4e19bec3).

declension:
Thanks!

PR #2674, opened 2017-12-30
use "performance" role for bare performer tag

zsau:
Patterns like `people:roles` currently generate non-instrument role descriptions like Arrangement, Lyrics, etc. This patch gives bare (role-less) performer tags a similar "Performance" role description (e.g. `<people:roles>` would output `Foo (Performance)` for a song with the tag performer=Foo).

### DIFF
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -510,10 +510,10 @@ class AudioFile(dict, ImageContainer):
 
         role_map = {}
         for key in role_tag_keys:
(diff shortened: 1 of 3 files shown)


declension:
Thanks. LGTM. 

In general, please open an issue first (then link the PR to that), makes tracking and discussion of features / fixes easier, and also updating changelogs etc simpler too.

declension:
Thanks!

davidSchlosser:
I have many performer tags that intentionally don't include a role, as the meaning of a blank role is clear to me. Could this become configurable?

PR #2675, opened 2017-12-30
Prompt for playlist song removals

Meriipu:
As #2667 only accounts for removing from library, this attempt to implement a similar prompt for removing tracks from playlists. #2248

I really dislike the way I ended up passing a flag around in the removal functions/event handlers to account for the tests that use the same interface (and otherwise get stuck waiting for the prompt?)

It could maybe be done differently, though I can not think of a way myself.
playlists/collections seem to be fairly complex...

### DIFF
--- a/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/browsers/playlists/main.py
@@ -39,7 +39,9 @@ from quodlibet.util import connect_obj
 from quodlibet.util.collection import Playlist
 from quodlibet.util.dprint import print_d, print_w
 from quodlibet.util.urllib import urlopen
-from .util import parse_m3u, parse_pls, confirm_remove_playlist_dialog_invoke, _name_for
+from .util import parse_m3u, parse_pls, _name_for, \
+    confirm_remove_playlist_dialog_invoke, \
+    confirm_remove_playlist_tracks_dialog_invoke
 
 DND_QL, DND_URI_LIST, DND_MOZ_URL = range(3)
 
@@ -352,6 +354,15 @@ class PlaylistsBrowser(Browser, DisplayPatternMixin):
             if not removals:
                 print_w("No songs selected to remove")
                 return
+
+            parent = self
+            songset = {removals[key] for key in removals}
+            response = confirm_remove_playlist_tracks_dialog_invoke(
+                parent, songset, self.Confirmer)
+            if not response:
+                print_d("Removal of track(s) from playlist stopped via prompt")
+                return
+
             if self._query is None or not self.get_filter_text():
                 # Calling playlist.remove_songs(songs) won't remove the
                 # right ones if there are duplicates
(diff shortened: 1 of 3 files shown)


lazka:
Would monkey patching `Gtk.Dialog.run()` for the duration of the test work?

Something like

```
def new_run(self):
     self.response(response_id)
     return response_id
```

Meriipu:
it unfortunately (still) does not test that the responses to the prompt correspond to True/False, though I added a test to simulate a "decline"

it was 4 extra lines so I included it in the existing test rather than split it into two tests and duplicate all the helpers + setup

Are there better ways to temporarily patch the prompt creation/running code that works with a much simpler change similar to module.function = new_function (the latter did not work out for me)?

Meriipu:
I can squash it into a single commit if that is cleaner

Meriipu:
I think this is a cleaner way to monkey-patch it. Prompt appears and works as expected when I test manually, as well as when given a fake true/false helper-function. 

This method where run is modified directly in the parent class (ConfirmationPrompt) as opposed to in each prompt that inherits it also makes it cleaner to add tests for all the other confirmation prompts too (in fact, it does not look like any of the other ConfirmationPrompt-using prompts have tests; this is the first one from what I can tell).

Meriipu:
I recall the comments from the on #2675 now. I will look into that.

Meriipu:
Would this benefit from having an option to disable the prompt? Removing songs from playlists sounds like it might be a more common operation, and some users might dislike getting prompted.

Meriipu:
I updated the pull-request to use the most recent confirmation prompt (sharing the mock-confirmer with the playlist-removal prompt).
Although there is a test-error with the ngettext here or something.

Lastly maybe it might be beneficial to control prompt-settings via a tab in preferences (out of all the prompts I can see some people being annoyed by this one for removal from playlists).
The motivation is that the delete-key is a shortcut for this and accidentally hitting it would lead to tracks being removed from the playlist.

Meriipu:
ngettext-issue probably fixed now.

declension:
Can this be fixed @Meriipu ?

Meriipu:
Most likely yes.

but I vaguely recall there was a discussion about whether some people might find too many of these prompts annoying.

and whether to have a tab in the preferences to handle whether to show various prompts (but without adding any notes or toggles for this to the actual prompts to avoid overloading the user with information, just mostly for users who care enough to look for an option to turn them off).

Meriipu:
should be fixed again now

declension:
I've tried it, works fine :+1: . 
Personally I find it a little annoying, as I'm so used to deleting things without pop-ups. Maybe configurable would be good, I dunno.

Also note that once [global undo](#4063) is done these kind of things shouldn't actually be necessary, but then that could be a while and probably only after a release.

Meriipu:
I was thinking maybe a tab in preferences called prompts

where users could toggle on or off specific prompts (only there though, no complicated option to "don't ask again" in the actual prompts).

Just a list of prompt description on the left and a checkbox on the right

declension:
Closing for now in favour of Global undo (#4063)  (see first attempt #4058)

PR #2684, opened 2018-01-02
don't use inline separators when changing multiple tag values

pfps:
Leaving in the newline works much better.


pfps:
I was trying to write tests for this but the functionality is buried in a class and I don't know how to access it in a test.

declension:
Thanks. Sorry about the delay

PR #2690, opened 2018-01-03
Improve waveform seekbar hoover time indication

Eyenseo:
Before this patch only a tiny line was drawn to indicate the current position.
With this patch the hover color is used to display the current playing
position if the mouse hovers before this position and the elapsed color is
used to display the new position, if the mouse hovers after the position the
elapsed color is used to display the current playing position and the hover
color is used to display the new position. This might sound confusing but
works intuitively very well.

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -3,6 +3,7 @@
 #        2016-17 Nick Boultbee
 #           2017 Didier Villevalois
 #           2017 Muges
+#           2017 Eyenseo
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


Meriipu:
I like it though the only thing that feels slightly off is the change of colours when the mouse is before the current position

lazka:
Looks good!

Any thoughts if we can remove the "Show current position" option and default to true? It seems like there is no good reason to disable it.

Meriipu:
I just defaulted it to False since the commit that changed the hover to transparent bailed out on including it iirc.

lazka:
(ah, there is the discussion https://github.com/quodlibet/quodlibet/pull/2550#issuecomment-327925110)

@declension can you test if you like this more than the single line, and if it can be always on maybe?

declension:
Tried it. I have to say I find the colours changing suddenly quite confusing (like @Meriipu). My brain doesn't understand what "dark" and "light" mean as you move over the currently playing position

I've found though, if you just _avoid_ doing that it does look nicer. @lazka I'm fine either way then I guess...

lazka:
Might be that I'm used to it because soundcloud draws it the same way.

Let's get this in first.

lazka:
Thanks @Eyenseo

PR #2696, opened 2018-01-05
Improve WaveformSeekbar usability

Eyenseo:
In continuation of #2690 this removes the discussed option, introduces an additional color (in an attempt to improve the usability) and uses the hover concept also for seeking.

The hover color (now preview) is blended with the elapsed one. This color is then used in front of the mouse position if the covered wave is elapsed. If no preview color is specified the elapsed one is used instead to calculate a shade there of. The preview color is used for the time that remains. The result is that while hovering the user moves a 'light tint' over the waveform and while seeking a 'solid color'.

I'm not sure if it is ok to just rename a config option or if there has tho be a migration of old config values?

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -198,7 +198,7 @@ class WaveformSeekBar(Gtk.Box):
         ! audioconvert
         ! level name=audiolevel interval={} post-messages=true
         ! fakesink sync=false"""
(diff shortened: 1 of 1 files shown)


lazka:
With the default Adwaita theme, the current position is (mostly) no longer visible when hovering a later position. I personally prefer how it's in master right now compared to this.

declension:
Doesn't look like this can / will get merged, so closing for now. Thanks.

Meriipu:
I use this version (up to 8353275 at least) in my build because I prefer how it looks here.

Issue #2487, opened 2017-07-18
UI: Option to have solid scrollbar


Meriipu:
Are solid scrollbars (with reserved space) even possible in GTK? It is a bit tricky for me to see the current scrollbars always, and they tend to just barely but in a somewhat distracting way cover up the bottom element in lists whenever I move my cursor.

lazka:
We default to it on macOS: https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/_init.py#L251

MartinBorg:
On 2017-07-18 13:23, Christoph Reiter wrote:
> We default to it on macOS: https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/_init.py#L251

Works great on Linux, too:

$ GTK_OVERLAY_SCROLLING=0
$ export GTK_OVERLAY_SCROLLING
$ quodlibet

lazka:
> is it a sensible thing to have an option for?

If it affects all gtk+ applications I think it would be better if you override the setting globally somehow as @MartinBorg suggested.

But if you care enough to provide a patch for adding it to the advanced settings plugin, that would be fine with me.

PR #2697, opened 2018-01-06
added advanded_preferences config for scrollbar visibility

Meriipu:
issue #2487

No change to default state on Linux (and I assume on windows too)
No idea what the behaviour is on osx, though there should be no change to default behaviour there either as far as I can tell


lazka:
Somethings not right with the diff

Meriipu:
wow I might have pulled stuff the wrong way at a point here

PR #2701, opened 2018-01-08
searchbar: Add shortcut Ctrl+Return to enqueue all in songlist

nemethf:
This patch makes more convenient to enqueue one or a few songs without using the mouse.  First, I filter the `song list` from the `search bar`, then without navigating to the `song list` I can press Ctrl+Return and the songs will be in queue.

I guess the new shortcut key should be documented, but I couldn't find where.

Thanks.

### DIFF
--- a/quodlibet/quodlibet/qltk/searchbar.py
+++ b/quodlibet/quodlibet/qltk/searchbar.py
@@ -16,6 +16,7 @@ from quodlibet import config
 from quodlibet import _
 
 from quodlibet.query import Query
+from quodlibet.qltk import is_accel
 from quodlibet.qltk.cbes import ComboBoxEntrySave
 from quodlibet.qltk.ccb import ConfigCheckMenuItem
 from quodlibet.qltk.x import SeparatorMenuItem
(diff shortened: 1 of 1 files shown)


lazka:
One problem is that in all browsers which have additional filters after the search bar, entering text does not changer the song list, so that would probably be confusing. But maybe that can be ignored for now.

nemethf:
I haven't noticed that, because I only use the paned browser.  An easy solution would be to limit this new functionality to the paned browser by adding the shortcut key only to SearchBarBox instance[ somewhere here ](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/browsers/paned/main.py#L83).

I don't really know what the general solution could be.  The Cover Grid does not even bind the Ctrl-Return combo.  A Ctrl-Return in the search bar should enqueue all the  visible items (songs/albums) in the associated pane(s).   One possibility would be to add a "virtual"  enque_all() method the the Browser class, which each derived class could overwrite; and a connect_searchbox(), which would take care of binding the shortcut key.

I prefer the first solution (because it's less work).  How should I proceed?  Thanks.

frestr:
I'll close this, as #2745 was merged.

PR #2729, opened 2018-01-17
waveformseekbar: Clamp seek time to valid range

psimonyi:
Mouse move events can be delivered when the mouse is outside the widget during a drag, so treat x-positions beyond the end of the seek bar as just at the start or end.

Without this, dragging from the waveform seek bar outside its range will show seek times that are negative or past the end of the track (though it is properly clamped to the valid range once you release the mouse button).


lazka:
Thanks

PR #2745, opened 2018-01-26
paned browser: Add shortcut Ctrl-Return to the searchbar

nemethf:
Ctrl-Return in the searchbar enqueues the complete songlist

Sorry for being impatient: I couldn't wait for your response, so this is an alternative implementation of  #2701.  Similarly to the previous pull request this patch makes more convenient to enqueue one or a few songs without using the mouse. First, I filter the song list from the search bar, then without navigating to the song list I can press Ctrl+Return and the songs will be in queue.

Thanks.

### DIFF
--- a/quodlibet/quodlibet/browsers/paned/main.py
+++ b/quodlibet/quodlibet/browsers/paned/main.py
@@ -12,12 +12,14 @@
 
 from gi.repository import Gtk, GLib
 
+from quodlibet import app
 from quodlibet import config
 from quodlibet import qltk
 from quodlibet import util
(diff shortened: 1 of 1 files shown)


declension:
I've got no real objections to this (but don't use that browser much). @lazka ?

frestr:
I think this seems quite useful.

It should also be documented in quodlibet/quodlibet/docs/guide/shortcuts.rst.

lazka:
I would prefer something which works in all browsers, but this is a start I guess.

Issue #2453, opened 2017-06-26
Discussion about confirmation prompts


Meriipu:
So there are quite a few confirmation prompts already, and also issues for a few new ones.
existing. Among the ones I know of/found:
* remove duplicates plugin
* change ratings of multiple songs (or remove)
* Delete playlist
* move to trash
* a few base classes I suppose in qltk/msg.py

some new/proposed ones:
* clear queue #2349
* confirm removal from playlists (and library, I suppose) #2248
* drag songs to playlist to add #2367

I am not sure about the ones in qltk/msg.py, but apart from move to trash (which has some detailed info in its message box), most of these look the same save for button icons/labels and text. 

Would it make sense to add a new tab to a preferences menu somewhere about which warnings to suppress, so that new dialogs could be added without making them mandatory all the time for all users (as well as adding the option of suppressing future dialogs of existing ones, though maybe that could be harmful)? 
I have no idea how it would be implemented, but gathering it in one place sounds the tidiest and most manageable. Doing configurations cleanly seems a bit tricky. A comment in config.py called a section dumping grounds, even. So yeah.

declension:
> would it make sense to add a new tab to a preferences menu somewhere about which warnings to suppress, so that new dialogs could be added without making them mandatory all the time for all users 

Sounds good, if done carefully. My gut feeling is there will be subtleties around this and the different levels of safety needed (e.g. really add to playlist vs really delete from disk...)

> Doing configurations cleanly seems a bit tricky.

There are definitely at least three ways of doing this kind of stuff (especially for plugins), see #1827. But at least there are plenty of examples in the code (and for tests) 

> A comment in config.py called a section dumping grounds, even. So yeah.

That comment is from 2005.... 

Newer config we've added is generally better organised / namespaced / commented now. The hard thing with _changing_ configuration here (and thus why that comment is still around) is that you have to support users upgrading from one QL version to the next. @lazka added some good [migration support](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/util/config.py#L103) a while back but IIRC we haven't used it that much (related: it's hard to test currently).

Meriipu:
Keeping "delete playlist" and "move to trash" (maybe also "remove from library" if that is added as a warning) always on sounds pretty reasonable for safety.

I was mostly trying to figure out if I were to try to add prompts for at least the proposed three issues, whether adding them to such a tab would be clean. 

[...]

PR #2746, opened 2018-01-26
Prompt on extending a playlist with another DND'ed playlist

### LINKED ISSUE #2367: Prompt on extending playlist with tracks from another dragged playlist
I dragged a playlist on top of another playlist in the playlist view in the belief that it would reorder them in that order, though it resulted in the target playlist being extended with all the tracks in the dragged playlist.

A prompt for this operation would in my belief in most cases not add a lot of extra interaction, while avoiding this particular somewhat more unfortunate interaction.

Meriipu:
initial attempt to fix #2367 
and in line with a few other commits/PRs regarding prompts for actions that are annoying to do on accident (see e.g #2453 )

A few issues with this:

* I dislike that I handled it by simply not calling qltk.selection_set_songs(sel, songs) if the prompt failed, which to my surprise was enough to not (visibly at least) have any tracks be removed/changes occur. This feels like a dodgy way to do it to me though so I might have overlooked something here... 
* _drag_data_get was the only place I found that gets called when a playlist is dragged onto another playlist, that is not also being called on any DND action whatsoever (like when tracks are getting dragged onto a playlist). Prompting in the latter case would in my opinion be annoying as dragging tracks onto a playlist is way more common than dragging a playlist onto another playlist.
* To add to the above, I did not find a way to differentiate between a dragged playlist and dragged tracks in the first method I looked into (__drag_data_received, I think it was?) 
* I found no way of getting the name of the target playlist (that tracks are being added to), in the current version of this commit it is simply referred to as "target_playlist"
* test case monkeypatches the run method of the prompt in a similar way to #2675

otherwise it seems to work from the testing I have done, but are there better ways to achieve this?

### DIFF
--- a/quodlibet/browsers/playlists/util.py
+++ b/quodlibet/browsers/playlists/util.py
@@ -8,7 +8,7 @@
 
 import os
 
-from quodlibet import _, print_w
+from quodlibet import _, print_w, ngettext
 from quodlibet import formats
 from quodlibet.qltk import Icons
 from quodlibet.qltk.getstring import GetStringDialog
@@ -39,6 +39,27 @@ def confirm_remove_playlist_dialog_invoke(
     ok_icon = Icons.EDIT_DELETE
 
     dialog = Confirmer(parent, title, description, ok_text, ok_icon)
+    prompt = dialog.run()
+    response = (prompt == Confirmer.RESPONSE_INVOKE)
+    return response
+
+
+def confirm_dnd_playlist_dialog_invoke(
+    parent, songs, target_playlist_name, Confirmer=ConfirmationPrompt):
+    """see confirm_remove_playlist_dialog_invoke above, except for
+       the action of attempting to extend a playlist with a second
+       dragged and dropped playlist.
+    """
+    title = ngettext(
+        "Extend \"{pl_name}\" with {num} additional track?",
+        "Extend \"{pl_name}\" with {num} additional tracks?",
+        len(songs),
+    ).format(pl_name=target_playlist_name, num=len(songs))
+
+    description = ""
+    ok_text = _("_Add Tracks")
+
+    dialog = Confirmer(parent, title, description, ok_text)
     prompt = dialog.run()
     response = (prompt == Confirmer.RESPONSE_INVOKE)
     return response
--- a/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/browsers/playlists/main.py
@@ -39,7 +39,9 @@ from quodlibet.util import connect_obj
 from quodlibet.util.collection import Playlist
 from quodlibet.util.dprint import print_d, print_w
 from quodlibet.util.urllib import urlopen
-from .util import parse_m3u, parse_pls, confirm_remove_playlist_dialog_invoke, _name_for
+from .util import parse_m3u, parse_pls, _name_for, \
+    confirm_remove_playlist_dialog_invoke, \
+    confirm_dnd_playlist_dialog_invoke
 
 DND_QL, DND_URI_LIST, DND_MOZ_URL = range(3)
 
@@ -376,6 +378,24 @@ class PlaylistsBrowser(Browser, DisplayPatternMixin):
         playlist = model[path][0]
         return playlist
 
+    def _add_drag_data_tracks_to_playlist(self, target_playlist, songs):
+        """helper-function to facilitate unit-tests without fiddling with views"""
+        # Accidentally extending through DnD'ing a playlist can be a fairly
+        # large operation, so prompt the user before proceeding.
+        # See issue #2367
+        playlist_name = target_playlist.name
+        response = confirm_dnd_playlist_dialog_invoke(
+            self, songs, playlist_name, self.Confirmer)
+        if response:
+            target_playlist.extend(songs)
+            self.changed(target_playlist)
+            was_modified = True
+        else:
+            print_d("DnD-extension of playlist cancelled through prompt")
+            was_modified = False
+
+        return was_modified
+
     def __drag_data_received(self, view, ctx, x, y, sel, tid, etime):
         # TreeModelSort doesn't support GtkTreeDragDestDrop.
         view.emit_stop_by_name('drag-data-received')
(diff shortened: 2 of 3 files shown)


Meriipu:
ok so I see why some of my PRs end up with a ton of extra commits now, trying to fix merge conflicts on another branch and commiting the changes into the original branch.

declension:
This feels neater definitely, but can I suggest we avoid (further) static dependencies (tight coupling). I'll point to what I mean...

quodlibet/quodlibet/browsers/playlists/main.py
```diff
@@ -418,12 +418,20 @@ def __drag_data_received(self, view, ctx, x, y, sel, tid, etime, library):
...
+        #print('adding', source_playlist, '-->', target_playlist)
+
         songs = []
         for itr in iters:
             if itr:
                 songs += model[itr][0].songs
+
         if tid == 0:
-            qltk.selection_set_songs(sel, songs)
+            prompt = ConfirmPlaylistDND(self, source_playlist, target_playlist)
```
declension: This implementation means that `PlaylistsBrowser` has a hard-coded dependency on the `ConfirmPlaylistDND` class - these are generally bad, hence big OOP projects typically use [Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection).

quodlibet/quodlibet/browsers/playlists/util.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, parent, playlist):
                              Gtk.ResponseType.YES)
 
 
+class ConfirmPlaylistDND(object):
```
declension: Can't this just subclass the `ConfirmationPrompt`?

quodlibet/quodlibet/browsers/playlists/util.py
```diff
@@ -46,6 +47,18 @@ def __init__(self, parent, playlist):
...
+    def __init__(self, parent, source_playlist, target_playlist):
+        title = _('Add the songs from playlist "{A}" to playlist "{B}"?'
+            ).format(A=source_playlist, B=target_playlist)
+        description = ""
+        ok_text = _("_Add Songs")
+        self.runner = ConfirmationPrompt(parent, title, description, ok_text)
+
+    def run(self):
```
declension: Please add docstrings if there's a meaningful return result here (as there seems to be)

quodlibet/tests/test_browsers_playlists.py
```diff
@@ -262,13 +262,21 @@ def test_default_display_pattern(self):
...
+
         b = self.bar
         song = AudioFile()
         song["~filename"] = fsnative(u"foo")
         sel = MockSelData()
         qltk.selection_set_songs(sel, [song])
         b._drag_data_get(None, None, sel, DND_QL, None)
 
+        # add back old behaviour for good measure
+        util.ConfirmPlaylistDND.run = tmp_confirmal
```
declension: And this wouldn't be a problem either then 😄

quodlibet/tests/test_browsers_playlists.py
```diff
@@ -262,13 +262,21 @@ def test_default_display_pattern(self):
         self.failUnless("<~name>" in pattern_text)
 
     def test_drag_data_get(self):
+        # Temporarily replace code that creates a prompt for the user
```
declension: So this is exactly what you have to do (monkey-patching and mutating global state) if you don't have the loose coupling.

If we'd injected the confirmer function / object (e.g. as an optional constructor parameter to the `PlaylistsBrowser`), we could simply instantiate the browser here in tests passing in a _mock_ confirmer - one with a `run` implementation similar to the one you've patched in below - but avoiding any monkey patching.

Meriipu:
A bit late, but I have problems seeing how an optional constructor parameter with a confirmer to the `PlaylistsBrowser` would make the coupling looser in a clean way.

Playlists have several prompts (this DND one, remove playlist, possibly remove songs from playlist), so would the constructor have to be passed one for each of these?

I can kind of see the confirmer being simple like a true/false lambda, and the confirmer is passed to the prompt-classes or prompt-functions (confirmDND, confirm_playlist_song_removal_invoke, ...), and then the prompt-functions either run their own confirmer (the gtk-dialog) or alternatively run the confirmer supplied via the parameter. 

This does not seem to remove the coupling though (does it?), just work around having to mutate global state.

Meriipu:
Is the strategy followed here more or less how you intended to make the coupling looser? #3396
There is no *hard* dependency on ConfirmationPrompt as long as an external confirmer is passed in, but it still gets imported (and used otherwise). I do not see a way to avoid that.

declension:
Hi! I'm struggling to remember that conversation but will have a bit more time this weekend perhaps.

In general, I think the new one seems closer to what I imagined which is 
Function of `*args -> bool` type
Then that can be fulfilled by, say a `GtkDialog.run` method, or equally by a simple `lambda _: True` in testing.

The callpoint code just calls `self.confirmer()` to see.

Does that help?

Meriipu:
I think I understand yes if
* the PlaylistsBrowser code does what it already does, and if a promptable action occurs then it calls needs_confirmation(args, ..., confirmer) 
* confirmer is usually none at init of PlaylistsBrowser but a unit-test might set it to a lambda.
* needs_confirmation by default has a gtkdialog as its confirmer, but if it gets passed a confirmer-lambda it instead calls this (and passes the lambda all the args it would have passed the gtkdialog)

on the other hand if the gtkdialog also needs to be passed in on initialization of PlaylistsBrowser then it may take some more time to understand.

Meriipu:
Updated with the newer method for making prompts and also has unit tests.

Since the source is not always a playlist, I decided it is easier to just show the number of tracks that the target would be extended with.
The target playlist is shown with the name and current number of tracks before extending.

Meriipu:
Maybe it is a good idea to find a way to differentiate between the source being a playlist, and the source being other tracks.
I imagine dragging from a browser to a playlist may get tedious with a prompt.

Meriipu:
Fixed code to work with recent changes to playlists/libraries

there might be other ways to DnD onto playlists that I have missed/not tested (like external playlist files?)

main issue is the case of dragging a few individual songs from a browser onto an element in the playlist-list. This causes a prompt too, which might be annoying; I did not find a way to distinguish between the source of the DnD being a playlist and the source being tracks.

In my case the annoyance of accidentally dragging a playlist is larger than a prompt when dragging tracks, but I understand if others have a different tradeoff here.

quodlibet/browsers/playlists/util.py
```diff
@@ -44,6 +44,25 @@ def confirm_remove_playlist_dialog_invoke(
     return response
 
 
+def confirm_dnd_playlist_dialog_invoke(
+    parent, songs, target_playlist_name, Confirmer=ConfirmationPrompt):
+    """see confirm_remove_playlist_dialog_invoke above, except for
+       the action of attempting to extend a playlist with a second
+       dragged and dropped playlist.
+    """
+    title = ngettext("Extend playlist {pl_name} with {num} additional track?",
```
declension, line 53:
```suggestion
    title = ngettext("Extend \"{pl_name}\" with {num} additional track?",
```

quodlibet/browsers/playlists/util.py
```diff
@@ -44,6 +44,25 @@ def confirm_remove_playlist_dialog_invoke(
...
+                     "Extend playlist {pl_name} with {num} additional tracks?",
```
declension, line 54:
```suggestion
                     "Extend \"{pl_name}\" with {num} additional tracks?",
```

declension (approved):
Tried it , LGTM thanks. See my suggestion

Meriipu:
Should be fixed now

declension:
Thanks @Meriipu !

PR #2761, opened 2018-02-05
Recognize rating/playcount tags in vorbis comments

### LINKED ISSUE #2760: QL doesn't recognize some rating/playcount tags in vorbis comments
If the "email" for rating/playcount tags contains upper-case characters those tags are not correctly identified because mutagen uses lower-case keys. 

Example:
"Email" is `Windows Media Player 9 Series`
Given a file with a tag `playcount:Windows Media Player 9 Series`
mutagen will return it as `playcount:windows media player 9 series` and as a result QL does not use it for the internal ~#playcount tag. Same for the rating tag.

slosd:
Use lower-case keys when looking for rating/playcount tags from mutagen.


lazka:
Can you add a test?

slosd:
Sure. I simply changed existing tests to include an upper-case letter in the email. Without my patch every single one fails.

PR #2772, opened 2018-02-16
Fix raising NotImplementedError

jwilk:
`NotImplemented` is a sentinel object, not an exception type.
It was mistakenly used instead of `NotImplementedError`.


lazka:
Thanks!

Issue #2768, opened 2018-02-08
AttributeError in qlscrobbler.py when trying to scrobble to last.fm


hell-g:
I set up AudioScrobbler Submission plugin to scrobble to last.fm. The configuration dialog for the AudioScrobbler Submission gives me an authentication successful when I hit the verify account data button. I was expecting, that the songs I play in quodlibet would show up in my last.fm account page. But I could not see any of the songs I played on my last.fm profile page. I tried with other software that I had set up to scrobble to last.fm and those worked. 

When starting quodlibet from the terminal, I see the following output after I start playing the first song:

> Exception in thread Thread-1:
> Traceback (most recent call last):
>   File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
>     self.run()
>   File "/usr/lib/python3.6/threading.py", line 864, in run
>     self._target(*self._args, **self._kwargs)
>   File "/usr/lib/python3.6/site-packages/quodlibet/ext/events/qlscrobbler.py", line 224, in run
>     if self.send_submission():
>   File "/usr/lib/python3.6/site-packages/quodlibet/ext/events/qlscrobbler.py", line 316, in send_submission
>     data['%s[%d]' % (key, idx)] = val.encode('utf-8')
> AttributeError: 'bytes' object has no attribute 'encode'

I am running Quod Libet 4.0.2, Mutagen 1.39 on Arch Linux 4.15.1-2-ARCH #1 SMP x86_64 with python 3.6.4. 

Thank you for looking into this!!

lazka:
thanks

lazka:
for reference: #2778

PR #2778, opened 2018-02-17
qlscrobbler: Fix issue #2768

luk1337:


quodlibet/quodlibet/ext/events/qlscrobbler.py
```diff
@@ -313,12 +313,14 @@ def send_submission(self):
...
             for key, val in song.items():
-                data['%s[%d]' % (key, idx)] = val.encode('utf-8')
+                data['%s[%d]' % (key, idx)] = val
             data['o[%d]' % idx] = 'P'
             data['r[%d]' % idx] = ''
 
         print_d('Submitting song(s): %s' %
-            ('\n\t'.join(['%s - %s' % (s['a'], s['t']) for s in to_submit])))
+            ('\n\t'.join(['%s - %s' % (
+                s[b'a' if b'a' in s else 'a'],
```
lazka: that looks hacky, why are there mixed types here?

luk1337: I have no idea. I dumped my 'to_submit' and it had both 'a' and b'a' for some reason.

luk1337: It might have been a part of unsubmitted songs that I played while plugin wasn't working but I'm unsure about that.

lazka:
I went for 9f826ce3c033 for now and 1f7a5b9725 so we get error reports in case that wasn't it.

Issue #703, opened 2015-03-14
Support un-shuffleable tracks for DJ mixes / classical pieces


lazka:
Original [issue 703](https://code.google.com/p/quodlibet/issues/detail?id=703) created by nick.boultbee on 2011-04-03T20:27:50.000Z:

From wiki post (http://code.google.com/p/quodlibet/wiki/Contact): 

Is there a way to set up Quod Libet so that it plays all the movements of a classical music piece in sequence? I prefer to use shuffle, rather than bother with creating playlists. But when QL happens to pick a track in its shuffle that is the first movement of a piece, I'd like it to play it, then play all the other movements of the same piece, in sequence, then go back to shuffle. And if QL picks a track in its shuffle that is the second, third, fourth, etc. movement in its shuffle, I'd it to detect that they are non-initial movements and skip over them. I know I can create tags to keep track of the movements. I just don't know how to get QL to use them the way I describe.

archy3:
This too has been an issue for me.

One solution I can think of is in the pane browser, allow for random item selection in the rightmost pane upon completion of the Song List.

So if I group all movements in a piece together under the grouping tag, and I want to shuffle by grouping, I could set my rightmost pane to <grouping|<grouping><album>|<album><discnumber><tracknumber>> and then shuffle that pane.

This would be very flexible and extendible. The previous panes would allow for easy filtering, and users could define groups however they want. It probably wouldn't be to hard to extend this to more than just the rightmost pane, so one day someone could shuffle by the Grouping pane and another day shuffle by the Album pane (all panes to its right would be set to all) without first removing the grouping pane.

The only downside is that having a pane like <grouping|<grouping><album>|<album><discnumber><tracknumber>> would add a lot of clutter and probably would not be used to actually drill down the library but just for this randomization purpose. Perhaps "randomization only" panes could be made invisible though it would make the Song List look odd given the visibly selected pane items (that probably wouldn't be a good fix, just speaking out loud here).

The above is the simplest solution I can think of. It could almost be already implemented by swapping the Grouping tag with the Album tag and using the Random Album Plugin [but then I wouldn't be able to shuffle by album anymore :(].

[...]

PR #2788, opened 2018-02-25
Add shuffle-by-grouping plugin #703

archy3:
This allows users to shuffle by a grouping of songs defined by a common tag, solving #703.

### DIFF
--- a/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
+++ b/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
@@ -38,18 +38,18 @@ class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
     display_name = _("Shuffle by grouping")
     priority = Reorder.priority
 
-    next_explicit_called = False
-
     def next(self, playlist, current):
+        return self._next(playlist, current)
+
+    def _next(self, playlist, current, explicit=False):
         grouping = str(pconfig.gettext("grouping")).strip()
         grouping_test = str(pconfig.gettext("grouping_test")).strip()
         delay = pconfig.getint("delay")
         delay_on = True
 
         # Check if explicitly called and alter variables accordingly
-        if self.next_explicit_called:
+        if explicit:
             delay_on = False
-        self.next_explicit_called = False
 
         # Keep track of played songs
         OrderRemembered.next(self, playlist, current)
@@ -106,8 +106,7 @@ class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
         return False
 
     def next_explicit(self, playlist, current):
-        self.next_explicit_called = True
-        return self.next(playlist, current)
+        return self._next(playlist, current, explicit=True)
 
     def previous(self, playlist, current):
         return OrderRemembered.previous(self, playlist, current)
--- a/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
+++ b/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
@@ -41,15 +41,10 @@ class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
     def next(self, playlist, current):
         return self._next(playlist, current)
 
-    def _next(self, playlist, current, explicit=False):
+    def _next(self, playlist, current, delay_on=True):
         grouping = str(pconfig.gettext("grouping")).strip()
         grouping_test = str(pconfig.gettext("grouping_test")).strip()
         delay = pconfig.getint("delay")
-        delay_on = True
-
-        # Check if explicitly called and alter variables accordingly
-        if explicit:
-            delay_on = False
 
         # Keep track of played songs
         OrderRemembered.next(self, playlist, current)
@@ -106,7 +101,7 @@ class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
         return False
 
     def next_explicit(self, playlist, current):
-        return self._next(playlist, current, explicit=True)
+        return self._next(playlist, current, delay_on=False)
 
     def previous(self, playlist, current):
         return OrderRemembered.previous(self, playlist, current)
--- a/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
+++ b/quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
@@ -134,22 +134,21 @@ class ShuffleByGrouping(ShufflePlugin, OrderRemembered):
             grouping_filter_entry.set_text(pconfig.gettext("grouping_filter"))
             delay_spin.set_value(pconfig.getint("delay"))
 
-        vbox = Gtk.VBox(spacing=12)
-        vbox.set_border_width(0)
+        def make_label(label_text):
+            label = Gtk.Label(label_text, selectable=True)
+            label.set_alignment(0.0, 0.5)
+            label.set_margin_end(3)
+            return label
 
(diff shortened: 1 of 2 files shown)


archy3:
Not sure why a Windows test is failing. I just set up a msys2 environment and ran a full test and no tests failed. I noticed another recent plugin PR (#2778) failed the same test on Windows.

lazka:
the test just timed out, ignore it

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+pconfig = PluginConfig("shufflebygrouping")
+pconfig.defaults.set("grouping", "~grouping~album~albumartist")
+pconfig.defaults.set("grouping_test", "grouping")
+pconfig.defaults.set("delay", 0)
+
+
+class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
+    PLUGIN_ID = "shufflebygrouping"
```
declension: On a recent count we had four separate naming conventions for this (argh), but the most common is just `Title Case Plugin` (35ish) `followed by `CamelCasePlugin` (17ish), less for `snake_case` and least for `modulestyle` as here, so it'd be great if you went for one of the first two...

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+from quodlibet.qltk import Icons
+
+
+pconfig = PluginConfig("shufflebygrouping")
+pconfig.defaults.set("grouping", "~grouping~album~albumartist")
+pconfig.defaults.set("grouping_test", "grouping")
+pconfig.defaults.set("delay", 0)
+
+
+class ShuffleByGrouping(ShufflePlugin, OrderInOrder, OrderRemembered):
```
declension: Is `OrderInOrder` needed here (for a shuffle plugin)?

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+        grouping = str(pconfig.gettext("grouping")).strip()
+        grouping_test = str(pconfig.gettext("grouping_test")).strip()
+        delay = pconfig.getint("delay")
+
+        # Keep track of played songs
+        OrderRemembered.next(self, playlist, current)
+        remaining = OrderRemembered.remaining(self, playlist)
+
+        # Check if playlist is finished or empty
+        if len(remaining) <= 0:
```
declension: `if remaining` is more Pythonic here (we assume it's never <0)

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+            OrderRemembered.reset(self, playlist)
+            return None
+
+        # Play next song in current grouping
+        next_song = super().next(playlist, current)
+        if (current is not None and next_song is not None and
```
declension: This long multi-line conditional would be good as a function...

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+            self._tag_defined(grouping_test, playlist, next_song) and
+            self._same_tag(grouping, playlist, current, next_song)):
+            return next_song
+
+        # Pause for a moment before picking new group
+        if delay_on and delay > 0:
+            app.player.paused = True
```
declension: BTW check the Random Album plugin for a nice UI-linked way of doing this (which is cancellable by the user)

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+                break
+            new_song_prev = playlist.get_iter(song_location - 1)
+            if not self._same_tag(grouping, playlist, new_song, new_song_prev):
+                break
+
+        return new_song
+
+    @staticmethod
+    def _tag_defined(tag_name, playlist, song_iter):
+        if tag_name == "":
```
declension: `if not tagname` is more Pythonic I'd argue

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+            return True
+        song = playlist.get_value(song_iter)
+        tag_value = song(tag_name)
+        if tag_value.strip() != "":
```
declension: `return bool(tag_value.strip())`?

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+            return True
+        return False
+
+    @staticmethod
+    def _same_tag(tag, playlist, song_iter_a, song_iter_b):
+        song_a = playlist.get_value(song_iter_a)
+        song_b = playlist.get_value(song_iter_b)
+        if song_a(tag) == song_b(tag):
```
declension: `return song_a(tag) == song_b(tag)`

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
```diff
@@ -0,0 +1,178 @@
...
+from quodlibet.order import OrderRemembered
+from quodlibet.order.reorder import Reorder
+from quodlibet.plugins import PluginConfig
+from quodlibet.plugins.playorder import ShufflePlugin
+from quodlibet.qltk import Icons
+
+
+pconfig = PluginConfig("shufflebygrouping")
+pconfig.defaults.set("grouping", "~grouping~album~albumartist")
+pconfig.defaults.set("grouping_test", "grouping")
```
declension: perhaps `grouping_filter` is a clearer name (though I might have misunderstood this)?

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
archy3: I was going to say yes because this plugin actually disables shuffling if the next song belongs to the same group as the current song, but removing it and replacing `next_song = super().next(playlist, current)` with `next_song = OrderInOrder.next(self, playlist, current)` seems to work (perhaps because the body of OrderInOrder.next does not ever use `self`, so is essentially static).

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
archy3: `grouping_test` was the variable name I was most displeased with. Its purpose is to see if grouping should be applied to a track. At first I just checked if the `grouping` tag was defined, but this had problems when the `grouping` tag was a tied tag (which is useful to distinguish different recordings of the same work). I agree that `grouping_filter` is a better name.

quodlibet/quodlibet/ext/playorder/shufflebygrouping.py
archy3: What is the purpose of this try-except block?
Lines 20-23 of [randomalbum.py](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/ext/events/randomalbum.py#L20):
```
try:
    from quodlibet.qltk import notif, Icons
except Exception:
    notif = None
```

declension:
Thanks!

PR #2791, opened 2018-02-25
travis-ci: try adding a macos 10.13 job

lazka:


lazka:
That's... not helpful :/

I'll try to update the macOS env some time and maybe include pytest-faulthandler

lazka:
Updating things didn't help so that needs to be debugged in 10.13.

PR #2797, opened 2018-02-28
Independent cover size for Album List and Cover Grid browser

urielz:
This PR allows to have independent values for the covers displayed in the Album List and Cover Grid browsers. The default value for the Cover Grid is set to 96, however users can set their preferred sized via the Advanced Preferences plugin.

### DIFF
--- a/quodlibet/quodlibet/browsers/albums/models.py
+++ b/quodlibet/quodlibet/browsers/albums/models.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2013 Christoph Reiter
+#           2018 Uriel Zajaczkovski
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


declension:
BTW you realise there's a scale factor in the Covergrid browser (which is independent of the Album List browser, and relative to the underlying preferences for both)?

urielz:
Is it really independent? It looks to me that it uses the cover size value of Album List browser multiplied by the scale factor...

declension:
Sorry @urielz, forgot to reply. Yes it's accessible via the covergrid prefs UI - click the cog in the right of the search bar in the browser and it's one of the settings there...

urielz:
@declension thanks. missed that! 

It's completely unintuitive that the cover grid size for Cover Grid Browser is accessible via that menu and the cover grid size for Album List Browser is hidden in a plug-in menu in advanced preferences. 

I might give it a go and a slider for Album List cover size just like in Cover Grid. Unless there's a reason it's hidden in advanced preferences... ?

declension:
> It's completely unintuitive that the cover grid size for Cover Grid Browser is accessible via that menu and the cover grid size for Album List Browser is hidden in a plug-in menu in advanced preferences.

Agreed. I only noticed it randomly at some point...

> I might give it a go and a slider for Album List cover size just like in Cover Grid. Unless there's a reason it's hidden in advanced preferences... ?

Yep, I definitely think this isn't an _advanced_ preference but belongs with the browser. For consistency if nothing else! Thanks :+1:

urielz:
Thanks! Will close this now.

PR #2799, opened 2018-03-03
Add cli options for setting repeat and shuffle type

### LINKED ISSUE #2785: Change repeat mode from the command line
Currently it's possible to set or toggle Repeat and Shuffle from the command line, but as far as I know there's no way to change the Repeat Mode (Repeat This Track and Repeat All), or the Shuffle Mode for that matter.

Could this functionality be added at some point?

frestr:
Fixes #2785 

There are a few things to consider here, which I would like to have some input on:

- Should order plugins (like Reverse) also be adjustable via the command line? If so, what should happen if set while the plugin is not enabled?
- Should changing the mode type also turn on the mode? Worth mentioning since setting multiple options at once doesn't seem to work too well.
- This change appears to have some overlap with the '--order' option mentioned in the manpage, but that option doesn't seem to have been implemented... I'm also not sure if the format of the --order option maps too well against the GUI.

### DIFF
--- a/quodlibet/data/quodlibet.1
+++ b/quodlibet/data/quodlibet.1
@@ -128,6 +128,10 @@ Refresh and rescan library
 .B \-\-repeat=off|on|t
 Turn repeat off, on, or toggle
 .TP
+.B \-\-repeat\-type=current|all|one
+Repeat the currently playing song, the current list, or stop after
+one song
+.TP
 .B \-\-seek=[+|\-][HH:]MM:SS
 Seek within the playing song
 .UNINDENT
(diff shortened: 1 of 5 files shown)


lazka:
> Should order plugins (like Reverse) also be adjustable via the command line? If so, what should happen if set while the plugin is not enabled?

Not a big fan of exposing all this in the CLI, so I'd prefer a minimal implementation, it at all.

> Should changing the mode type also turn on the mode? Worth mentioning since setting multiple options at once doesn't seem to work too well.

What about yes, turn on the mode, and in addition add an "off" option to both repeat-type/shuffle-type? That way it's just a more specialized variant of the non-type commands and can be used independently.

> This change appears to have some overlap with the '--order' option mentioned in the manpage, but that option doesn't seem to have been implemented... I'm also not sure if the format of the --order option maps too well against the GUI.

Yeah, that got lost in the last rewrite, should be removed from the man page. Please ignore.

frestr:
> Not a big fan of exposing all this in the CLI, so I'd prefer a minimal implementation, if at all.

Ok, then I think we'll keep it as it is now.

> What about yes, turn on the mode, and in addition add an "off" option to both repeat-type/shuffle-type? That way it's just a more specialized variant of the non-type commands and can be used independently.

Sounds good.

Issue #2774, opened 2018-02-16
It still will not STOP


catmatist:
As of 4.0.2, there is still no effective way to tell quodlibet to always stop after playing what I told it to play. I have to tell it every time "stop after this file". I know there are more features to make it stop sometimes, but they just don't work for me, they are too complicated and what I want is so simple.

declension:
What's wrong with _Control_ -> _Stop After This song_?

Or if you want one song at a time, use the _One Song_ repeat mode (and make sure repeat is on)

catmatist:
The "stop after this file" command is better than nothing, but it is still inconvenient to repeat this command every time I start playing a file. Often I play a file from some other program like a file manager or podcast catcher, and it is always an unpleasant surprise when something unrelated starts playing afterward. I have also tried the "Stop when empty" setting on the Queue, and the Queue Only plugin, yet somehow Quod Libet usually manages to play something else I didn't want after it plays what I told it to play. I guess I just don't understand the Queue at all, but I don't think I want to understand it really. Usually what I want to play is a single file, or an Album, or a Playlist I created. The Stop after this file command won't work for stopping after an Album or Playlist even if I could remember to use it. I want a saved setting that means always stop after playing whatever unit of material was specified in the command to start playing.

I can only assume that suggesting I turn repeat mode on to make the program STOP is some kind of a joke that I don't get.

declension:
> I can only assume that suggesting I turn repeat mode on to make the program STOP is some kind of a joke that I don't get.

No this is entirely correct (currently). The shuffle mode plugins control _what_ tracks are selected, and the repeat mode controls _how_ [often] they are played. The One Song plugin stops the player after every track - i.e. enforcing a 1 track list, whilst track-repeat and repeat-all are variations on an infinite-track list.

 See #2518 for related enhancement proposals.

catmatist:
It took a while to grasp what you were getting at with the One Song business. I see that it is a setting that sticks, and involves stopping play, but it still isn't what I'm looking for, because it doesn't address the case where I want to play multiple songs, and then stop.

The feature that comes closest, I think, to what I want is the "stop when queue is empty" option, even though what it really does is subtly different from what it says it does. What it really does is set the "stop after this track" flag when the last song is removed from the queue and made the playing track (or so it appears). What I expected from the name was that it would, when it reached the end of a song, look at the queue, and if the queue was empty, stop playing. That would have been more useful to me.

[...]

PR #2801, opened 2018-03-04
Queue stop once empty: do check at end of song instead

frestr:
Stops playback directly when a song has ended and the queue is empty, as opposed to enabling the "Stop After Empty" setting when the last song has started.

See #2774

### DIFF
--- a/quodlibet/quodlibet/qltk/queue.py
+++ b/quodlibet/quodlibet/qltk/queue.py
@@ -130,7 +130,6 @@ class QueueExpander(Gtk.Expander):
 
         clear_item = MenuItem(_("_Clear Queue"), Icons.EDIT_CLEAR)
         menu.append(clear_item)
-        clear_item.connect("activate", self.__clear_queue)
 
         button = SmallMenuButton(
             SymbolicIconImage(Icons.EMBLEM_SYSTEM, Gtk.IconSize.MENU),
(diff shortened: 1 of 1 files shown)


catmatist:
Interesting. What if you play a song directly, not going through the queue? With this change, will Quodlibet honor the "stop if queue is empty" flag when that song ends, or does that situation bypass this code entirely? I'll try to find time to figure out how to pull this from GitHub and test it, but it might take awhile as I'm new to this site.

frestr:
As it is now, playing a song directly will not bypass the setting. When I come to think of it, this is exactly the same as how Repeat>One Song works, so this is not really desirable.

IMO, playing a song directly should bypass the setting, while it is honored as long as you are playing from the queue. This makes it so that the setting doesn't have to be disabled when not playing from the queue.

If you want to test it, you can follow the instructions at [Creating a Development Environment](https://quodlibet.readthedocs.io/en/latest/development/devenv.html). The only difference will be that instead of cloning the main repo (`git clone https://github.com/quodlibet/quodlibet.git`), you'll have to clone my fork of it (`git clone https://github.com/frestr/quodlibet.git`) and change to the queue_stop_fix branch (`git checkout queue_stop_fix`) inside the cloned repo.

catmatist:
This is so frustrating! You had it working the way I wanted, and then when I asked how it worked, you changed it and made it useless. Or at least, that is the case if I have understood what you have said about how each version works. Does no one understand the idea that whenever possible, options should have the effect they say they have, and not something subtly different instead? Do I have to check and see if the "one song" option has no effect when songs are playing from the queue? Because the effect I was looking for was exactly that to play more than one song, I can add them to the queue, but if I just "play" a song, Quodlibet does not go on to play some other song that I did not tell it to play.

frestr:
See my reply at #2774

Issue #374, opened 2015-03-14
Add wildcard support for selecting artwork (cover) file


lazka:
Original [issue 374](https://code.google.com/p/quodlibet/issues/detail?id=374) created by forumscollective on 2009-12-28T05:52:40.000Z:

I'm not sure what algorithm is presently used to ascertain which file
represents the album cover when scanning a library, but it would be really
useful to be able to specify a filename as the coverart.  As an example, in
my library, the coverart is always folder.jpg for every album, but
quodlibet on occasion picks up other jpg files as the coverart.

Implementing this would enable users to optionally specify what file should
be considered the album cover and would also possibly speed up the library
scan process (ever so slightly).

brunob:
Copy/paste of patch proposed in https://github.com/quodlibet/quodlibet/issues/374#issuecomment-80744036 :

```diff
diff -r aa9c4c504835 quodlibet/quodlibet/formats/_audio.py
--- a/quodlibet/quodlibet/formats/_audio.py	Tue Jul 05 10:19:50 2011 +0200
+++ b/quodlibet/quodlibet/formats/_audio.py	Tue Jul 05 10:29:27 2011 +0200
@@ -588,13 +588,15 @@
             return self.get_format_cover()
 
         base = self('~dirname')
-        images = []
+        fns = []
+        add_negative = False
 
         # Issue 374: Specify artwork filename
         if cfg_getboolean("albumart", "force_filename", False):
+            add_negative = True
             path = os.path.join(base, config.get("albumart", "filename"))
-            if os.path.isfile(path):
-                images = [(100, path)]
+            for path in glob.glob(path):
+                fns.append((None, path))
         else:
             get_ext = lambda s: os.path.splitext(s)[1].lstrip('.')
 
@@ -617,33 +619,34 @@
                         if get_ext(lsub_entry) in self.__cover_exts:
                             fns.append((entry, sub_entry))
 
-            for sub, fn in fns:
-                score = 0
-                lfn = fn.lower()
-                # check for the album label number
-                if "labelid" in self and self["labelid"].lower() in lfn:
-                    score += 20
+        images = []
+        for sub, fn in fns:
+            score = 0
+            lfn = fn.lower()
+            # check for the album label number
+            if "labelid" in self and self["labelid"].lower() in lfn:
+                score += 20
 
-                # Track-related keywords
-                keywords =  [k.lower().strip() for k in [self("artist"),
-                    self("albumartist"), self("album")] if len(k) > 1]
[...]

PR #2814, opened 2018-03-23
Add wildcard support to `albumart.filename` pref #374

zsau:
I updated the patch in #374 for current master. Can't tell who wrote the original from the issue history, but credit goes to them.


lazka:
thanks

Issue #2792, opened 2018-02-26
Reverse DNS App ID


lazka:
For flatpak we need to unify our dbus name/desktop id/appdata id.

We currently only have "net.sacredchao.QuodLibet" for dbus. We could either just use that or go with "io.github.quodlibet.QuodLibet" and "io.github.quodlibet.ExFalso" maybe.

Open questions: Is there anything we can do to improve the transition (re gnome software appdata etc)

declension:
+1 to `io.github.quodlibet.QuodLibet` etc

lazka:
on IRC, re transition:

```
<brtln> you can use <provides> <id>old_id</id></provides> in your appstream file
```

fkrull:
Another point: for any Flatpak that exports multiple desktop files, all the desktop files must have the app ID as prefix (see https://github.com/flatpak/flatpak/issues/750). If Ex Falso should be included in the same Flatpak as QL, its app id would have to be something like "io.github.quodlibet.QuodLibet.ExFalso" so that it's prefixed with the QL app id.

(Something I just thought of: it might be possible to set the Flatpak app id to "io.github.quodlibet" with desktop files "io.gh.ql.QuodLibet" and "io.gh.ql.ExFalso". But I don't know if that's allowed or even a good idea.)

lazka:
@fkrull thanks for the info. I slightly tend towards separate flatpaks for QL/EF, so we get multiple entries/screenshots in gnome software/app stores etc. Any thoughts?

fkrull:
@lazka The one issue with that is that the applications would not share a config directory since they'd be in different Flatpaks, so if someone set up their tag editing plugins in QL, they wouldn't show up in EF. I don't terribly mind that personally but it's something to keep in mind.

Separate packages should hopefully solve the app identity issue I was having with EF over in #1806, though.

lazka:
Uh, ok. I don't mind either really. Maybe we can just start with QL only and handle EF later.

lazka:
> How important is backwards compatibility for the D-Bus names?

If you mean net.sacredchao.QuodLibet, I don't mind if that goes away with flatpak.

lazka:
I've updates the PPAs for the file renames, if anyone sees anything funky please report.

Thanks @fkrull

PR #2842, opened 2018-05-12
Reverse DNS app ID

fkrull:
See issue #2792. I renamed everything I could find that needs to match the Flatpak app ID, in particular:

* the desktop file
* the appdata file
* the D-Bus names
* the application icon
* the GNOME search provider

This does mean the `net.sacredchao.QuodLibet` D-Bus name goes away. Any third-party clients using that name that may or may not exist would have to be updated.

I also added `StartupWMClass` to the desktop file. Without it, Unity had difficulties matching the window to the desktop file.

I left Ex Falso alone for the moment.


lazka:
Thanks!

Can your revert the "quodlibet/quodlibet/qltk/dbus_.py" changes?

I'd prefer if it just gets blocked in flatpak (is that a problem?), and once we move MPRIS into the core we might drop it altogether.

fkrull:
I played around with it and it turns out that yes, I can. ;) I left the other D-Bus names (search provider and the app menu) as-is since no one should write those names down and talk to them.

lazka:
Thanks :)

CreamyCookie:
Regarding the icon, it seems that with this change, every icon pack ([Numix Square](https://github.com/numixproject/numix-icon-theme-square), for example) has to create a symlink from "quodlibet" to "io.github.quodlibet.QuodLibet". It's not a big hassle, but is that change necessary?

lazka:
> It's not a big hassle, but is that change necessary?

afaik we could have only done it for the flatpak build, but then you would have two names depending on how things are installed. But maybe I've missed something?

fkrull:
> afaik we could have only done it for the flatpak build, but then you would have two names depending on how things are installed

Well, there's no way to *not* rename the icon for Flatpak. Having both names sounds like a maintenance nightmare to me, especially considering how many places this PR ended up touching. Also, any icon theme would have to support both names anyway if they want to work for Flatpak installations.

CreamyCookie:
Makes sense! I'm looking forward to flatpak support. =)

lazka:
> Makes sense! I'm looking forward to flatpak support. =)

No need to wait :) https://flathub.org/apps/details/io.github.quodlibet.QuodLibet

PR #2086, opened 2016-11-05
fix for #2072 - The "random album" plugin thinks it can function in the playlist browser

draxil:
Fix for #2072 - The "ugly" fix mentioned in that issue that should stop the random album plugin causing annoying effects within the playlists browser until the architecture is ready to enact one of the other fixes mentioned in that issue.

PR #2844, opened 2018-05-15
plugins: Make random album work on non-album browsers again

### LINKED ISSUE #2809: Allow random album plugin to work on non-album browsers again
From memory, this worked, but doesn't any more. Maybe this is good, but I miss the random album plugin, yet use search or playlist browsers mainly...

There is code there to deal with this though.

frestr:
Fixes #2809 

It seems like #2086 introduced this bug. `browser.can_filter("album")` was changed to `browser.can_filter_albums()`, which (albeit a bit confusingly) is more restrictive. Furthermore, the playlists browser check is categorically false, which is kind of overshadowed by the first check.

It's a simple fix that seems to have solved the issue, but it would be nice if someone could check over it to make sure I haven't overlooked anything.


frestr:
Thanks

Issue #2771, opened 2018-02-16
Cannot switch output (Mac)


kylebfredrickson:
Suppose I have a laptop connected to a monitor via HDMI. When I switch output from the mac internal speakers to the speakers on the monitor, the app fails to switch output. Also, if I disconnect the HDMI and then plug it back in the app will crash.

madeindjs:
This problem still exists on QuodLibet 4.3.0 / Mac OS Catalina 10.15.4

marians:
I can confirm this problem for macOS BIg Sur v11.1.

Also https://github.com/quodlibet/quodlibet/issues/3210 seems to be related or even a duplicate.

chrisresnikrakuten:
I originally created #3210
A workaround I found out in the meantime: double-click a track to restart it, the new audio device will be used
as workaround, it's acceptable for me but then again, I see other software such as Zoom that automatically recognise new audio devices and start using them as default device (plug in usb-headsets or removing one). so I'm not sure what prevents quodLibet from doing so.
apart from, quite a nice player.

waaykool:
@chrisresnikrakuten 

Thanks for this tip! I am having the same issue and have been quitting and restarting the app in order to change audio device. Double-click on a track is totally acceptable for me as well and works just fine.

solstice0:
A stop button would be helpful for me too.
I use
`alsasink device=hw:3,0`
in my output pipeline, so gstreamer has exclusive rights to my DAC. This is how I want it. BUT sometimes I want to stop listening to music and watch a video or something in my browser, then resume listening to music. Right now I need to close quod libet to do so. If I'm understanding this plugin correctly, I could just hit stop and I could watch a video in my browser, then hit play in quod libet, and it will initialize gstreamer again. That would be nice to have.

Issue #2845, opened 2018-05-16
New Backend: VLC


othalan:
I am interested in using VLC as a new backend for QuodLibet.

The VLC project provides automatically generated python bindings available either through a PyPi package [python-vlc](https://pypi.org/project/python-vlc/) or directly from their web site as defined in the [VLC Wiki for Python Bindings](https://wiki.videolan.org/python_bindings).  VLC is stable and provides robust support both for audio file formats and output devices.

My primary use for VLC as an alternate backend is as a fix for bug #2771 which I encounter in the form of difficulties switching to a Bluetooth speaker.  VLC detects changes in audio device output and switches automatically.

The only difficulty is including VLC as a backend is it requires the VLC application to be installed before the python bindings can be used.

I have working code for a new VLC backend and am testing it on my own system.  I am happy to submit a pull request when I am finished, however I am uncertain about the best implementation.  I have considered the following possibilities and would like feedback on the preferred implementation:

- Create a new backend with the requirement that the user install VLC before it may be used.
  - This is my current implementation.
- Create a plugin for enabling VLC as an alternative backend, noting in the plugin that VLC must be installed separately.
  - I do not see an ability to add a custom backend through plugins under the current plugin structure, but I may be missing something.

I also considered and rejected the possibility of including a compiled version of the VLC player locally in the QuodLibet distribution in order to eliminate the installation issue but rejected that because it seems an excess of work for a backend which is likely to be used by a relatively small number of people.

Any thoughts on including a VLC backend option?

lazka:
> The only difficulty is including VLC as a backend is it requires the VLC application to be installed before the python bindings can be used.

Does this mean installing VLC as a separate app is enough as long as QL includes the Python bindings?

> I have working code for a new VLC backend and am testing it on my own system. I am happy to submit a pull request when I am finished, however I am uncertain about the best implementation. I have considered the following possibilities and would like feedback on the preferred implementation:
> 
>     Create a new backend with the requirement that the user install VLC before it may be used.
>         This is my current implementation.

I'd prefer this.

>     Create a plugin for enabling VLC as an alternative backend, noting in the plugin that VLC must be installed separately.
>         I do not see an ability to add a custom backend through plugins under the current plugin structure, but I may be missing something.

Yeah, there is not plugin interface for this.

[...]

PR #2848, opened 2018-05-17
New plugin to send a stop signal to the player backend on pause

othalan:
This is a plugin to workaround issue #2771, which appears to be a fault in the audio backend.

The basic problem solved by this plugin:

On macOS, play music through the system speakers then connect a bluetooth speaker.  Typically audio output will transition to the new bluetooth speaker.  However, QoudLibet and other media players using the gstreamer backend will not transition to the new bluetooth speaker (or other audio output device) as expected.  This behavior has also been observed in other media players using gstreamer, such as Clementine player.

In order to provide a workaround for this problem, this plugin will detect Pause events and send a stop signal to the backend player.  This relies on the fact that the gstreamer backend will destroy the current pipeline when stop() is called. 

In order to maintain current functionality to the best extent possible, the plugin attempts to seek to the same playback position after stopping playback, thus emulating pause functionality.  If the stream reports it is not seekable, playback is left stopped.  An option in the plugin can be set to only enable this functionality for seekable sources (in order to ensure playback position is not lost by using this plugin).


othalan:
It occurs to me I should note that the VLC backend I am working on (see #2845) will also provide a solution to this same bug.  I have submitted this plugin for two reasons:  First, for the record that this solution is available (I have been using it for months), and second because using VLC as a backend may not be desirable for some people.

lazka:
What about just restarting QL instead? It should restore all state and I don't think anyone hitting the same problem as you would figure out that this plugin will fix it.

othalan:
@lazka, restarting QL is an option.  However, doing so was enough of a nuisance that I took the time to create this plugin to work around the problem.  The two annoyances with restarting QL are first, the extra time involved waiting for shutdown and restart, and second that I use virtual desktops to keep my apps organized and must ensure QL opens where I want it to.  Minor, I admit, but annoying.

In truth, I'm not certain I'd include this plugin in your position, especially if the VLC backend is merged in. But I wanted to offer it for the sake of it being available and referenced in the issue.  It occurs to me, this plugin would be more appropriate in a list of available plugins that can be downloaded separately.

Out of curiosity, why not include a "stop" button in the GUI?  That provides a workaround most users discover on their own.

bassstorm:
I'd love to have Stop button. Because, sometimes I really want to... actually **stop** playback.

rogatty:
Can this PR be merged? I just hit the issue with bluetooth headphones and would love to have this workaround available.

pklampros:
The added benefit of this plugin is that with it, QL actually allows bluetooth headphones to go to sleep if left on and unused because it lets go of the sound interface.

declension:
Cleaning up old PRs.

Seems like a good workaround to several users' audio problems. Perhaps we can fix the conflicts, rebase / merge from master and try it out (CC @lazka)?

bassstorm:
Would be really nice to have. I already experienced that unpleasant audio output change behavior on macOS :smiley:

othalan:
> Cleaning up old PRs.
> 
> Seems like a good workaround to several users' audio problems. Perhaps we can fix the conflicts, rebase / merge from master and try it out (CC @lazka)?

I am not currently using QuodLibet, however I realize the utility of this plugin as it was critical to my own use when I did use QuodLibet as my music app. I have  done my best to rebase the plugin and fix problems so that it can be merged.

declension:
Thank you @othalan !

PR #2854, opened 2018-05-19
Add releases to appdata

fkrull:
[Another thing the Flathub requirements explicitly point out is releases](https://github.com/flathub/flathub/wiki/App-Requirements#appstream); probably because they're used for the version displayed on the website. So this is two more places that have to be updated on a release, which isn't ideal, but I didn't see any sort of maintainer script. The appdata releases could be generated from NEWS, I suppose, possibly including changelog entries.

I also added some other fields and fixed a few things `appstream-util validate-strict` was complaining about.

Semi-related: how does the release process work? Since there's now a separate release branch, would these changes not be in there or do you cherry-pick selectively? I'd love to have QL on Flathub for 4.1 but that would require these changes (probably).

### DIFF
--- a/quodlibet/data/exfalso.appdata.xml.in
+++ b/quodlibet/data/exfalso.appdata.xml.in
@@ -1,11 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2013 Christoph Reiter -->
 <application>
-    <id type="desktop">exfalso.desktop</id>
+    <id type="desktop-application">exfalso.desktop</id>
     <name>Ex Falso</name>
     <_summary>Edit tags in your audio files</_summary>
     <metadata_license>CC0</metadata_license>
     <project_license>GPL-2.0</project_license>
     <translation type="gettext">quodlibet</translation>
+    <launchable type="desktop-id">exfalso.desktop</launchable>
+    <provides>
+        <binary>exfalso</binary>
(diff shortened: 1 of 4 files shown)


lazka:
> Another thing the Flathub requirements explicitly point out is releases; probably because they're used for the version displayed on the website. So this is two more places that have to be updated on a release, which isn't ideal, but I didn't see any sort of maintainer script. The appdata releases could be generated from NEWS, I suppose, possibly including changelog entries.

ok. there is a script (https://github.com/quodlibet/quodlibet/tree/master/release_db) which generates the info for the download page and the update checker, but those are meant to be used after the release. I'll just add this to the release todo list.

> Semi-related: how does the release process work? Since there's now a separate release branch, would these changes not be in there or do you cherry-pick selectively? I'd love to have QL on Flathub for 4.1 but that would require these changes (probably).

yes, I'll just cherrypick everything. It's a bit more than expected but easy to do :)

https://quodlibet.readthedocs.io/en/latest/development/maint.html#tags-branches
----

re the test failure of appstream-util: I guess the one in 16.04 is too old.

lazka:
> re the test failure of appstream-util: I guess the one in 16.04 is too old.

suggested fix: https://bpaste.net/show/584961bcb3e2

fkrull:
There we go, now it skips properly. That's what I get for trying to be fancy.

PR #2857, opened 2018-05-20
Update Polish translation for 4.1

piotrdrag:
Please also cherry-pick to master.

PR #2860, opened 2018-05-21
Update German translation

Mellthas:
Testing results in the following errors, although the translation seems to be okay:

    de.po:6099: a format specification for argument 'title', as in 'msgstr[0]', doesn't exist in 'msgid_plural'
    de.po:6146: number of format specifications in 'msgid_plural' and 'msgstr[0]' does not match

These seem to be the same as in #2857.

Additionally, when running Quod Libet in German, `%d` is not replaced with the actual number in the translation for _Show %d Files in File Manager_ (_%d Dateien in der Dateiverwaltung anzeigen_). See the attached screenshot. Might be related?

translation-bug

### DIFF
--- a/quodlibet/po/de.po
+++ b/quodlibet/po/de.po
@@ -7,10 +7,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Quod Libet 4.0\n"
+"Project-Id-Version: Quod Libet 4.1\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


lazka:
I've pushed some (hopefully) fixes. Please try to rebase.

I'll add some tests so this doesn't happen again.

Mellthas:
Did a rebase, works fine now. Thanks!

lazka:
thanks

Issue #2861, opened 2018-05-22
Quick access to composer's Wikipedia page


archy3:
We already have quick access plugins to artist's wiki and album's wiki and it would be nice to have the same thing for composers. Sure we can use the Website Search plugin, but one more submenu to navigate through for a feature like this can be a little annoying.

I can think of some other tags that perhaps deserve this, but YAGNI.

PR #2862, opened 2018-05-22
Add composer to Wikipedia plugin

archy3:
This address #2861.


frestr:
Hmm, it is perhaps a bit redundant to add a new plugin for this, especially when we already have the website search plugin which largely replaces the wikipedia plugin.

An alternative can be to add a submenu with tags to be searched, which can then be modified through the (wikipedia) plugin's preferences.

Just my two cents. I guess this can work for now though.

lazka:
> An alternative can be to add a submenu with tags to be searched, which can then be modified through the (wikipedia) plugin's preferences.

Agreed

> Just my two cents. I guess this can work for now though.

yeah

Issue #2856, opened 2018-05-20
Permanent Queue


Treferwynd:
I'd like the queue to behave like in most other media players, so that it doesn't consume songs and you can repeat/shuffle/etc on it. Like an actual playlist with a reserved place in the UI.  I know this is specifically against the idea of queue in QL, so feel free to just tell me "NO".

Reasons:
* search the library without affecting what I'm listening to
* listen to different songs from different albums/artists/etc (on loop/shuffle/whatever) without creating a playlist

This could probably be done with a plugin, but the queue's behaviour seems hardcoded, which is a bummer.

It'd be cool to change the queue's behaviour via settings or plugins. It really is the only thing that keeps me from having the perfect music player!

frestr:
This has been a wanted feature for some time now. See #1890 for a discussion.

Treferwynd:
I'm not sure it's the same thing, but if one could override the queue behaviour it'd be trivial to implement #1890

frestr:
It's unfortunately not that simple because of the way the queue is implemented and how it interacts with the other components in the UI.

I've begun working on a possible solution though, so we may see some movement on this.

Treferwynd:
I've tinkered a bit with the code, and while admittingly getting a bit lost, I thought a nice solution could be a `PlayOrder: Queue` plugin. This way with a couple of plugins  (one for the queue and one for the "normal" playing) one could make QL behave exactly as they want.

frestr:
Do you mean being able to use the play order plugins on the queue while still keeping the queue's play order separate from the main song list? That's a good suggestion, but is perhaps a separate issue from having a permanent queue (which should be possible now with #2865).

Correct me if I'm misunderstanding something though.

Treferwynd:
Exactly. To be more specific: define methods like `next`, `next_implicit`, `set`, `set_implicit`, etc. just for the queue.
So for example to simulate how things work right now you'd have the main song list's `next` which checks if the queue is empty, end if not it calls `next` on the queue. Or maybe use an event plugin? So when a song ends `plugin_on_song_ended` calls `next` on the queue if it's not empty, and `next` on the main song list otherwise.

(In the next days I'll give a try to your PR!)

frestr:
@Treferwynd did you try the PR / Keep songs feature? I think it would be clearer to have a separate issue about the play order (somewhat related: #1411), so feel free to make that. I'll close this for now.

Treferwynd:
[...]

Issue #1890, opened 2016-03-12
Make the queue more like a timeline, with recently played songs visible


CreamyCookie:
There are some actions that I mainly do after I've listened to a song. Most of the time those are rating a song or adding it to a playlist. And because the songs are coming from all over the place, the only option AFAIK to find the last played songs, is to run the default search "Recently played".

I don't know how common that is, but having a menu that contains the names of the last 10 songs or so with their context menus would save a lot of time. Although one could open another browser with the last played songs, this would require much more space (and the window doesn't minimize and maximize with the main window).

declension:
Yep, I would like this too. However: #152 (which I'm tempted to have another go out actually, it's semi-blocking a _lot_ of features - or worse, increasing numbers of UI hacks necessary)

FWIW this is almost a mirror of #894. For me, an ideal (if slightly unusual) UI would be a timeline that extends into the past and future

CreamyCookie:
Huh.. how is this almost a mirror of #894 ? Isn't party shuffle about future (to be played) songs, while this is about past (already played) songs.

declension:
Yes exactly that @CreamyCookie (mirror as in "inverse") - and I think the simplest UIs, that a few other players use, show both past and future songs without needing to switch context or look elsewhere too much.

In QL I'm imagining this as a rework / enhancements of the queue, but there are other solutions too no doubt.

CreamyCookie:
Ahh, that makes sense. So your idea is that the queue also shows the previously played songs, if you scroll up. So, based on the scrolling more songs would be displayed. So it would like this, right?

(The first two elements would be slightly greyed out.)

- Everything Sucks — Descendents
- Forever Young — Alphaville
- ▶️ **Augenblick — Farin Urlaub**
- Technologic — Daft Punk
- Stressed Out — Twenty One Pilots
- [Harder, Better, Faster, Do it — Daft Labeouf](https://www.youtube.com/watch?v=hAEQvlaZgKY)
- Chandelier — Sia




That sounds like a good idea! :+1: 

It would also have the nice benefit of making it easy to re-add a song that you accidentally skipped, for example, by clicking on another song.

urielz:
Yes. I still haven't get used to the queue in QL because of this. Being able to scroll forward and back in time would be a great feature to add.

e-anima:
as far as i understand that, that mean to not remove items from the queue after they have been played.

TheSachin:
[...]

PR #2865, opened 2018-05-23
Make queue able to keep songs after playing them

frestr:
This PR adds a new checkbox to the queue preferences - "Keep Songs".

When enabled, the queue will get a "CurrentColumn" and act more like the main song list. Songs can be added to the queue as normal, but they will be kept in the queue when playing them.

When "Keep Songs" is enabled, it's also possible to enable the "Prioritize Queue" checkbox. If the user selects a song in the main song list with this option enabled, then the player will switch to the queue after the song is done playing. This is the same as how the queue works now.

If "Prioritize Queue" is disabled, then the player will not switch to queue automatically (but rather play the next song in the main song list), and the user will have to click on a song in the queue to begin playing from it again. This makes it possible to occasionally ignore the queue without having to clear it.

It is quite rough around the edges, and some feedback would be very appreciated.

Some things to consider:
- There may be a better wording for the feature than "Keep Songs".
- The "Prioritize Queue" checkbox is not very pretty wrt. how it's disabled when "Keep Songs" is not checked. It's not very clear that the two are connected.
- It could perhaps be useful to have the queue remember its location when switching to the main song list, such that the queue automatically continues where it left off when the main song list is done.

Related issues: #1890, #2856.

### DIFF
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -75,6 +75,7 @@ INITIAL = {
         "queue_expanded": "false",
         "shufflequeue": "false",
         "queue_stop_at_end": "false",
+        "queue_keep_songs": "false",
         "queue_prioritize": "true",
 
         # <reversed?>tagname, song list sort
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -76,7 +76,7 @@ INITIAL = {
         "shufflequeue": "false",
         "queue_stop_at_end": "false",
         "queue_keep_songs": "false",
-        "queue_prioritize": "true",
+        "queue_ignore": "false",
 
         # <reversed?>tagname, song list sort
         "sortby": "0album",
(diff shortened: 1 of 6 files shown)


lazka:
Works nicely from quick test :)

> There may be a better wording for the feature than "Keep Songs".

Nothing comes to mind, except "Remove Finished Songs" which is on by default, but I'm not sure that's better.

> The "Prioritize Queue" checkbox is not very pretty wrt. how it's disabled when "Keep Songs" is not checked. It's not very clear that the two are connected.

We could have a "pause" button on the queue which makes things just ignore it as if it was empty in all cases.

---

feel free to merge this and leave potential improvements for later

(I wont be cherry picking this to 4.1)

frestr:
Thanks for the feedback!

> We could have a "pause" button on the queue which makes things just ignore it as if it was empty in all cases.

I like that idea. Would be useful for the original queue behavior too.

> feel free to merge this and leave potential improvements for later

All right. I'll finish it up and try to add some tests too.

Issue #2595, opened 2017-10-18
Can I Manually Order Songs in a Playlist?


hmc-cs-bcastro:
Can someone tell me how to manually order songs in a playlist? I'd like to sort songs in a specific order, but I can only find how to sort songs by certain tags. I suppose I could add in a numerical tag, but it would be convenient if I did not have to update the numbers anytime I wanted to add a song to the middle of the playlist.

declension:
You can drag-and-drop songs to reorder them

hmc-cs-bcastro:
Should I be able to reorder songs within a playlist? I found I can add songs in a certain order to a playlist. But if I want to move a song already in there up or down, it will not change for me;  but if I delete the song and add it back in, then I can put it wherever I want.

declension:
Hmm, I can't reproduce this

hmc-cs-bcastro:
Oh? Well, I do have a way around that to get an order I want, in any case. Thank you for checking.

aMoniker:
I'm seeing the same issue. I cannot re-order songs within a playlist via drag & drop.

othalan:
I too have this same problem.  I cannot manually reorder playlist songs via drag & drop.  When I drag a song I get highlights indicating it will work but nothing happens when I drop the song in the new location.

OS: MacOS High Sierra
QuodLibet: 4.0.2

declension:
@aMoniker, @hmc-cs-bcastro can you confirm what OS you're on? This is sounding like a MacOS GTK+ issue...

aMoniker:
@declension could be - I'm on MacOS 10.13.3

urielz:
I can reproduce this as well - MacOS.

urielz:
quick comment for those mac users out there, a cumbersome workaround is to manually edit the playlist file. All playlists files are in ASCII and can be found in: ~/.quodlibet/playlists/

othalan:
I have begun looking into this issue, but will be away from my computer for 3 days.  I have not made any progress, but wanted to detail what I see looking at the code for my own reference, and thought here would be the best place in case someone has input or I do not get back to this issue.

- All drag and drop events appear to occur normally (though I don't have a working system to compare to ... setup a Linux virtual env to compare to?).
- In `songlist.py`, the actual move command is `model.move_after(...)`.  Examination of variables seems to indicate good looking values though not everything has been fully examined.  
- First thought is that the location to begin look for a problem is in the `move_after` method or its arguments.
[...]

PR #2867, opened 2018-05-25
Bug Fix: Cannot re-order playlist songs with DnD 

othalan:
This is a bug fix for Issue #2595, "Cannot re-order playlist songs with DnD".  See comments in the code and the issue for full details.

This fix should work on all operating systems, however it has so far only been tested on MacOS as I lack an ability to test on Linux and Windows at this time.


lazka:
Thanks, I'll test this on Linux first :)

PR #2876, opened 2018-06-06
Port Inhibit Screensaver extension to GDBus

qarkai:
According to https://www.freedesktop.org/wiki/Software/DBusBindings/  dbus-python is obsolete. This patch starts porting QL to recommended GDBus binding.


lazka:
Please use `GLib.Error` instead of `Exception` for error handling

lazka:
Thanks!

PR #2878, opened 2018-06-07
Add scootergrisen (da) as translator

scootergrisen:


lazka:
Thanks, I need to automate that somehow..

Issue #2111, opened 2016-11-21
Add spinner during slow UI loading


declension:
Some of the Browsers take a long time to render initially, during which time the user might think something has crashed, and even the window manager can think the process has hung (often indicated by fading the window to monochrome).

Obviously, ideally these operations would always be fast, but for now it would be good if we could feed back to the user that some processing is happening in order to restore confidence in the UI.

The spinner paradigm usually works well, but we could also perhaps use the Task pulse (bottom right), or suggestions welcome...

lazka:
This has recently also come up in #2100

lazka:
A good start would be to split the browser change into three steps:

1) load browser, set insensitive
2) set library, fill widgets <- UI freeze
3) set sensitive

PR #2880, opened 2018-06-08
Show spinner during browser switch (and set the cursor to a watch)

frestr:
Related: #2111 

This makes it clear that the browser is loading when switching to browsers that take some time to load.

The spinner doesn't really spin though, since the gtk main loop is not run during the browser loading. I don't think this is so easy to fix, as the browser loading probably (?) can't be put in a separate thread. I don't consider this to be much of an issue though.

The waiting for the main loop also makes the set-browser test in `test_commands.py` generate a couple of these warnings:

```
 /usr/lib/python3.6/site-packages/gi/overrides/GObject.py:472: Warning: gsignal.c:2563: instance '0x563052220530' has no handler with id '3595'
```

Not sure why.

### DIFF
--- a/quodlibet/quodlibet/qltk/quodlibetwindow.py
+++ b/quodlibet/quodlibet/qltk/quodlibetwindow.py
@@ -1195,6 +1195,19 @@ class QuodLibetWindow(Window, PersistentWindowMixin, AppWindow):
                 self.remove_accel_group(self.browser.accelerators)
             container.destroy()
             self.browser.destroy()
+
+        spinner = Gtk.Spinner()
+        spinner.start()
+        spinner.show()
+        self.__browserbox.add(spinner)
+        window = self.get_window()
+        window.set_cursor(Gdk.Cursor.new(Gdk.CursorType.WATCH))
+
+        # Wait for the old browser to be destroyed before
+        # setting up the new one
+        while Gtk.events_pending():
+            Gtk.main_iteration()
+
         self.browser = Browser(library)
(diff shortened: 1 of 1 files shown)


lazka:
I'm not a fan of the flickering even in cases where the switching time is small (small library, or simple browser). I do like the cursor change, but resetting it on each draw might conflict with other things and I'm not sure how performant it is in general. A simple one-time idle_add() should do as well there.

frestr:
> I'm not a fan of the flickering even in cases where the switching time is small (small library, or simple browser).

Do you think it would be better to first set up the empty browser, and then fill the panes? There will still be some flicker, but probably not as bad. (On a second glance, this seems like what you suggested in #2111.) I tried doing something like that at first, but the browsers don't seem to have a common way of filling the panes, so I think that would require some changes to each browser.

> I do like the cursor change, but resetting it on each draw might conflict with other things and I'm not sure how performant it is in general. A simple one-time idle_add() should do as well there.

Good point, didn't think of that.

frestr:
> Do you think it would be better to first set up the empty browser, and then fill the panes?

I did some further experimentation with this, and I'm not really convinced that it would be any better than the current implementation (which is updating the browser when everything has loaded). 

Since I also suspect fully implementing this would be a bit of work, I suggest we just add the cursor change (which by itself makes the switch look quite a bit better imo, as the player doesn't appear to freeze then) and leave the rest be for now.

lazka:
Sounds good, thanks for the effort :)

PR #2884, opened 2018-06-11
German translation: Fix two untranslated strings

Mellthas:
Commit 84bb8242cc75f6b2bbda455afbf7964c559c2eca changed two strings after the last update to the translation. Would be nice to have the translation complete again in case you do a 4.1.1 release :-).


lazka:
Thanks, also pushed to master

PR #2885, opened 2018-06-14
Port util.environment module to GDBus

qarkai:
According to https://www.freedesktop.org/wiki/Software/DBusBindings/ dbus-python is obsolete. This patch ports another QL module to recommended GDBus binding.

### DIFF
--- a/quodlibet/quodlibet/util/environment.py
+++ b/quodlibet/quodlibet/util/environment.py
@@ -22,14 +22,20 @@ def _dbus_name_owned(name):
         return False
 
     try:
-        import dbus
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/util/environment.py
```diff
@@ -22,14 +22,20 @@ def _dbus_name_owned(name):
         return False
 
     try:
-        import dbus
+        from gi.repository import Gio
+        from gi.repository import GLib
     except ImportError:
```
lazka: remove the try/except, we depend on gio/glib on all platforms

PR #2903, opened 2018-07-11
Port Telepathy Status Messages extension to GDBus

qarkai:
According to https://www.freedesktop.org/wiki/Software/DBusBindings/ dbus-python is obsolete. This patch ports Telepathy Status Messages extension to recommended GDBus binding.


lazka:
The CI error is unrelated and due to MSYS2 moving to Python 3.7, I'm working on it..

Issue #442, opened 2015-03-14
The Play Queue should use the Play Order setting


lazka:
Original [issue 442](https://code.google.com/p/quodlibet/issues/detail?id=442) created by DaveParillo on 2010-02-21T18:46:34.000Z:

The Play Queue has it's own 'random' check box which takes precedence over
the Play Order drop down list at the bottom of the player window.

This is not only redundant, but confusing because the queue's random check
box is only visible when the queue is expanded, while the Play Order is
always visible.

Also, (I think) the random check box means that you can't control the play
order from the tray icon.

PR #2911, opened 2018-07-15
Repeat and Shuffle also affect the Queue

Treferwynd:
Like the main song list does.

If Shuffle is enabled QL consumes the queue in
random order.

If Repeat and Keep Songs are enabled, the queue will loop.If Repeat is enabled, but Keep Songs is not, the queue is consumed as usual (or randomly if Shuffle is enabled).

If Ignore is checked, the queue is ignored so it's order doesn't really matter.

I don't think there is need to introduce any other UI element, I think this is a pretty good and intuitive behaviour. Let me know if I've missed some scenarios!

This PR should be the last step to have the queue behave like the "Now Playing" section of other media players.

There are 2 issues though:
* the Queue Only plugin may not work as intended
* there is also the Random checkbox in the queue's settings

Related issues: #2856 and #442 a bit

### DIFF
--- a/quodlibet/quodlibet/qltk/playorder.py
+++ b/quodlibet/quodlibet/qltk/playorder.py
@@ -241,11 +241,12 @@ class PlayOrderWidget(Gtk.HBox):
         'changed': (GObject.SignalFlags.RUN_LAST, None, ()),
     }
 
-    def __init__(self, model, player):
+    def __init__(self, model, queue_model, player):
         super(PlayOrderWidget, self).__init__(spacing=6)
         self.order = None
         self.__inhibit = True
         self.__playlist = model
+        self.__queue = queue_model
         self.__player = player
 
         def create_shuffle(orders):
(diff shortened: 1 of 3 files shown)


declension:
I like the unity in this idea - but be aware some users (myself included) have become accustomed to shuffle within the browser, but a strictly ordered queue: this allows you to "occasionally play an album start to finish" 


> there is also the Random checkbox in the queue's settings

Well, if there is unification of the shuffle modes, and thus this button is ignored, then this checkbox has to go (but I haven't tried the branch yet)

It actually then allows a long-term goal which is better shuffle modes (e.g. weighted) for the queue too...

Treferwynd:
> some users (myself included) have become accustomed to shuffle within the browser, but a strictly ordered queue

There are two solutions I can think of:
*  replace the Random checkbox with something like "Shuffle & Repeat" with help text "Shuffle and Repeat modes also affect the Queue" (checked on by default, as IMHO it's the most natural behaviour)
* manually disable Shuffle (which admittedly is not exactly a solution)

Treferwynd:
I can't see how to conciliate @declension use case with this behaviour, so I'm closing this PR.

Treferwynd:
The only problem with that is that we lose the possibility of ordered playback in the browser and random queue.

I'll see what I can do!

frestr:
> The only problem with that is that we lose the possibility of ordered playback in the browser and random queue.

One would need to turn shuffle on/off when changing between the two instead. It's a slight efficiency/simplicity trade-off, but overall I would say it increases the usability.

Issue #2846, opened 2018-05-16
App menu doesn't work on Wayland


fkrull:
The application menu (the one you get when you click on the window title in gnome-shell) doesn't work on Wayland:

![app-menu-wayland](https://user-images.githubusercontent.com/2378306/40128882-210c89e6-5933-11e8-9a83-0af86885b9e6.png)

This was with QL 4.0.2 on Ubuntu 18.04 with Wayland.

[The code piece](https://github.com/quodlibet/quodlibet/blob/a0da55222397d7b759451d9660a74347fe3203b1/quodlibet/quodlibet/qltk/quodlibetwindow.py#L478) that registers the app menu only works on X11. The "proper" way to fix this would be to use GtkApplication and its [set_app_menu method](https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-set-app-menu), but I suspect that requires some in-depth changes.

(This one isn't a Flatpak issue, I checked.)

lazka:
Yeah, thanks.

Valeryan24:
Hi, is it a similar problem (I don't want to create a duplicate issue) that makes Quod Libet icon not visible in Ubuntu dock with Wayland session on 18.04 / 18.10 (bottom of the screenshot) ?
https://framapic.org/U5PfilFgyGtS/0eNA4cqulf1Y.png
Thanks for your great work !

lazka:
@Valeryan24 how did you install QL?

Valeryan24:
First with the default Ubuntu package (I use development version 18.10 on Wayland), then with your unstable ppa: both give the same result for dock and top bar icon.

lazka:
> First with the default Ubuntu package (I use development version 18.10 on Wayland), then with your unstable ppa: both give the same result for dock and top bar icon.

Thanks, that's a bug then.  Might be that us not using GtkApplication is the same cause, but please open a new issue for this. Thanks!

lazka:
Gnome is moving away from app menus now, the we can probably just get rid of it altogether.

fkrull:
Wait, they are? Bloody hell, can't they make up their minds.

lazka:
https://wiki.gnome.org/Design/Whiteboards/AppMenuMigration

I wont miss them at least.

lazka:
It's a bit more official now: https://mail.gnome.org/archives/desktop-devel-list/2018-June/msg00021.html

lazka:
[...]

PR #2915, opened 2018-07-19
Display cover art image with the correct size on multi-monitor setups

frestr:
When clicking on the cover art of the playing song, the image is enlarged and fills approximately half of the width and height of the monitor. This seems to be the intended behavior.

On multi-monitor setups, the image does not necessarily enlarge in the same way. I have multiple monitors with varying sizes and orientations, and the image always fills to the edge of the current monitor. The reason for this is that the dimensions of `Gdk.Screen` (which combines all monitors into one big (virtual) one) are used when scaling the image.

A better way is to use the current monitor as a base when calculating the desired width and height of the image, which is what this commit does.

Leaving the patch here in case someone wants to have a say in this.

### DIFF
--- a/quodlibet/quodlibet/qltk/cover.py
+++ b/quodlibet/quodlibet/qltk/cover.py
@@ -42,11 +42,18 @@ class BigCenteredImage(qltk.Window):
             width = int(width / 1.1)
             height = int(height / 1.1)
         else:
-            win = Gdk.Window.at_pointer()
-            disp = Gdk.Display.get_default()
-            if win[0] and disp:
-                mon = disp.get_monitor_at_window(win[0])
-                rect = mon.get_workarea()
+            win = parent.get_window()
+            if win:
+                if qltk.gtk_version[:2] >= (3, 22):
+                    disp = Gdk.Display.get_default()
+                    mon = disp.get_monitor_at_window(win)
+                    rect = mon.get_geometry()
+                else:
+                    # The result should be the same as above, just using
+                    # deprecated methods instead
+                    screen = Gdk.Screen.get_default()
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/qltk/cover.py
```diff
@@ -42,8 +42,16 @@ def __init__(self, title, fileobj, parent):
             width = int(width / 1.1)
             height = int(height / 1.1)
         else:
-            width = int(Gdk.Screen.width() / 1.8)
-            height = int(Gdk.Screen.height() / 1.8)
+            win = Gdk.Window.at_pointer()
+            disp = Gdk.Display.get_default()
+            if win[0] and disp:
+                mon = disp.get_monitor_at_window(win[0])
```
lazka: this API is gtk 3.22+

lazka:
Any idea re the test failures?

quodlibet/quodlibet/qltk/cover.py
frestr: Ah, right. I'll make it compatible with older versions too.

frestr:
It seems like the docker container running the tests is being SIGTERM'ed (exit code 143). I initially though it may have had something to do with the CircleCI outage, but it's strange that it's still happening if that's the case.

lazka:
Yeah,  had the same problem/error with #2846 and I had to revert. Maybe something xvfb related.

frestr:
Argh, still failing. I'll try to investigate.

frestr:
Ok, so that problem was simply that my fork's master was updated right after the commit that made the tests fail. Fetching upstream and rebasing fixed it.

PR #2922, opened 2018-07-25
Random album playback plugin - make it Python 3 compatible

sphh:
When using the `Random album playback` plug-in on Linux Mint 19 with Python 3.6.5 I always get exception when QL switches the album. This was due to the line 193 in `quodlibet/ext/events/randomalbum.py`:
```python
                album_scores = sorted(self._score(chosen_albums))
```
The exception was cause by Python 3 not being able to sort class instances without a `__le__` method.

The original code sorted the albums and picked the maximum. I changed the code to find the maximum score, filter the albums by this maximum score and pick an album by random from the resulting list. This also solved the problem, that the original `randomalbum.py` always picked the last album (in alphabetical order), if more than one album has the maximum score.

### DIFF
--- a/quodlibet/quodlibet/ext/events/randomalbum.py
+++ b/quodlibet/quodlibet/ext/events/randomalbum.py
@@ -188,23 +188,27 @@ class RandomAlbum(EventPlugin):
                 # Select 3% of albums, or at least 3 albums
                 total = len(values)
                 nr_albums = min(total, max(int(0.03 * total), 3))
-                print_d("Choosing from %d library albums. Best:" % nr_albums)
+                print_d("Choosing from %d library albums:" % nr_albums)
                 chosen_albums = random.sample(values, nr_albums)
                 album_scores = self._score(chosen_albums)
-                for score, album in album_scores[-5:]:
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/ext/events/randomalbum.py
```diff
@@ -190,10 +190,17 @@ def plugin_on_song_started(self, song):
                 nr_albums = min(total, max(int(0.03 * total), 3))
                 print_d("Choosing from %d library albums. Best:" % nr_albums)
                 chosen_albums = random.sample(values, nr_albums)
-                album_scores = sorted(self._score(chosen_albums))
+                album_scores = self._score(chosen_albums)
                 for score, album in album_scores[-5:]:
                     print_d("%0.1f scored by %s" % (score, album("album")))
```
frestr: This print should probably be removed, since without the sort it just prints five random albums as opposed to the highest rated ones (the print above would also need to be modified a little, removing the 'Best'). Perhaps print the chosen album instead.

frestr:
Thanks. You'll need to fix the failing tests though.

PR #2930, opened 2018-08-05
Scrolling the mouse wheel seeks waveform seekbar

blimmo:
Similar functionality to that available with the alternative seekbar.

I considered making the seek amount setting in seconds instead of milliseconds because I couldn't think of a use case for millisecond accuracy and I felt it would be clearer. I settled on milliseconds since that is the precision it can offer. If people think seconds would be better I can switch to that fairly easily.

~~The workaround for the double calls of the scroll callback was the best I could come up with. Getting my callback called with 2 different numbers of arguments baffled me. Is this a bug with PyGObject?~~

### DIFF
--- a/quodlibet/quodlibet/ext/events/waveformseekbar.py
+++ b/quodlibet/quodlibet/ext/events/waveformseekbar.py
@@ -670,6 +670,7 @@ class WaveformSeekBarPlugin(EventPlugin):
             adjustment=Gtk.Adjustment(CONFIG.seek_amount,
                                       0, 60000, 1000, 1000, 0)
         )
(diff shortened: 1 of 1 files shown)


sjohannes:
> double calls of the scroll callback

In PyGObject, you can either override `do_<signal_name>`, or `connect` to the signal; don't do both.

blimmo:
Ah, thanks! I couldn't find very good docs on events in PyGObject so I had to guess somewhat using the GTK reference on events and what the other plugins had done. Should be fixed now.

CreamyCookie:
Great! This is the other thing the waveform seekbar was missing for me :+1:

PR #2936, opened 2018-08-13
Missing tag plugin

### LINKED ISSUE #1940: Add query syntax to match on missing tag values
I'd like to search for all the songs with high ratings that I haven't listened to in a while or which have no recorded "lastplayed" date. The query might look like this

`&(#(rating:max > 0.8), |(#(lastplayed > 6 month ago), #(lastplayed is None)))`

Similarly, searching for songs that haven't been rated yet could be interesting if you want to rate most or all of the songs in your library.

blimmo:
A plugin to match missing tags. Would (kind of) close #1940

### DIFF
--- a/quodlibet/po/POTFILES.in
+++ b/quodlibet/po/POTFILES.in
@@ -95,6 +95,7 @@ quodlibet/ext/playorder/shufflebygrouping.py
 quodlibet/ext/playorder/skip_songs.py
 quodlibet/ext/playorder/track_repeat.py
 quodlibet/ext/query/conditional.py
+quodlibet/ext/query/missing.py
 quodlibet/ext/query/pythonexpression.py
 quodlibet/ext/query/savedsearch.py
 quodlibet/ext/_shared/squeezebox/base.py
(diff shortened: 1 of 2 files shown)


frestr:
Works nicely from a quick test.

It would probably be a good idea to strip the tag though, as queries like `@(missing: date)` and `@(missing:date)` do not yield the same results.

blimmo:
I added a strip to the body as requested.

I'm a little unhappy with the handling of numeric tags. E.g. rating must be entered as `~#rating`. I believe this could be fixed by using the call interface of `data` instead of the dict one but when I do that `@(missing:rating)` matches everything.

Since this is a plugin I think merging it as is will be fine but if there are any ideas to improve this behaviour they'd be welcome.

frestr:
> I believe this could be fixed by using the call interface of data instead of the dict one but when I do that @(missing:rating) matches everything.

You still need to include the `~#` prefix when using the call interface.

Even though `~` and `#` are metacharacters, they are still part of the tag key, so `rating`, `~rating` and `~#rating` are regarded as different tags. I guess it would be possible to introduce some kind of fallback mechanism, but that may introduce ambiguity and perhaps be a bit confusing.

A list of the internal tags (with the prefix included) exists in the [documentation](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html), so I don't consider it an issue having to use the whole tag key.

PR #2937, opened 2018-08-14
Animated display plugin: Add more preset positions

### LINKED ISSUE #2927: Animated On-Screen Display plugin: preset positions
Is it possible to have bottom left and bottom right as preset positions (currently only top, middle or bottom of screen) or even better to be able to drag the popup somehow to set a custom position? The latter is probably much more work and just having the additional presets would be great. This would make it much more convenient to have the popup be on the primary monitor.

From #2920 originally.

frestr:
By replacing the combo box with a grid of radio buttons.

Fixes #2927


lazka:
Looks good!

PR #2945, opened 2018-08-22
Made the ordering of the album list and cover grid browsers deterministic

dpitch40:
All unit tests passing that were passing before.


frestr:
Thanks :)

PR #2953, opened 2018-08-31
Unify Wikipedia plugins

### LINKED ISSUE #2877: Unify Wikipedia plugins
At the moment we have three separate Wikipedia plugins which essentially does the same thing (searching for a tag in Wikipedia), just with different tags.

An alternative to this can be to have a single Wikipedia search plugin which has a submenu with tags to be searched, which can then be modified through the plugin's preferences. This also makes it possible to add new tags without creating a new plugin.

frestr:
Instead of three separate plugins with hardcoded tag keys, have a single
plugin with customizable tags.

Fixes #2877

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/wikipedia.py
+++ b/quodlibet/quodlibet/ext/songsmenu/wikipedia.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2005 Inigo Serna
-#           2018 Phoenix Dailey
+#           2018 Phoenix Dailey, Fredrik Strupe
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/ext/songsmenu/wikipedia.py
```diff
@@ -50,35 +58,61 @@ def PluginPreferences(self, parent):
...
+        vb = Gtk.VBox(spacing=6)
+        vb.pack_start(hb, False, False, 0)
+
+        def _open_editor(widget):
+            def _editor_closed(widget):
+                tags = widget.get_strings()
+                config.setlist("plugins", "wiki_tags", tags)
+
+            tags = config.getlist("plugins", "wiki_tags", self.DEFAULT_TAGS)
+            editor = TagListEditor("Test", [] if tags == [''] else tags)
```
lazka: Test?

frestr: Whoops

Issue #2499, opened 2017-07-23
Gnome Multimedia Keys: Not all actions supported


lazka:
There are a few unhandled actions which we could support: Rewind, FastForward, Repeat, Shuffle

lazka:
The events can be emulated under X11 using xdotool:

```
xdotool key XF86AudioRewind
xdotool key XF86AudioRepeat
xdotool key XF86AudioForward
xdotool key XF86AudioRandomPlay
```

LudoBike:
I'm working on this issue so I'm trying to understand how it works. 

Currently I'm understanding that there is the backends which catch the signal and then it communicates with the MMKeysHandler using MMKeysAction. The keys are associated with action in MMKeysHandler._callback().

In fact my problem is to know what I have to put in GnomeBackend._EVENTS (in gnome.py). For example for the shuffle I don't know if it is `"Shuffle": MMKeysAction.SHUFFLE` or `"RandomPlay": MMKeysAction.SHUFFLE`.

Furthermore as I am on i3wm, my backend isn't GnomeBackend (`D: 0.495: mmkeys.find_active_backend: Found 'KeybinderBackend'` in the log with `./quodlibet.py --debug`) so I can't test my changes. Is there a way to test them without installing a virtual machine ?

lazka:
> In fact my problem is to know what I have to put in GnomeBackend._EVENTS (in gnome.py). For example for the shuffle I don't know if it is "Shuffle": MMKeysAction.SHUFFLE or "RandomPlay": MMKeysAction.SHUFFLE.

It should be "Shuffle": https://github.com/GNOME/gnome-settings-daemon/blob/master/plugins/media-keys/README.media-keys-API

> Furthermore as I am on i3wm, my backend isn't GnomeBackend (D: 0.495: mmkeys.find_active_backend: Found 'KeybinderBackend' in the log with ./quodlibet.py --debug) so I can't test my changes. Is there a way to test them without installing a virtual machine ?

You can also implement it for the KeybinderBackend and test it there. xdotool might work there too.

> And I also need to know what the rewind function is, does it restart the current title ?
> Edit: same problem with forward, I guess it's the opposite of rewind.

[...]

PR #2954, opened 2018-09-01
#2499 − Add the missing support for the Gnome Multimedia Keys

LudoBike:
## Features
These commits implement support of Repeat, Shuffle, Rewind and Forward mmkeys in GnomeBackend and KeybinderBackend according to issue #2499 (if this pull request is accepted the issue could be close).

## Tests
The callbacks are all ok and tested.

Using xdotool (as described in #2499) with KeybinderBackend, RandomPlay and Repeat are not catch.

I don't test the GnomeBackend, if there is any gnome user, feel free to try.


lazka:
Thanks!

PR #2972, opened 2018-09-16
Big Album Cover View Now Updates on Song Change

TheYokai:
The big album cover view used to stay on the same image regardless of what song was currently playing. Now, when the album widget changes due to a new song, it will also change the album art displayed. If there's no album art found, the widget will simply close. 

edit: Some changes should be made to this class in the future, such as allowing the album art to update without creating / destroying the widget. Some groundwork was set for splitting functionality into it's own functions.

### DIFF
--- a/quodlibet/quodlibet/qltk/cover.py
+++ b/quodlibet/quodlibet/qltk/cover.py
@@ -35,12 +35,55 @@ class BigCenteredImage(qltk.Window):
         parent = qltk.get_top_parent(parent)
         self.set_transient_for(parent)
 
+        self.set_position(Gtk.WindowPosition.CENTER_ON_PARENT)
+
+        #If image fails to set, abort construction.
+        if not self.set_image(fileobj, parent):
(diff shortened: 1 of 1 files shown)


TheYokai:
Made some of those necessary corrections. Thanks for the feedback!

frestr:
Nice. Just one last thing that I noticed: `set_image` should either return false (or raise some exception) on failure, such that the rest of the code in `__init__` can be skipped.

As it is now, if the pixbuf fails to load in `set_image` the first time it is run, `self.__image` will not exist, but still tried to be accessed in `__init__` after the call to `set_image`.

TheYokai:
Ok I believe I've made changes to make that more secure. 

I'm not sure on the validity of calling self.destroy inside of the constructor. That's how the code was before I contributed though, so I'm assuming that type of behavior is ok / safe?

frestr:
Yeah, I think that should be okay.

PR #2973, opened 2018-09-18
[[Plugin]] Automatically Update Library Files

WammKD:
I've been looking for this feature in every Linux music player I could possibly find since development on Banshee ceased so I'm super excited about being able to create this for Quod Libet.

I tried choosing Python functions such that this should work on Linux, MacOS, or Windows. The only thing I'm still missing is I couldn't figure out how I might force a rescan of the library (or, if feasible, only the song in question? Though I suppose that'd avoid recognizing that the old location doesn't exist, anymore).

Other than that – though –, this is all set, if it looks alright to you.

_Commits_
* Add in my new plugin; still need to figure out how to rescan the library on file updates.

### DIFF
--- a/quodlibet/quodlibet/ext/events/libraryfileorganizing.py
+++ b/quodlibet/quodlibet/ext/events/libraryfileorganizing.py
@@ -71,11 +71,11 @@ class MyPlugin(EventPlugin, PluginConfigMixin):
                         if(temporary[:temporary.find(")")] == str(num_to_use)):
                             num_to_use += 1
 
-                psbl_new_location = dirname + start_with + \
+                psbl_new_location = dirname + "/" + start_with + \
                                     " (" + str(num_to_use) + ")" + \
                                     file_extension
 
-            shutil.move(current_location, psbl_new_location)
+            shutil.move(current_location, os.path.normpath(psbl_new_location))
 
             self.removeEmptyDirs(os.path.dirname(current_location))
 
--- a/quodlibet/quodlibet/ext/events/libraryfileorganizing.py
+++ b/quodlibet/quodlibet/ext/events/libraryfileorganizing.py
@@ -8,82 +8,82 @@
 
 import os
 import shutil
-from   gi.repository            import Gtk
-from   quodlibet                import _
-from   quodlibet                import util
-from   quodlibet                import qltk
-from   quodlibet.qltk.entry     import UndoEntry
-from   quodlibet.pattern        import Pattern
-from   quodlibet.plugins        import PluginConfigMixin
-from   quodlibet.plugins.events import EventPlugin
-from   quodlibet.util.library   import get_scan_dirs
+from gi.repository import Gtk
+from quodlibet import _
+from quodlibet import util
+from quodlibet import qltk
+from quodlibet.qltk.entry import UndoEntry
+from quodlibet.pattern import Pattern
+from quodlibet.plugins import PluginConfigMixin
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.util.library import get_scan_dirs
 
 class MyPlugin(EventPlugin, PluginConfigMixin):
-	PLUGIN_ID       = "libraryfileorganizing"
-	PLUGIN_NAME     = _("Automatically Update Library Files")
-	PLUGIN_DESC     = _("After making any update (namely, to tags), update "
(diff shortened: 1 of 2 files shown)


WammKD:
Alright, so it's still not passing one test but I think that one might just be a fluke since there's only one test of that test that isn't passing and it passed on the previous commit and the error it's throwing is `Could not connect to sha256.badssl.com: Operation timed out (24)`.

I dunno a way of rerunning it other than submitting yet another commit so I'm just gonna leave this one.

quodlibet/quodlibet/ext/events/libraryfileorganizing.py
```diff
@@ -0,0 +1,119 @@
...
+        psbl_new_location = os.path.abspath(
+            os.path.expanduser((Pattern(self.config_get(self.CFG_PAT_PLAYING,
+                                                        self.DEFAULT_PAT)) %
+                                song) + file_extension))
+
+        if(not(psbl_new_location == re.sub(" \\([0-9]+\\)$",
+                                           "",
+                                           file_split[0]) + file_extension)):
+            dirname = os.path.dirname(psbl_new_location)
+            if(not(os.path.exists(dirname))):
```
frestr: To make our codestyle somewhat consistent, please add a space after keywords like "if" and "not" and only use outer parenthesis on multi-line statements (the linter should have warned about this...). So `if not os.path.exists(dirname):` in this case.

quodlibet/quodlibet/ext/events/libraryfileorganizing.py
```diff
@@ -0,0 +1,119 @@
...
+                for f in sorted(os.listdir(dirname)):
+                    if(f.startswith(start_with)):
+                        temporary = f[length_chk:]
+
+                        if(temporary[:temporary.find(")")] == str(num_to_use)):
+                            num_to_use += 1
+
+                psbl_new_location = dirname + "/" + start_with + \
+                                    " (" + str(num_to_use) + ")" + \
+                                    file_extension
```
frestr: Use `os.path.join` to make this cross-platform

quodlibet/quodlibet/ext/events/libraryfileorganizing.py
```diff
@@ -0,0 +1,119 @@
...
+
+            shutil.move(current_location, os.path.normpath(psbl_new_location))
```
frestr: There should be some exception handling here. While testing the plugin, this line failed a few times because of an unknown current location

quodlibet/quodlibet/ext/events/libraryfileorganizing.py
```diff
@@ -0,0 +1,119 @@
...
+
+            scan_library(app.library, False)
+
+    def PluginPreferences(self, parent):
+        outer_vb = Gtk.VBox(spacing=12)
+        vb = Gtk.VBox(spacing=12)
+
+        # Naming Pattern
+        hb = Gtk.HBox(spacing=6)
+        entry = UndoEntry()
```
frestr: The pattern is a bit awkward to edit since its so long. Consider using a PatternEditBox instance instead.

quodlibet/quodlibet/ext/events/libraryfileorganizing.py
```diff
@@ -0,0 +1,119 @@
...
+from quodlibet.pattern import Pattern
+from quodlibet.plugins import PluginConfigMixin
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.util.library import get_scan_dirs
+from quodlibet.util.library import scan_library
+
+
+class MyPlugin(EventPlugin, PluginConfigMixin):
+    PLUGIN_ID = "libraryfileorganizing"
+    PLUGIN_NAME = _("Automatically Update Library Files")
```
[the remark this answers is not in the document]
CreamyCookie: @frestr I agree. When I saw the pull request, I thought this would change something inside the files (like updating the tags). That said, I think "automatically" (or similar) should still be part of the name, to make clear that this plugin doesn't require manual triggering. Maybe it could be shortened to "Auto-Update File Locations"

declension:
Cleaning up old PRs.

@WammKD (and @frestr ) - any chance this MR can be rebased, and review items fixed / mitigated? 

If not, I'll close this PR. Thanks!

Issue #2818, opened 2018-03-27
Consider pipenv for project dependencies and environment


declension:
[Pipenv](https://docs.pipenv.org/) is a great tool sitting on top of Pip, Virtualenv.. and is py.test compatible. Now that we're PY3 only, it perhaps makes more sense than Tox (#2317).

### Pros
 * Isolated (from OS installs), easily set up Python dependencies
 * Lock dependencies at known versions (and uses hashing) 
 * Thus - makes our Docker / CI steps simpler, and makes these builds more reproducible locally
 * Easy to switch Python versions (and: to mandate them)
 * Some integration of linting tools etc

### Cons
 * More tooling (though after converting several small projects - it's pretty good)
 * Still doesn't deal with system dependencies e.g. Gobject et
 * A few unknowns (for me) about bigger projects with more committers.

lazka:
Not sure about docker -  we want to test against system dependencies which users are actually using.

lazka:
+1 for adding a Pipfile

declension:
> Not sure about docker - we want to test against system dependencies which users are actually using.

Good point, I guess we should leave it (in fact, use it as an "insurance policy" really)

PR #2977, opened 2018-09-24
Add a Pipfile. See #2818

lazka:
```
pipenv install --dev
pipenv shell
./quodlibet.py
```

see #2818


declension:
Oh right, we've already had that exact conversation 😆

PR #2980, opened 2018-09-25
Remove @github from end of name

scootergrisen:


lazka:
OK :) my plan was at some point to parse the list and make the names links in the about dialog, but I never finished that.

Issue #2389, opened 2017-05-18
Enhancement suggestion: Store Eq settings


alan12345678:
Using the equalizer available in the QL plugins, it seems I cannot store discrete eq settings but simply use "custom" to store a single current configuration.

Would it be possible to enable "save as" for eq settings, particularly as this makes back to back testing of sound/song simple, but also because it's very useful to have eq settings that can be restored after an upgrade or reinstall.

ohel:
I just implemented the feature, see PR. Below is a screenshot how it looks. I did some minor UI tweaks too, should be a bit more user friendly now.

![eq](https://user-images.githubusercontent.com/10726032/46985114-51a0f980-d0f1-11e8-899a-be2158bf1157.png)

ohel:
Using Quod Libet default functionality, I don't think so. It might be a worthwile enhancement though to be able to pass parameters to plugins on startup. Don't know if there's a request for it already, or any other use cases yet.

However, what you want is possible using the standard command line tools. Try this:

**`ql_with_eq.sh`:**
```sh
#!/bin/sh
if [ ! $1 ]; then exit 1; fi

qlexe=/usr/bin/quodlibet.py
conf=~/.quodlibet/config

levels=$(grep -o "equalizer_levels = .*" $conf | grep -o "'$1': \[[-0-9 .,]*\]" | cut -f 2- -d ' ')
sed -i "s/\(equalizer_levels = .*'Current': \)\(\[[-0-9 .,]*\]\)\(.*\)/\1$levels\3/" $conf

[ -x $qlexe ] && $qlexe &
```
Modify the qlexe variable to match your setup and give the script execute rights (`chmod +x`).

Usage: `ql_with_eq.sh MyEq` where MyEq is the name of a preset you'd like to use as current preset when starting up Quod Libet.

If you want a desktop file to go with it:
**`quod_libet_with_myeq.desktop`:**
```
[Desktop Entry]
Name=Quod Libet (EQ: MyEq)
Exec=/path/to/ql_with_eq.sh MyEq
Terminal=false
Type=Application
Icon=quodlibet.svg
Categories=AudioVideo;Audio;Player;GTK
```

PR #2995, opened 2018-10-16
Save custom eq presets

ohel:
Implements #2389 (saving and loading of custom EQ presets) and makes the UI a bit more user friendly. Should be backwards compatible with old config files (converts them when loading).

### DIFF
--- a/quodlibet/quodlibet/ext/events/equalizer.py
+++ b/quodlibet/quodlibet/ext/events/equalizer.py
@@ -87,7 +87,7 @@ def interp_bands(src_band, target_band, src_gain):
 
 def get_config():
     try:
-        config_str = config.get('plugins', 'equalizer_levels')
+        config_str = config.get("plugins", "equalizer_levels", "[]")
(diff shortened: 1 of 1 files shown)


quodlibet/quodlibet/ext/events/equalizer.py
```diff
@@ -86,10 +87,25 @@ def interp_bands(src_band, target_band, src_gain):
 
 def get_config():
     try:
-        eq_levels_str = config.get('plugins', 'equalizer_levels')
-        return [float(s) for s in eq_levels_str.split(',')]
-    except (config.Error, ValueError):
-        return []
+        config_str = config.get('plugins', 'equalizer_levels')
```
frestr: You can add a default value as the third argument, to prevent the `except` below from printing an error message the first time the plugin is used.

ohel: Thanks, fixed. I added the disable buttons UI tweaks also.

Issue #2653, opened 2017-12-13
bug in enqueue cli command


pfps:
As far as I can tell the enqueue document consists of

–enqueue filename|query
    Enqueue a filename or query results

I exepected 

quodlibet --enqueue ./"Madonna - True Blue - White Heat.mp3"

to enqueue a song but it instead tries a query.  

As far as I can tell, to enqueue a song requires an absolute file name (i.e., on Linux starting with a /).

The documentation should be revised, or, better, the argument checked to see if there is a file with that name and if so enqueue the file.

pfps:
Aah, it's a bug!

The problem is that enqueue doesn't turn its input into an absolute path.  That's tried in the running quodlibet, but if the path is relative it generally won't work there.

Test by starting quodlbet in a directory with songs.  Enqueue a relative path from that directory - it works.  Enqueue a relative path from a different directly - it doesn't work.   The fix should be simple - stand by.

pfps:
Both should be fixed in https://github.com/quodlibet/quodlibet/pull/2656

pfps:
Fixed in recent pull

pfps:
I have a clean pull request that I think fixes this bug
https://github.com/quodlibet/quodlibet/pull/2685
This also includes some testing.

pfps:
The test harness doesn't work right on the Mac.  I think this is due to some symbolic linking.  I guess the solution is to just skip the test on the Mac, but I don't know how to do thi.

PR #2998, opened 2018-10-18
Fix filesel relative paths

ohel:
From the Python docs for os.path.normpath:
> Normalize a pathname by collapsing redundant separators and up-level references so that A//B, A/B/, A/./B and A/foo/../B all become A/B. This string manipulation may change the meaning of a path that contains symbolic links. On Windows, it converts forward slashes to backward slashes.

I couldn't figure out a case where there would be a . or .. in the path in normal use, as all the paths Quod Libet or Ex Falso receives are either from the file browser or drag'n'dropping or enqueueing if I am correct. The former two never result in . or .. in the path, and the latter actually does not support relative paths at the moment - I think this might be worth another issue maybe?

Therefore I think it's a problem occurring only with that specific test case, but it shouldn't break anything either to do the path normalization in DirectoryTree's go_to and not just fix the test.

### DIFF
--- a/quodlibet/quodlibet/qltk/filesel.py
+++ b/quodlibet/quodlibet/qltk/filesel.py
@@ -287,9 +287,9 @@ class DirectoryTree(RCMHintedTreeView, MultiDragTreeView):
     def go_to(self, path_to_go):
         assert isinstance(path_to_go, fsnative)
 
-        # The path should not be relative in normal situations.
+        # The path should be normalized in normal situations.
         # On some systems and special environments (pipenv) there might be
-        # a relative path at least during tests, though.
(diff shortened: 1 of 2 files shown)


lazka:
looks ok to me, but there are no relative paths here, `/usr/lib/python-exec/python3.6/../../..` is an absolute path. Maybe call it non-normalized instead?

ohel:
Ahh, true. I got caught in thinking relative paths as containing .. but you're absolutely (d'oh) correct, it is an absolute path, albeit a non-normalized one. I'll fix the terms.

The enqueuing not working with relative paths is true, though, and I just realized there's already an issue (and a PR) about it: #2653

lazka:
Thanks!

Issue #1994, opened 2016-07-22
Docs: build list of plugins with descriptions


declension:
Should probably happen at the Sphinx stage.

lazka:
In general I'd prefer improving the plugin description if something isn't clear.

declension:
+1

I was thinking that this list would be _built_ from the descriptions

lazka:
What's its use then?

urielz:
@lazka It would increase the exposure of the plug-ins if they're documented. This could help decide potential new users whether QL is what they are looking for, and help new users discover the plugins, as right now they are a bit buried in the UI.

PR #3001, opened 2018-10-19
docs: Generate list of plugins from source

frestr:
Initial attempt at implementing #1994.

Works pretty well I would say, but doesn't include plugins with optional dependencies, as exceptions are thrown when importing the modules. Need to find some workaround for this.

The generated page looks like this:

plugin_list_docs

### DIFF
--- a/quodlibet/docs/conf.py
+++ b/quodlibet/docs/conf.py
@@ -26,7 +26,8 @@ needs_sphinx = "1.3"
 
 sys.path.append(os.path.join(dir_, "ext"))
 
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.extlinks', 'contributors']
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.extlinks',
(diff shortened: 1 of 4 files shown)


frestr:
So I guess the quodlibet imports in the plugin module won't work, but it might be possible to use `exec_module` in `conf.py` instead? That worked for executing `const.py` and using it in 9d83bd27e1367b92477cd8e48c93b66093ae6ca6 at least.

lazka:
That works because const.py only depends on the stdlib.

frestr:
Right. I haven't tried it yet, but it seems to be possible to install external dependencies as described in [the rtd docs](https://docs.readthedocs.io/en/latest/faq.html#my-project-isn-t-building-with-autodoc). Non-Python dependencies like Gtk can be [mocked](https://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules), unless some Gtk code needs to be executed when loading the modules.

PR #3004, opened 2018-10-21
ext/mediaserver: Show config file in description

### LINKED ISSUE #2829: UPnP MediaServer plugin appears completely undocumented
I would like to use the UPnP / DLNA / etc media server plugin and there appears to be no documentation for it anywhere, including in the source. Enabling it and running 'rygel' before or after 'quodlibet' does not appear to do anything.

frestr:
Fixes #2829 

Regarding the failing test: should we loosen up the requirement that a plugin description has to end with a period? It's a bit artificial to add one in this case.

Another option could be to add an extended description attribute with different requirements, but that might be a bit overkill.


lazka:
I'd add it separately as a label in the preferences below. It's not really a description.

frestr:
Good idea, I agree.

lazka:
(azure CI is broken right now, please ignore)

Issue #2380, opened 2017-05-09
Ex Falso Song list: disappearing focus


sphh:
How it behaves:
1. I open a directory with some song files. The list is longer than can be displayed in the Song list widget of Ex Falso.
2. I select a song near the end of the list by scrolling down in the Song list widget.
3. I add a tag to the song.
4. I Save my changes.
5. The selected song disappears because the list shows the first song at the top of the Song list widget.

How I expect it to behave:
- After saving my changes I would like to still see the selected song in the Song list widget, preferably at the same position in the widget before saving.

PR #3005, opened 2018-10-21
feat: #2380: remember filelist scroll position on tag save

ohel:
Fixes usability issue #2380: the file list was always scrolled to top when saving tags in Ex Falso. Now it remembers the position.

I had to handle this with the draw signal, as any attempts to change the vadjustment during other signals were in vain - the scroll position was always reset.


lazka:
We have a similar thing for the main songlist where we scroll to the re-selected row after refill.

PR #3027, opened 2018-11-03
Upload Hebrew Translation for Quodlibet 4.2

avma:
https://gist.github.com/avma/7e32691c190441a20fe99ea6a5f5ce64#file-he-po

Hope i'm doing it right....
my regards
Avi


lazka:
@avma Thanks. Looks like something went wrong there as the PR contains unrelated changes and nothing of your changes.

I'll try importing the gist manually

avma:
hummm, ok just let me know if there is any other way to submit it...

lazka:
I've create #3028 with your changes.

Submitting per gist etc is fine. The downside is that we have various automated tests that run on the translations when you open a PR with them which makes things easier for me.

PR #3032, opened 2018-11-04
Improve song ratings options on command line

Phidica:
Change the `set-rating` option to `rating` by analogy to the `volume` option, and allow it to take relative ratings such as `+0.1` to increase the rating of the currently playing song by 0.1.

Add additional `rating-up` and `rating-down` options by analogy to the `volume-up` and `volume-down` options which increase or decrease the rating of the currently playing song by one star, based on the user's active preference of the number of stars to rate songs out of.

Additionally change the notation for the help of the `volume` option from `(+|-|)0..100` to `[+|-]0..100` which is then consistent with the notation used on other help lines for indicating when characters are optional.


lazka:
The tests in test_commands are pretty basic atm, which is why most command implementations just call other API which is easier to test.

Phidica:
`test_commands` still won't pass because the `song-list` option doesn't seem to exist any more. Otherwise I think it's good

PR #3039, opened 2018-11-04
mediaserver: Make label about conig file wrap and text selectable

Mellthas:
If the translation for the label is very long, the plugin window can become very large without the label being wrapped.

All other plugin text is selectable, so make this label selectable, too.


frestr:
Thanks

Issue #2948, opened 2018-08-28
Can't get back to full library after selecting a playlist


echosa:
I'm trying out quodlibet on macOS. When I select a playlist, there's no way to "unselect" it. The result of this is that I can't get back to the full library view unless I close and reopen the program.

Update: I found two ways, but they're not great. 

1.) If I go back to the "Search Library" view after selecting a playlist, it only shows the the songs in the playlist. The search bar is empty. However, clicking the "delete" icon on the right of the empty search bar shows all the songs. I guess clicking that "unselects" the playlist, but that's really confusing UI.

2.) If I select a playlist and then go into the Paned Browser view, the bottom pane still just shows the songs in the selected playlists, but the artist and album panes show all artists and albums. It seems like those panes should also be filtered to just what's in the playlist, but until that bug is fixed, this is at least a workaround for the "can't unselect playlist" bug.

frestr:
This seems to be the same issue as described in #2374.

We should update the song list on browser changes, as there's not much point in keeping the old view visible, especially since playing a song from an old view removes the song from the view.

echosa:
Yes, I believe that that the second point in that issue describes what I'm experiencing. I'll close this duplicate ticket. Thanks.

Issue #2374, opened 2017-05-06
filtering/sorting in one Browse view affect others in unexpected ways


fecraow:
this may warrant multiple bugs -- if this should be split up, please let me know.

i'm only going to describe how this affects switching between "Paned Browser" and "Playlists", as these are the ones I use, although other views may be affected in other ways....

1)
okay, so when I open quodlibet, it's in "Paned Browser", because that's what I most often use. If I switch to "Playlists", the right pane starts filled with whatever songs show up with the filters that were selected in Paned Browser (or all songs if no filters were applied), whereas I'd expect no songs to show up until one of the playlists is selected (and that is actually what happens if you open quodlibet and the last view was Playlists). Once you select a playlist, the view appears as you'd expect.
(I don't consider the above an especially big deal, but I include it to help illustrate what's going on.)

2)
now, when I switch back to "Paned Browser", the list of songs remains the last selected playlist from Playlist view. There's no UI indication that a playlist is selected. All my filters have "All" selected, and yet the song list is including only songs from a particular playlist. I have to click on one of the (already selected) "All" lines again in order to show all my songs.

3)
once i've got all the songs showing again, I notice that the song list no longer has any apparent sorting (in Paned Browser, I normally sort by Title)

should the filters and sorting be done per-view to avoid these types of issues?

declension:
Thanks for the report and sorry about the delay. This kind of issue has come up a few times. The songlist is confusing sometimes in QL especially in the way it persists across browser changes as you describe.

This sounds like two issues to me:
 * Refreshing the songlist moving to Playlists browser (capture in #2053)
 * Sorting per browser or sort order being forgotten somewhow. 

Perhaps we should reduce this issue to the latter and continue the other discussion on #2053

frestr:
Fixed in #3041.

Regarding the 3rd point: the sorting is reset when switching to the playlist browser since one usually wants to restore the original playlist order. When switching back to another browser the sorting is reset, since state across browser switches aren't retained (captured in #1850).

PR #3041, opened 2018-11-05
Refresh song list on browser changes

frestr:
The issue about song selections persisting across browser switches has been described in #2948 and #2374. I think this refresh behavior is more intuitive.

Since the song list will now be empty after switches if no album is selected, the cover grid and album list browser will automatically select the "All Albums" album. This is the same behavior as for the paned browser (and practically the same as the search and iradio browser).

### DIFF
--- a/quodlibet/quodlibet/browsers/albums/main.py
+++ b/quodlibet/quodlibet/browsers/albums/main.py
@@ -401,8 +401,7 @@ class AlbumList(Browser, util.InstanceTracker, VisibleUpdate,
 
     def finalize(self, restored):
         if not restored:
-            # Select the "All Albums" album, which is None
-            self.view.select_by_func(lambda r: r[0].album is None, one=True)
+            self.view.set_cursor((0,))
 
     @classmethod
     def _destroy_model(klass):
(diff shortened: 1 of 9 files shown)


frestr:
The `browser.active()` call may need to be moved (or the refresh done in some other way).

lazka:
Would `if not restored: self.browser.unfilter()` also work?

Ideally the song list should be only filled once on QL start as that can slow things down a lot with a large library.

frestr:
> Would if not restored: self.browser.unfilter() also work?

Yeah, that's probably a better way to do it.

> Ideally the song list should be only filled once on QL start as that can slow things down a lot with a large library.

Perhaps we could clear the song list instead of reloading it? Not as intuitive I would say, but I  see the point with a large library.

lazka:
> Perhaps we could clear the song list instead of reloading it?

I'm ok with filling it, I just wanted to point out that this currently calls activate() twice on restore afaics, which makes the QL start slower (I should probably add a test for this so we don't regress)

frestr:
Ah, I see. That makes sense.

frestr:
Any idea why the MPRIS test fails after a browser activate()?

lazka:
> Any idea why the MPRIS test fails after a browser activate()?

No :/

PR #3067, opened 2018-11-23
Added 'Thumb Rating' Plugin

TheYokai:
## Thumb Rating System

> This plugin allows users to vote on the currently playing song using a thumb-up/thumb-down tally system. This has some benefits over the vanilla star-rating system by allowing users to make an easier decision multiple times in order to accurately calculate a final star total. It will also more accurately measure a users affinity to a song over a period of time. Unlike the existing auto-rating plugin, this plugin allows for a neutral state where the rating does not change. It also doesn't make the assumption that a skipped song is an inferior song.
    
>  You can sort you songs by a score tag ("~#score") in order to sort your music from least liked to most liked.

<p align="center">
  
</p>

This plugin currently uses the Addon GUI space to add the thumb up and down buttons. However, I think it would be useful to add the ability to add plugin buttons to the playhead area to allow plugins to add additional play controls. A mock-up of what I think might be useful at some point:

<p align="center">
     
</p>

I'm curious to see if others agree with me on this idea. Basically, I see the secondary area after the play controls as a HBox of multiple plugin controls inserted into that area. It would be useful in any circumstance like this where you want to add a few simple buttons for either playback or actions on an active song.

### DIFF
--- a/quodlibet/quodlibet/ext/events/thumbrating.py
+++ b/quodlibet/quodlibet/ext/events/thumbrating.py
@@ -1,0 +1,154 @@
+# -*- coding: utf-8 -*-
+#
+# View Lyrics: a Quod Libet plugin for viewing lyrics.
+# Copyright (C) 2018 Eoin O'Neill (eoinoneill1991@gmail.com)
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+from gi.repository import Gtk
+from quodlibet import _
+from quodlibet.plugins.events import EventPlugin
+from quodlibet.qltk import Icons, ToggleButton
+from quodlibet.plugins.gui import UserInterfacePlugin
+
+
+class RatingBox(Gtk.VBox):
+    def __init__(self):
+        super(RatingBox, self).__init__(self)
+
+        self.thumb_ups = 1
+        self.thumb_downs = 1
+
+        self.title = Gtk.Label("")
+        self.title.set_line_wrap(True)
+        self.title.set_lines(2)
+
+        hbox = Gtk.HBox()
+        self.upvote = ToggleButton("👍", "")
+        self.downvote = ToggleButton("👎", "")
+        self.upvote.connect("toggled", self.__thumb_toggled)
+        self.downvote.connect("toggled", self.__thumb_toggled)
+        self.score_label = Gtk.Label("----")
+        self.upvote.set_property("height-request", 50)
+        self.downvote.set_property("height-request", 50)
+        hbox.pack_start(self.upvote, True, True, 5)
+        hbox.pack_start(self.downvote, True, True, 5)
+
+        self.hbox = hbox
+        self.pack_start(self.title, False, False, 10)
+        self.pack_start(self.score_label, True, True, 5)
+        self.pack_start(self.hbox, False, False, 5)
+
+    def set_current_title(self, title):
+        self.title.set_text(title)
+
+    def set_current_score(self, cth_up, cth_down):
+        self.thumb_ups = cth_up
+        self.thumb_downs = cth_down
+        self.__set_pending_score_value(self.thumb_ups - self.thumb_downs)
+
+    def poll_vote(self, reset=True):
+        upward = 1 if self.upvote.get_active() else 0
+        downward = 1 if self.downvote.get_active() else 0
+        vote = (upward, downward)
+        if reset:
+            self.downvote.set_active(False)
+            self.upvote.set_active(False)
+        return vote
+
+    def __set_pending_score_value(self, score):
+        existing_score = self.thumb_ups - self.thumb_downs
+        if score == existing_score:
+            self.score_label.set_markup("<b>"
+                                        + str(int(score))
+                                        + "</b>")
+        elif score > existing_score:
+            self.score_label.set_markup("<b><span foreground=\"green\">"
+                                        + str(int(score))
+                                        + "</span></b>")
+        else:
+            self.score_label.set_markup("<b><span foreground=\"red\">"
+                                        + str(int(score))
+                                        + "</span></b>")
+
+    def __thumb_toggled(self, button):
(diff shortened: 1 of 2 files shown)


TheYokai:
I'm curious to know how to best solve this issue with missing translation text.

frestr:
> I'm curious to know how to best solve this issue with missing translation text.

You need to add your new file to `quodlibet/po/POTFILES.in`.

TheYokai:
@frestr Thanks for the help!

frestr:
I'm wondering if you could elaborate a bit on the use case of this plugin.

Are you supposed to thumb up/down a song each time you listen to it, which will gradually increase/decrease the rating (since it seems like you can only rate once per play)? And would you ever thumb-down a song you've already given thumbs up?

In essence, what is the benefit over just setting the rating directly (or maybe using the automatic rating plugin)?

TheYokai:
@frestr Yes, I'm sorry I was a bit sparse on the details in the PR. I'll actually add any relevant details to the main post.

>Are you supposed to thumb up/down a song each time you listen to it, which will gradually increase/decrease the rating (since it seems like you can only rate once per play)? 

Yes. I find that it's difficult to give songs a blanket rating as I tend to believe that my affinity with a song will change over time or based on conditions (season, work or leisure, etc.) This was something I dealt with by giving all songs a win-loss ratio in my tagging data which I basically hacked together with a custom foobar2000 configuration and a combination of plugins. While that worked, I began / developed my own rating system plugin that I could install using foobar2000's plugin system. Since then, I have moved to linux and have such decided I need to shift my focus away from foobar and focus more on programs that work across operating systems. 

It's a very simple work around to my problem with blanket rating. Instead of giving a singular rating that can be arbitrary to decide (what makes a song 5/5 vs 4/5?), I decided I'd rather increment the rating over time in a way where I can give any song feedback as I'm listening to it. I have a rather large library of music with multiple people who may be listening to it at a time, and it's also useful for keeping an aggregate of how many votes have been casted for a given song. 

>And would you ever thumb-down a song you've already given thumbs up?

Probably not within the same day, no. But since I have a very large music library, there's a chance that the next time I hear the same song I could feel differently about it. Additionally, the more ratings I give the song (more positive), the better the score gets. For star ratings, this isn't apparent. However, there's an additional metadata of the overall score of a song (upvotes - downvotes) that means you can sort a song from worst to best. This is useful when I want to clean shop (remove songs that I've said I didn't like enough times that I know exactly how I feel about it and there's no chance I'll change my mind.) 

>In essence, what is the benefit over just setting the rating directly (or maybe using the automatic rating plugin)?

I began my use of Quodlibet using the auto rating system but it began to pose a few problems for me. The most obvious one to me is that it didn't really have a good 'neutral' state -- you either like a song or you don't which skewed some of my ratings. The second is that it doesn't keep a running aggregate of how many times you gave direct feedback (skipped, in the case of auto rating case) which means that it can be somewhat difficult to determine whether or not a song was truly "bad" or if I just wasn't in the mood to listen to it.

quodlibet/quodlibet/ext/events/thumbrating.py
```diff
@@ -0,0 +1,148 @@
...
+        self.thumb_ups = 1
+        self.thumb_downs = 1
+
+        self.title = Gtk.Label("")
+        self.title.set_line_wrap(True)
+        self.title.set_lines(2)
+
+        hbox = Gtk.HBox()
+        self.upvote = ToggleButton("👍", "")
+        self.downvote = ToggleButton("👎", "")
```
frestr: I'm not entirely sure how cross-platform it is to use emojis as icons, but I suppose it should be fine. @lazka Do you have any input on this?

lazka: > @lazka Do you have any input on this?

It might not look as nice everywhere (there is no color on Windows for example), but it should be fine.

quodlibet/quodlibet/ext/events/thumbrating.py
```diff
@@ -0,0 +1,148 @@
...
+        vbox.show_all()
+        return vbox
+
+    def disabled(self):
+        self.rating_box.destroy()
+
+    def plugin_on_song_ended(self, song, stopped):
+        if song is not None:
+            ups = max(int(song.get("~#wins") or 0), 1)
+            downs = max(int(song.get("~#losses") or 0), 1)
```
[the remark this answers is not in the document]
TheYokai: I will actually look into handling this better. Basically, it was intentional but it's a work around for two conditions. The first is divide by zero cases, and the second is having ratings start in a more 'neutral' condition. I believe I have found a work around, but I want to test it out with my own collection for a while before I push it upstream.

quodlibet/quodlibet/ext/events/thumbrating.py
TheYokai: Ideally, this would use SVG but I found that emoji were a decent work around for getting the prototype ready and legible. I'm not sure on what the policy is with plugins that may require additional resources (like images or sounds.)

TheYokai:
I've made a few changes now to the rating algorithm to address my original concerns of divide by zero conditions without having to pad the score with an extra loss. The rating calculation will now use a baseline of 2 votes in order to kick-off the score to rating conversion. Also, tags are no longer updated when no vote has been made (in order to prevent any potential data change on neutral playback state.)

FYI, I've rebased my branch to a single commit for a more legible git history when / if it gets pulled into the main branch. I'm not sure if that's the standard you guys have here on this project, but I believe it will be of benefit to remove any excess and unnecessary commit messages.

frestr:
> I've made a few changes now to the rating algorithm to address my original concerns of divide by zero conditions without having to pad the score with an extra loss.

Nice, this seems better.

> I've rebased my branch to a single commit for a more legible git history when / if it gets pulled into the main branch.

IMHO, it's better if you just make a new commit on top of the old ones (i.e., no rebasing). This makes it easier to review the new changes, as I otherwise kind of have to review the whole thing again. The commits can be squashed (in GitHub) before merging to keep the commit log clean.

Not a big issue here, but can be good to keep in mind for next time.

Issue #2912, opened 2018-07-15
"Ignore queue" feels unintuitive


declension:
Since #2865 was merged (apologies I couldn't feed back more on the original PR, been pretty busy), we now have some new queue options which is great.

One of them, now named _Ignore Queue_ disables the queue (thus changing how all song selection works), but remains within the queue widget, and within _its_ configuration widget. This can lead to some confusion e.g.:

### Steps
Consider a common usage pattern (at least I use it a lot):
 * Enable _Ignore queue_
 * Come back later
 * Browse some music
 * Listen to a track
 * Select some interesting other tracks (e.g. an album)
 * Press ctrl-enter to enqueue.
 * Keep browsing to something else

### Expected
These enqueued tracks / album to play next

### Actual
The next in my current browser plays.

### Ideas
I can see benefit to some for the original use-case (_to occasionally ignore the queue without having to clear it_), but I'd suggest if we keep it we do some UX work. Some ideas:
1. Rename the option itself: _disable_ sounds stronger and thus more appropriate to me here.
2. Move it out of the prefs widget for the queue. It feels unusual to find a preference saying "disable me" alongside some fine-tuning.
3. If the queue is effectively disabled, indicate this with the UI, e.g. by greying out all the tracks, or some visual indication at the menu / queue widget level maybe. 
4. Related to above, an obvious enable / disable toggle on the queue itself

declension:
Related note: _Queue Only_ shuffle mode, with _Ignore Queue_ checked is a little bit... _mutually assured destruction_ :thinking:

Treferwynd:
About you related note, IMHO Queue Only shouldn't be a shuffle mode, or even a plugin, but a setting.

[I proposed a selector with 3 choices:](https://github.com/quodlibet/quodlibet/issues/2856#issuecomment-405021234)

* Standard Behaviour: the queue works as it always did
* Ignore Queue: as if the queue was empty
* Only Queue: only stuff in the queue gets played. Which would include the Queue Only plugin

But @frestr (the author of #2865) didn't particularly like the selector, so he may have other ideas!

frestr:
> If the queue is effectively disabled, indicate this with the UI, e.g. by greying out all the tracks, or some visual indication at the menu / queue widget level maybe.

[...]

PR #3074, opened 2018-12-05
Improve ignore & keep songs functionality in queue

frestr:
This is an attempt to address the issues discussed in #2912.

The two main changes are:
- Replace the ignore option with a disable button in the queue header bar. When the queue is disabled, it cannot be played from (but songs can still be added to it).
- Replace the keep songs option with a mode submenu.

@declension @Treferwynd thoughts?

### DIFF
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -68,7 +68,7 @@ INITIAL = {
         "shufflequeue": "false",
         "queue_stop_at_end": "false",
         "queue_keep_songs": "false",
-        "queue_ignore": "false",
+        "queue_disable": "false",
 
         # <reversed?>tagname, song list sort
         "sortby": "0album",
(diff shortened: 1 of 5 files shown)


Treferwynd:
Very nice!

Do you think we can add an option to loop the queue? A simple check would conflict with `Stop at End`, an alternative could be a `Behaviour` submenu with mutually exclusive entries like `One Time`, `Stop at End` and `Loop` (#2911 is related to this, but I'll probably close it)

lazka:
Looks good!

declension:
@frestr nice work. Will take a bit of getting used to, but feels good :+1: 

Maybe the tool-tip on the padlock button could be more descriptive: _disabling_ the queue isn't quite what I intuitively expected as I can still add things to it (or should _that_ be disabled too?)

Maybe add "- playing will ignore the queue", or similar?

frestr:
Thanks for the feedback, guys.

> Maybe the tool-tip on the padlock button could be more descriptive

Agreed. I'll update it.

frestr:
@lazka Maybe we could build docs for release 4.2 (and 4.1) too at RTD? As the queue documentation will be different for 4.2 and master/4.3.

lazka:
We no longer have version specific docs starting with 4.1 because it turned out to be more confusing than helpful all in all.

I can build them, but the QL help menu links to the latest one for those versions.

frestr:
Ok, makes sense. I guess we can include both the old and new description with a reference to the version they apply to instead (we kind of do this already in the current queue doc).

PR #3076, opened 2018-12-10
Refactored loop to generator comprehension in LibraryBrowser.save

blimmo:
Instead of creating a list, filling it up and then iterating over it, just pass a generator comprehension to `str.join`.


declension:
Thanks

Issue #809, opened 2015-03-14
Allow user to build searches from selectable sub-searches / global filters


lazka:
Original [issue 809](https://code.google.com/p/quodlibet/issues/detail?id=809) created by AbstauBaer on 2011-09-13T12:22:30.000Z:

We know the huge potential of the search-filters in this nice piece of codework.

But I think there is yet missing some more convenient and intuitive ways of integrating that power to the UI.

On the one hand I want to switch multiple custom pre-defined (saved and named) filters on and off while maintaining my music library.

On the other hand I want to have a intuitive UI - for example when friends are using quodlibet - where the &quot;search&quot;-bar is free to be used for simplest things like typing the first three letters of the artist i wanna hear instead of scrolling in the alphabetical list.

Shortterm I think there's not even much to do: What I'd love to have is the ability to save searches to appear in the &quot;Filter&quot;-Menu and not just the ability to select them and have the results displayed but also a little checkbox that makes it possible to select one or combine multiple filters and have them stay active while I can still use my searchbar for whatever I want.

Of course I could do that by combining my search-strings in the search-bar. But as I said: what if a friend comes around and doesn't know anything about qoudlibet?
Now I could copy my search-string to the global filter. But that's where it becomes unhandy for me as it's hidden on the second pane of settings dialogue.
I think we could actually leave that global filter away in settings and save that passage in the faq (&quot;Maybe you have a global filter active&quot;) if we find an intuitive way of activating/deactivating custom made filters and showing that they are active.

Longterm one might find a decent looking way of integrating that into the UI.

++++++++

Shortterm I think having Filters checked and unchecked in the &quot;Filter&quot;-Menu is already pretty intuitive.

++++++++

(Longterm one might also make it possible to link certain filters to certain browser-views and so on but that is really 5 Star-Convenience)

Well, I hope you get my point; and like it :]

Just if you still wanna read more some examples of use that made me come up with this idea:
I loved it when I found out qoud libet could feed it's library from multiple sources.
Cause I got one big bunch of unsorted music in one folder, where I can go to and search like on a junkjard. I got one folder with sorted music that I yet didn't hear and need to before I decide about whether I wanna have it in the last folder with the sorted music i like to have in my library.

Now it's great to have all of them in quod libet, because by renaming, musicbrainz integration and the Filesystem view I got everything I need to manage my music in one program and I can just listen to it without something opening because that program is a genious music-player.

[...]

PR #3082, opened 2018-12-18
Option for multiple queries in Search Browser

blimmo:
This is a prototype for ~~a new browser, similar to the Search browser, that creates a query by &ing together subqueries.~~ an option on the search browser to allow multiple queries to be entered, with the overall query generated by &ing together the subqueries.

The idea is that the user can leave some subqueries alone (say `#(rating >= 0.5)`) while changing another (say `genre = classical`). I think it could also help people unfamiliar with the ql query language use the search browser especially if using saved searches.

Possibly addresses #809 but that issue is rather broad.

Example:
image

Todo:
- [x] Separator between queries
- [x] Save queries on close
- [x] ~~Automatically add queries when chosen from combobox dropdown~~ Add button

Possible future changes
- Clear button (on right click?)
- Display names for saved searches added
- Negation button?
- Edit queries?


blimmo:
I've got stuck with the third item in the todo list: detecting when a saved search has been clicked in order to automatically add it.

The `SearchBarBox` simply emits a `query-changed` signal whenever a valid query has been input and doesn't distinguish between typed ones and saved ones chosen.  
In fact digging a bit deeper it looks like `ComboBoxEntrySave` doesn't distinguish either.

Should I edit `ComboBoxEntrySave` to emit a new signal when a saved query is chosen?  
(also I can't find anything on how Gtk signals work: are they propagated to parent widgets like events?)

blimmo:
On second thoughts adding queries automatically when chosen from the combobox is probably against the idea of comboboxes.  
Instead I've added a button to add chosen queries easily with the mouse.

I'd appreciate comments on the UX (I'm biased since I wrote it):
- Do I need some more explanation of what's going on to the user? I think it's pretty obvious except possibly that the queries are being &ed but I can't think of a good way to display that.
- I think the name could use some work.

frestr:
This seems like a nice idea. We already have quite a few browsers though, and this is pretty similar to the search browser. Perhaps it could be added to the already existing search browser instead? It doesn't interfere with the old use case (single query) as far as I can see.

frestr:
> Maybe it should be an option on the preferences button to enable multiple searches?

That could work too.

blimmo:
Ok, I've attempted to move the functionality I had in my separate browser into a new version of `SearchBarBox`.

Unfortunately I obviously want the listbox underneath the `SearchBarBox`'s combobox so I've changed `SearchBarBox` to inherit from `Gtk.Grid` instead of `Gtk.HBox` which Gtk recommends over boxes apparently.

This promptly broke all the other browsers using `SearchBarBox.pack_start` so I implemented a replacement pack_start.

I think just having an HBox like before in a VBox with the listbox also in would be much easier but I can't think of a way to do that within the `SearchBarBox` object since it is itself the HBox I'd want to put in the VBox.

Anyway I also seem to have broken a whole bunch of other tests and I have no idea why. I'll look at them in more detail tomorrow.

blimmo:
I've changed `SearchBarBox` back to inheriting from `Gtk.Box` and fixed the tests (apparently missing a config option from the default config breaks 221 tests :P).

To get the listbox underneath the rest of the search bar I'm packing the listbox into the actual `SearchBar` browser outside of the `SearchBarBox` which is obviously less than ideal (`MultiSearchBarBox` doesn't work unless the user does this).

blimmo:
Ok, I'm quite happy with this now.

The beheaviour of the Search Browser is unchanged unless the user ticks a checkbox in the preferences button so this shouldn't be disruptive to existing users.

I think the other items I had on the todo can be added later; I might make another PR or two later.

The only thing I'm really concerned about is packing the `FlowBox` in `SearchBar` instead of `SearchBarBox` but I can't think of a solution to that.

Any comments or suggestions welcome.

frestr:
Nice, it looks pretty good now.

Some feedback:

- The flowbox acts is a bit funky with multiple lines, as it tries to make the items in every "column" of the same length. This can be a bit annoying when later removing queries, as everything is reflowed. Not sure how to prevent this though.
- The "search after typing" option seems to be ignored in multiquery mode.
- Doing queries in non-multiquery mode adds queries to multiquery even though it is disabled. This can be seen by doing the following:
  1. Disable multiquery and "search after typing".
  2. Make a query and press enter to execute it. Repeat a few times.
  3. Turn on multiquery.
- A "clear queries" button when right-clicking the flowbox would've been nice (not very important though).

> The only thing I'm really concerned about is packing the FlowBox in SearchBar instead of SearchBarBox but I can't think of a solution to that.

It's not very clean, but I don't think it should be an issue.

frestr:
> I guess it's mostly a problem when removing a few queries at once?

Yeah, it's not a big issue, so it's okay to leave it for later.

One last thing: it would probably be a good idea to disallow adding empty queries (which happens when pressing enter with no query text).

blimmo:
Ah good point. Thanks again for your feedback!

blimmo:
Yeah, there's nothing more I want to do atm so unless you can see any problems feel free to merge!

PR #3086, opened 2018-12-21
Github: add / update templates for new Issues and PRs

declension:
* Split issue templates into bug-fixes and feature requests. :)
 * Formalise these a bit, but largely keeping to existing contribution guidelines.
 * Move existing Contributing guide to Markdown (wasn't showing in Github)
 * Add a PR template, with some sections and guidelines
 * TODO: the actual Github config to use different template types

Comments welcome :)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,12 @@
+Contributing to Quod Libet
+--------------------------
+
+When raising a new issue:
+ 1. Choose whether it's a bug of a feature request
+ 1. Check the existing issues.
+ 1. Fill in the appropriate template as described there
+
+If you have related ideas, please file them separately and mention
+the issue numbers of previous ideas.
+
+For more details, see the [contributing guide](../quodlibet/docs/development/contributing.rst)
--- a/.github/CONTRIBUTING.rst
+++ b/.github/CONTRIBUTING.rst
@@ -1,9 +1,0 @@
-::
-
-    When creating a new issue please try to answer the following questions:
-
-    * What did you try to do?
-    * What did you expect to happen?
-    * What did happen instead?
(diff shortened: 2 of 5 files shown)


lazka:
looks good

frestr:
Very nice!

One comment: some tickets are more about support than bugs / feature requests. Should we have a template for this too, or rather refer people to the mailing list / IRC (which I'm not really sure if many people are using nowadays)?

declension:
> One comment: some tickets are more about support than bugs / feature requests. Should we have a template for this too, or rather refer people to the mailing list / IRC


It's a good point. I'm not sure what the template would say though... other than "feel free to file this on Github... but it'll probably get closed as a polite `wontfix` or `invalid`" of some sorts... Maybe if it were some kind of "question" template, this would be a good enough compromise between Issue noise and useful way of supporting users?

>  (which I'm not really sure if many people are using nowadays)?

Yeah IRC does seem a little dead now. We could use a Gitter / Slack, but there are of course some valid reservations about using non-free solutions for communication. Anyway, that's a _whole_ separate discussion :grimacing:

frestr:
> I'm not sure what the template would say though...

Yeah, I don't really have any good suggestions there. I suppose some questions can be formulated as bug reports or feature requests (if the feature already exists, then explain how to use it), so maybe we can begin with these two templates and see if it's sufficient or not (one can also opt not to use a template if none of them really fit, if I'm not mistaken).

> We could use a Gitter / Slack, but there are of course some valid reservations about using non-free solutions for communication. Anyway, that's a whole separate discussion

Agreed. Could be worth taking in another issue though.

declension:
Cool. I'll merge this now then and configure the two current templates, then we can think about a third one or any further tweaks as it gets used more...

PR #3098, opened 2019-01-04
Scale player volume properly in mpris2 API

luk1337:
* This fixes an issue where KDE Connect Android app
  is showing different volume than Quod Libet GUI.

### DIFF
--- a/quodlibet/quodlibet/ext/events/mpris/mpris2.py
+++ b/quodlibet/quodlibet/ext/events/mpris/mpris2.py
@@ -338,7 +338,7 @@ value="false"/>
             elif name == "Shuffle":
                 player_options.shuffle = value
             elif name == "Volume":
-                player.volume = value
+                player.volume = value ** 3.0
 
     def get_property(self, interface, name):
         player = app.player
@@ -385,7 +385,7 @@ value="false"/>
             elif name == "Metadata":
                 return self.__get_metadata()
             elif name == "Volume":
-                return player.volume
+                return player.volume ** (1.0 / 3.0)
             elif name == "Position":
                 return player.get_position() * 1000
             elif name == "MinimumRate":


lazka:
We use a different scaling for the UI slider. If KDE Connect uses a different one there might be a difference.

luk1337:
Well, I copied these magic calculations from `quodlibet/build/lib/quodlibet/qltk/controls.py` so I assumed that it's not KDE Connect's fault but Quod Libet's.

luk1337:
Btw `quodlibet/build/lib/quodlibet/ext/events/mpdserver/main.py` does the very same thing. See https://github.com/quodlibet/quodlibet/blob/22190604238d9d04afb6cb728bc1a4416a17bae6/quodlibet/quodlibet/ext/events/mpdserver/main.py#L273 https://github.com/quodlibet/quodlibet/blob/22190604238d9d04afb6cb728bc1a4416a17bae6/quodlibet/quodlibet/ext/events/mpdserver/main.py#L311

luk1337:
I also tested https://github.com/acrisci/playerctl and after these changes it's able to get and set proper volume.

lazka:
> Btw `quodlibet/quodlibet/quodlibet/ext/events/mpdserver/main.py` does the very same thing. See

Heh, yeah, see 44aa244408ae. I guess I was similarly annoyed by the scale mismatch :)

MPRIS is a bit different as there are many clients and we can't just expose non-linear volume as a linear one. It might improve things with your client but break things with other clients applying a scaling themselves.

lazka:
> Heh, yeah, see [44aa244](https://github.com/quodlibet/quodlibet/commit/44aa244408aef31236a22e145dfc31565b14ca06). I guess I was similarly annoyed by the scale mismatch :) I should have filed a bug with mpdroid instead. :/

Ah, that's because mpd exposes the internal pulseaudio volume directly, so this is more of a workaround for a mpd bug.

luk1337:
Uhm, so what clients are known to work with linear volume?

lazka:
I've opened an issue for the MPRIS spec: https://gitlab.freedesktop.org/mpris/mpris-spec/issues/8

jonaski:
@lazka Did you ever receive any reply form the MPRIS2 folks?
I recently implemented volume synchronization Strawberry and would like to know if I should use linear or cubic for MPRIS2, I have been using linear until now. All other players I tested uses cubic in MPRIS2, but I assume this might be because these are all GTK and they have cubic volume slider for the GUI. The advantage with using cubic is that it will match the percent in the PulseAudio volume control.

Issue #1244, opened 2015-03-15
More videogame formats tagging support


lazka:
Original [issue 1244](https://code.google.com/p/quodlibet/issues/detail?id=1244) created by  aurelien...@laposte.net on 2013-09-28T13:23:32.000Z:

The tagging tool is great for dealing with multi artists and others, so I would like to use this player to read all my audio files among which I have some rarer formats, such as SID, SPC, VGM or PSF.

The less uncommon format in this list is SPC, for which tagging is already supported, but not playback yet as far as I know (I only have an error message).

Of course, one may say I should use Ex Falso for tagging only and another player for playing back, but as Rhythmbox or Audacious for example won't allow me to select tracks through the multi artist tag, this very tagging will be useless.

Result: I have to choose between nicely tagged files + 2 players (1 for standard files and 1 for exotic files, i.e.: no shuffle, not handy) and poorly tagged files + 1 player, because for now I am unable to find any player which can do both.

Many geeks with large libraries who like videogame musics will face this dilemma, which could be solved by adding support for these file formats.

I admit such a support is not in the priorities, but it would definitely be a great addition to have a complete player.

PR #3100, opened 2019-01-05
Improved `VGM` Metadata Tag Parsing. (GD3 Support)

TheYokai:
## Overview
This PR gives Quodlibet the ability to properly scan `vgm` files for title, album, artist, and other tags available via the [gd3 tag standard](https://vgmrips.net/wiki/GD3_Specification). It uses Quodlibet's multiple tag system to store both Japanese and English entries (when applicable.) This should help improve the user-experience of browsing a library with VGM files. ( related to: #1244 )

### Future Goals
There are more improvements that can be made here. Specifically, it would be nice if QuodLibet could work with additional APETAG footers on `vgm` and `spc` files to store information such as genre, rating, replaygain-data, and any other additionally unsupported tag entries. This is how some media players (foobar2000) like to embed additional tag data to chiptune formats. It would also be nice to have proper tag writing features for all supported chiptune formats. 

### The State of VGM w/ GStreamer (on Linux)
It's worth noting that the VGM format libraries used by GStreamer on Linux are very out of date. The current situation regarding libgme on linux is complicated, but the short of it is that there are many different forks that exist, each one supporting a different range of systems. 

If you want to gain additional system playback options (and additional sample playback accuracy) compared to the standard libgme installation, I would suggest you build [wyatt8740/Game_Music_Emu](https://github.com/wyatt8740/Game_Music_Emu) from source. Wyatt's repository is a fork of @kode54 's GME playback libraries but with added CMake support (allowing linux systems to build the library.) I'm currently working on my own fork of wyatts work locally, hoping to merge with it and eventually merge it with the KDE managed branch at [bitbucket::mpyne/game-music-emu](https://bitbucket.org/mpyne/game-music-emu/). If you are running quodlibet on Windows, you may be better off using [kode54/Game_Music_Emu](https://github.com/kode54/Game_Music_Emu).

My expectation is that GStreamer's vgm playback support will improve significantly in the future. @ValleyBell is working on a new vgm playback library [ValleyBell/libvgm](https://github.com/ValleyBell/libvgm) which should be able to support GStreamer with an eventual plugin (see discussion here: vgmrips/vgmplay#45)
 
Hopefully that wasn't too long winded. My main reason for explaining this stems from the potential to use a custom libgme build for quodlibet's flatpak, which would allow quodlibet to use a custom binary without interfering with system level libraries (ideal solution.)

### DIFF
--- a/quodlibet/quodlibet/formats/vgm.py
+++ b/quodlibet/quodlibet/formats/vgm.py
@@ -5,7 +5,9 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 import struct
+import os
 
+from senf import path2fsn, fsn2text
 from ._audio import AudioFile, translate_errors
 
 # VGM and GD3 SPECs:
@@ -61,16 +63,12 @@ class VgmFile(AudioFile):
                 h.seek(GD3_TAG_PTR_POS + gd3_position)
                 self.update(parse_gd3(h.read()))
 
+        self.setdefault(
+            "title", fsn2text(path2fsn(os.path.basename(filename)[:-4])))
         self.sanitize(filename)
 
     def write(self):
         pass
-
-    def reload(self, *args):
-        title = self.get("title")
-        super(VgmFile, self).reload(*args)
-        if title is not None:
-            self.setdefault("title", title)
 
     def can_change(self, k=None):
         if k is None:
@@ -85,7 +83,7 @@ def parse_gd3(data):
         print('Invalid Gd3, Missing Header...')
         return tags
 
-    version = data[4:8]
+    # version = data[4:8]
     # Should be [0x00, 0x10, 0x00, 0x00] currently.
     # We should hold onto it for possible branching if standards change.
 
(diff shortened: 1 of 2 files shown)


quodlibet/quodlibet/formats/vgm.py
```diff
@@ -56,6 +79,40 @@ def can_change(self, k=None):
...
+
+    version = data[4:8]
+    # Should be [0x00, 0x10, 0x00, 0x00] currently.
+    # We should hold onto it for possible branching if standards change.
+
+    gd3_length = struct.unpack('<i', data[8:12])[0]
+    # Length of gd3 footer. This means we can actually add more tags to the end
+    # with APETAG at some point. (Or at least consider...)
+
+    entries = data[12:12+gd3_length].decode('utf-16').split('\0')
```
lazka: This should use 'utf-16-le'

quodlibet/quodlibet/formats/vgm.py
```diff
@@ -56,6 +79,40 @@ def can_change(self, k=None):
...
+
+    tags["title"] = entries[GD3_ENGLISH_TITLE] \
+        if (entries[GD3_ENGLISH_TITLE] == entries[GD3_JAPANESE_TITLE]) \
+        else \
+        '\n'.join(filter(None, [entries[GD3_ENGLISH_TITLE],
+                                entries[GD3_JAPANESE_TITLE]]))
+
+    tags["artist"] = '\n'.join(filter(None, [entries[GD3_ENGLISH_ARTIST],
```
lazka: What if there is no GD3_ENGLISH_ARTIST or similar in entries.

lazka:
> I do have to enquire about an error I ran into with the unit tests though:
> 
> ```
>     def test_reload(self):
>         self.song["title"] = "foobar"
>         self.song.reload()
> >       self.failUnlessEqual(self.song("title"), "foobar")
> 
> tests/test_formats_vgm.py:22: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> tests/__init__.py:39: in assertEqual
>     super().assertEqual(second, first, msg)
> E   AssertionError: 'foobar' != 'Chaos Emerald'
> E   - foobar
> E   + Chaos Emerald
> ----------------------------- Captured stdout call -----------------------------
> before -- foobar
> after -- foobar
> ```
> 
> I'm very confused by this. I'm actually not even sure why vgm.py seems to be the only format that overrides the reload functionality in the first place. I'm also not sure where in the code path `parse_gd3` is being called in the case that `reload` is being called. If anyone more familiar with the could could help me out, that'd be great.

It's there to allow setting a title which persists even if you reload the file from disk. Other formats save their tags to the file and load them from there again. And the test assumed that the file has not tags to begin with.

You can just update the test to check for the new value.

quodlibet/quodlibet/formats/vgm.py
```diff
@@ -38,24 +56,84 @@ def __init__(self, filename):
...
         if k is None:
             return ["title"]
         else:
             return k == "title"
 
 
+def parse_gd3(data):
+    tags = {}
+    if data[0:4] != b'Gd3 ':
+        print('Invalid Gd3, Missing Header...')
```
lazka: Please use print_d() instead of print()

quodlibet/quodlibet/formats/vgm.py
```diff
@@ -38,24 +56,84 @@ def __init__(self, filename):
...
+    # Should be [0x00, 0x10, 0x00, 0x00] currently.
+    # We should hold onto it for possible branching if standards change.
+
+    gd3_length = struct.unpack('<i', data[8:12])[0]
+    # Length of gd3 footer. This means we can actually add more tags to the end
+    # with APETAG at some point. (Or at least consider...)
+
+    entries = data[12:12 + gd3_length].decode('utf-16-le').split('\0')
+
+    titles = gd3_filter_entries([entries[GD3_ENGLISH_TITLE],
```
lazka: `entries[GD3_ENGLISH_TITLE]` can still raise IndexError

quodlibet/quodlibet/formats/vgm.py
```diff
@@ -38,24 +56,84 @@ def __init__(self, filename):
...
+
+
+def gd3_filter_entries(entries):
+    # First, filter out empty strings...
+    filtered = list(filter(None, entries))
+    if len(filtered) == 0:
+        return filtered
+
+    # Then, filter out any duplicate strings...
+    filtered = list(set(filtered))
```
lazka: There also is quodlibet.util.list_unique() if you care about the original order

quodlibet/quodlibet/formats/vgm.py
TheYokai: Good point. New patch now make sure that the entries are within the bounds of the array for each element along the way.

PR #3101, opened 2019-01-05
Simplify launcher for macOS

a-vrma:
Check-list
----------

No difference when using `time` but it should theoretically be faster

I have dash (a faster posix only shell) for my /bin/sh. It doesn't work with the current launcher script.
While I was there I cleaned some other things up.

I added some changes by consulting shellcheck and tested them on my own machine

- We don't need to cd ... pwd because dirname outputs the path anyway.
- Actually put $bundle_bin to use.
- Change comparisons to be POSIX-compatible, `==` is not going to work when someone has dash as their /bin/sh.

### DIFF
--- a/osx_bundle/misc/bundle/launcher.sh
+++ b/osx_bundle/misc/bundle/launcher.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-bundle="$(dirname "$(dirname "$(dirname "$0")")")"
+bundle=$(cd "$(dirname "$(dirname "$(dirname "$0")")")"; pwd)
 bundle_contents="$bundle"/Contents
 bundle_res="$bundle_contents"/Resources
 bundle_lib="$bundle_res"/lib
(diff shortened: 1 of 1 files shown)


osx_bundle/misc/bundle/launcher.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
```
lazka: I guess it would be a good idea to change this to /bin/bash then? Since we don't test with anything else.

osx_bundle/misc/bundle/launcher.sh
```diff
@@ -1,6 +1,6 @@
...
 
-bundle=$(cd "$(dirname "$(dirname "$(dirname "$0")")")"; pwd)
+bundle="$(dirname "$(dirname "$(dirname "$0")")")"
```
lazka: This is needed to make the path absolute in case a relative path gets passed (relative to the bundle root).

osx_bundle/misc/bundle/launcher.sh
a-vrma: It's entirely up to you. In its current incarnation, the script works as POSIX-compliant, but it could be safer to just use bash.

lazka: Yeah,can you change it while you're at it?

lazka:
Thanks!

PR #3102, opened 2019-01-05
mpris: Drop MPRIS1 support

lazka:
I'd guess that it doesn't have users anymore and everyone switched to MPRIS2 by now.


frestr:
+1. Usage of version 1 also [seems to be discouraged](https://www.freedesktop.org/wiki/Specifications/mpris-spec/).

Issue #293, opened 2015-03-14
Add option to sort by orginaldate, then by date in Album List view


lazka:
Original [issue 293](https://code.google.com/p/quodlibet/issues/detail?id=293) created by hewhocutsdown on 2009-08-28T15:37:43.000Z:

I have to admit, I love the Album List view, but I have one major question:

If an album has both the originaldate and the date tags, is there a way the
albums can sort by originaldate if present and then by date tags? This is
very beneficial with all the deluxe version/re-re-releases of albums that
many have.

With this sort view, you will see albums in the order of original release
if that information is available, and only by the 'date' tag if the
originaldate is absent.

Caveat: for consistency, I would argue that albums should be considered as
whole entities. Therefore, a compilation of 'best of' works, where each
track has a different originaldate set, would use the date tag as it is the
most consistent across tracks.

lazka:
Fixed by #3103

Issue #1486, opened 2015-03-15
Custom ordering options in Album List browser


lazka:
Original [issue 1486](https://code.google.com/p/quodlibet/issues/detail?id=1486) created by danielkza2 on 2014-11-01T19:43:52.000Z:

Right now the ordering options in the Album List browser are quite limited, if not somewhat incomplete.

I'd like to be able to specify a tag string to use a custom sort order. Also, the Date order should take into account the 'originaldate' tag instead of just 'date'.

tkelley65:
Seconded ... this would be great.

dseomn:
I'd like this for the cover grid too. Should I file a separate issue about that, or do both browsers share sort options?

Coffee-9003:
Adding to this, would love to see this feature! I really want to be able to sort the cover grid or album list browsers by a custom tag like I can for song lists, specifically I want to sort albums by ~#lastplayed so that every time I play an album it gets bumped to the top of the browser, but so far I'm unable to find any way to create a sort order for the browser that isn't the default eight options.

Having this one addition would easily make this the best music player I've come across so far, as it's the only feature that I want that quod libet doesn't seem to have.

PR #3103, opened 2019-01-05
Add original date sort option to album browser

### LINKED ISSUE #913: Sort according "Date" better, taking original release date and number of versions to account
Original [issue 913](https://code.google.com/p/quodlibet/issues/detail?id=913) created by jasa.david on 2012-02-10T21:14:50.000Z:

When I have albums of the same artist sorted according to date, I expect that they're sorted according to original release date - as long as there is just one instance of the album.

When there are more instances of the same album (more releases from single release group in MusicBrainz terms), I'd expect oldest release to be sorted using original (oldest) release data and others in place.

Current behavior of taking date tag only into account leads to weird sorting of auto-tagged releases when they're among 1990's and 2000's releases instead of among their older counterparts.

ruuda:
Check-list
----------

 * [x] *There is a linked issue discussing the motivations for this feature or bugfix*
        Fixes #913. It is not quite as generic as #1486, but at least it is a small step forward.
 * [x] *Unit tests have been added where possible*
 * [ ] *I've added / updated documentation for any user-facing features*
        I could not find any documentation this specific, I don't think anything needs to be updated.
        I perceive no difference between sorting by date or original date when browsing my own library of ~1100 albums.

> This change allows sorting by original date in the album list browser. I have many albums where the date tag contains the release date, which might be a CD reissue from the 90s of an album released in the 70s. Previously such albums used to be sorted among albums released in the 90s. Now they are properly sorted by original release date.

### DIFF
--- a/quodlibet/quodlibet/browsers/albums/main.py
+++ b/quodlibet/quodlibet/browsers/albums/main.py
@@ -3,6 +3,7 @@
 #           2012-2018 Nick Boultbee
 #           2009-2014 Christoph Reiter
 #           2018      Uriel Zajaczkovski
+#           2019      Ruud van Asseldonk
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


lazka:
Do you think this also fixes #293?

ruuda:
Yes, it also fixes #293. It takes the original date from the album, and falls back to date if no original date is set.

lazka:
Thanks!

Issue #3110, opened 2019-01-13
Escape characters in filtering query


eauchat:
Steps to reproduce
------------------

This is not exactly a bug, but it's more a bug than a feature I think.

In the [doc](https://quodlibet.readthedocs.io/en/latest/guide/searching.html#searching), it's said that 

> The search terms can’t use quotes ("), slashes (/), hashes (#), pipes (|), ampersands (&), or bangs (!); these characters have special meanings for advanced searches.

This is understandable, but nothing is said about a possibility to escape those characters if necessary to search strings containing those characters.
Imagine that you want to filter all music that is only in the folder: `/Music/folder1/` you can use the following query `filename=folder1`, but imagine there is another folder called: `/Music/folder2/folder1/`, all content from this second folder will also match the query.
It would be great to be able to query `filename=/Music/folder1` for example.
The problematic "/" could be escaped this way for example `filename=\/Music\/folder1`.

Test System
-----------

### Which version of Quod Libet?
3.5.3

### Which operating system 
Ubuntu 16.04

frestr:
It's maybe not very intuitive, but you can escape characters inside regex searches. So your example can be written as `filename=/\/Music\/folder1/`.

eauchat:
Thank you, this is great.
I wrote a micro patch to the doc to explain this, since it's what I was looking for and wasn't finding it in the doc.
I think maybe this issue could be closed if the patch is accepted. It probably wouldn't be very useful to create another way to escape characters since it the intended search can already be achieved.

PR #3112, opened 2019-01-15
Added small explanation on how to escape special characters in search terms

eauchat:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix: #3110 

It wasn't clear in the doc that it's possible to have a search including special characters. This is just to explain that it can be achieved using regex.


eauchat:
Cool :)

PR #3114, opened 2019-01-15
qltk/cover: Always scale down to parent window size

frestr:
Check-list
----------

Always scale the cover image down to the parent window size when clicking it, as opposed to using the monitor size for scaling.

I feel it's more natural that the cover image is contained within the QL main window, rather than potentially covering over the whole window if the window size is small.

This also fixes/removes the slightly confusing magnification difference when left/right clicking.

### DIFF
--- a/quodlibet/quodlibet/qltk/cover.py
+++ b/quodlibet/quodlibet/qltk/cover.py
@@ -228,7 +228,7 @@ class CoverImage(Gtk.EventBox):
         self.__file = None
         self.__current_bci = None
         self.__cancellable = None
-        self._scale = None
+        self._scale = 0.9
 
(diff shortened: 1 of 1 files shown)


frestr:
> My only thought, while we're here, perhaps there should also be an overall upper limit on the size, but I've got a really big main monitor so it's unusually imposing perhaps.

The question is what that limit should be though. I'm also not sure if GTK's `Window.get_size` is HiDPI-aware, so we might need some extra logic for that too.

> I'm not sure I even knew about right-click before, heh.

It was only added a few months ago, so that's forgivable :p

frestr:
Ok, so I'll merge this now, and then we can look at the size limit later.

PR #3126, opened 2019-02-15
Add operon bash completion

elboulangero:
Bash completion for operon.

It's still missing the installation part, as I'm not familiar with QL. At a quick glance, it looks like I should add a module `quodlibet/gdist/bash_completions.py`, following the zsh completion example, am I correct?

Also, this is based on QL 4.2.1-1, I don't know if there was any change in operon in master since then?

Thanks

### DIFF
--- a/quodlibet/data/operon.bash
+++ b/quodlibet/data/operon.bash
@@ -1,0 +1,319 @@
+# operon(1) completion                             -*- shell-script -*-
+# vim: sts=4 sw=4 et
+
+__tags() {
+    operon tags -a -t -c tag | tr '\n' ' '
+}
+
+_in_array() {
+    local i
+    for i in "${@:2}"; do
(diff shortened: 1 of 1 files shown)


lazka:
> It's still missing the installation part, as I'm not familiar with QL. At a quick glance, it looks like I should add a module `quodlibet/gdist/bash_completions.py`, following the zsh completion example, am I correct?

You might also get away with using data_files directly if the filename can stay the same, see https://docs.python.org/3.7/distutils/setupscript.html#installing-additional-files but just copying the zsh way is fine too.

> Also, this is based on QL 4.2.1-1, I don't know if there was any change in operon in master since then?

No changes there :)

lazka:
Thanks!

lazka:
btw, you can do "python3 setup.py install --root foo" to install into foo for testing the install structure.

PR #3128, opened 2019-02-16
Bash completion additional work

elboulangero:
Follow up from yesterday, so this PR finishes the bash completion work!

I added license and copyright to the files, GPLv2 or later to match QL copyright (unless I'm mistaken).


quodlibet/data/operon.bash
```diff
@@ -21,6 +41,9 @@ _operon() {
     local cur prev words cword
     _init_completion -n = || return
 
+    local audio_extensions='@(a?(l)ac|ac3|ape|flac|m[4k]a|mid|mp3|og[ag]|opus|w?(a)v)'
```
elboulangero: From the top of your head, any other extension we should add here?
(for the background, this list is used so that only files with these extensions are shown by the completion)

quodlibet/data/operon.bash
```diff
@@ -21,6 +41,9 @@ _operon() {
...
+    local image_extensions='@(jp?(e)g|png)'
```
elboulangero: Same question here. This list is used to show image files for the `image-set` command of operon

quodlibet/data/operon.bash
lazka:
```
>>> from quodlibet.formats import init, loaders
>>> init()
>>> loaders.keys()
dict_keys(['.dsf', '.aac', '.adif', '.adts', '.wav', '.vgm', '.wma', '.asf', '.wmv', '.wv', '.mp4', '.m4a', '.m4v', '.3gp', '.3g2', '.3gp2', '.mid', '.ape', '.mp3', '.mp2', '.mp1', '.mpg', '.mpeg', '.669', '.amf', '.ams', '.dsm', '.far', '.it', '.med', '.mod', '.mt2', '.mtm', '.okt', '.s3m', '.stm', '.ult', '.gdm', '.xm', '.aif', '.aiff', '.aifc', '.mpc', '.mp+', '.ogg', '.oga', '.flac', '.oggflac', '.spx', '.ogv', '.opus', '.spc', '.tta'])
```

quodlibet/data/operon.bash
lazka: gif comes to mind

quodlibet/data/operon.bash
elboulangero: Ack, thanks! I will rework the completion script to generate this list dynamically, instead of hard-coding it.

quodlibet/data/operon.bash
elboulangero: ok, added

elboulangero:
Hi!

So I improved things a bit. Now the completion scripts generate the list of audio extensions by using the python snippet you provided. The result is cached, so the command is invoked only once per bash session, making it efficient.

Tell me if you don't like it (because it's hitting into QL internals and might break in the future?).

Also, I merged the two completion files for operon and quodlibet into one file, to be able to easily share some code and variables. However, efter doing that, I needed to install a symlink `operon -> quodlibet` in `/usr/share/bash-completion/completions/`. This was not completely straightforward (but not hard either) with QL build system. See the last commit.

If you don't like it, I can rework that, probably revert to having two completion files instead of one.

lazka:
> So I improved things a bit. Now the completion scripts generate the list of audio extensions by using the python snippet you provided. The result is cached, so the command is invoked only once per bash session, making it efficient.
> 
> Tell me if you don't like it (because it's hitting into QL internals and might break in the future?).

Depends on whether you want to maintain/deal with it in the future. It's likely that things will change and I wont notice because I don't use bash (except on Windows, maybe, see below)

> Also, I merged the two completion files for operon and quodlibet into one file, to be able to easily share some code and variables. However, efter doing that, I needed to install a symlink `operon -> quodlibet` in `/usr/share/bash-completion/completions/`. This was not completely straightforward (but not hard either) with QL build system. See the last commit.

Having the same source is fine, but I'd just duplicate it on install as symlinks are a paint there. QL works on Windows with MSYS2 and that uses bash + bash-completion.

elboulangero:
> > Tell me if you don't like it (because it's hitting into QL internals and might break in the future?).
> 
> Depends on whether you want to maintain/deal with it in the future. It's likely that things will change and I wont notice because I don't use bash (except on Windows, maybe, see below)

Another possibility would be to have a "secret option" for QL, like `--list-extensions` or something, that provides the list of supported extensions, either one per line, or space separated (without the leading dash would be best). Would that make it easier to maintain (is there a place where I could add a unit test)?

I don't follow QL development actively, and I use bash completion only with `operon`, so I won't notice either. Although if someone notice and ping me I'll be willing to fix things.

> Having the same source is fine, but I'd just duplicate it on install as symlinks are a paint there. QL works on Windows with MSYS2 and that uses bash + bash-completion.

Ah ok, let me rework that then.

elboulangero:
> Having the same source is fine, but I'd just duplicate it on install as symlinks are a paint there. QL works on Windows with MSYS2 and that uses bash + bash-completion.

I've been asking [bash-completion](https://github.com/scop/bash-completion/issues/281) a bit on that, and they bring to my attention that bash-completion relies on symlinks already. Since it's installed on MSYS2, does that mean that the MSYS2 build system replaces symlinks by copies automatically, or does that mean that symlinks are available on MSYS2? (just being curious)

lazka:
> Another possibility would be to have a "secret option" for QL, like `--list-extensions` or something, that provides the list of supported extensions, either one per line, or space separated (without the leading dash would be best). Would that make it easier to maintain (is there a place where I could add a unit test)?

OK, I'll have a look at adding the option.

> I've been asking [bash-completion](https://github.com/scop/bash-completion/issues/281) a bit on that, and they bring to my attention that bash-completion relies on symlinks already. Since it's installed on MSYS2, does that mean that the MSYS2 build system replaces symlinks by copies automatically, or does that mean that symlinks are available on MSYS2? (just being curious)

bash is built with cygwin which I think just copies things instead of really symlinking. QL gets build with native (somewhat, using mingw) tools and symlinks need admin rights, at least by default.

elboulangero:
> Having the same source is fine, but I'd just duplicate it on install as symlinks are a paint there. QL works on Windows with MSYS2 and that uses bash + bash-completion.

OK, so I reworked the install part, and I now duplicate the files during the install step, no more symlinks.

I also reworked the git history to clean and simplify things, so it should be good enough for a merge now.

> OK, I'll have a look at adding the option.

Thanks!

lazka:
Thanks!

Issue #3129, opened 2019-02-16
Plugins in the ~/.quodlibet/plugins dir do not get detected


petko10:
Steps to reproduce
------------------

Install quodlibet on Manjaro linux (Arch based). Create the dir ~/.quodlibet/plugins . Add a file playlist_server.py with contents:
```
from quodlibet.plugins.playlist import PlaylistPlugin

class PlaylistServer(PlaylistPlugin):
    PLUGIN_ID = "playlist_server"
    PLUGIN_NAME = "Playlist server"
```

Expected Output
---------------
The plugin should apper in the plugins list, but it does not. No errors in the console or plugin errors dialog.

Test System
-----------

frestr:
If your QL config files are in `~/.config/quodlibet` (which they probably are unless you have migrated older configs), you will need to put your plugin there instead (i.e., in `~/.config/quodlibet/plugins`).

The plugin doc is a bit dated in this regard.

petko10:
Actually I tried that beforehand and it didn't work either. Turns out the syntax of the PLUGIN_NAME string assignment from the docs was causing that: `PLUGIN_NAME = _("My Plugin")`

frestr:
Good, thanks for #3133. I'll close this then.

PR #3133, opened 2019-02-17
Update user plugins dir and remove internationalization underscore

petko10:
Defining the PLUGIN_NAME with an underscore produced an error: "NameError: name '_' is not defined". #3129

### DIFF
--- a/quodlibet/docs/development/plugins.rst
+++ b/quodlibet/docs/development/plugins.rst
@@ -5,7 +5,7 @@ Plugin Development
 
 A Quod Libet / Ex Falso Plugin is a simple python module or package which 
 contains sub-classes of special plugin classes and is placed anywhere in 
-``~/.quodlibet/plugins`` so it can be found by QL. These classes can provide 
+``~/.config/quodlibet/plugins`` so it can be found by QL. These classes can provide 
 special methods that get used by QL or can take action on certain events.
 
 At the moment the following plugin classes exist:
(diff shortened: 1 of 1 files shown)


quodlibet/docs/development/plugins.rst
```diff
@@ -66,7 +66,7 @@ Creating a new Plugin
 
     class MyPlugin(EventPlugin):
         PLUGIN_ID = "myplugin"
-        PLUGIN_NAME = _("My Plugin")
+        PLUGIN_NAME = "My Plugin"
```
declension: The thing is, this isn't really the advice. There are currently just over 100 instances of this code `PLUGIN_NAME = (_"`, and _none_ of the uninternationalised one, so I'd suggest we leave this.

If you're having problems with this on _your_ plugin, you probably need to import `gettext` consistently with QL / conventions, see [the `quodlibet` base](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/__init__.py#L21) for a hint.

quodlibet/docs/development/plugins.rst
```diff
@@ -5,7 +5,7 @@ Plugin Development
 
 A Quod Libet / Ex Falso Plugin is a simple python module or package which 
 contains sub-classes of special plugin classes and is placed anywhere in 
-``~/.quodlibet/plugins`` so it can be found by QL. These classes can provide 
+``~/.config/quodlibet/plugins`` so it can be found by QL. These classes can provide
```
declension: Thanks, this is good. The true location is a bit more complex for this, but this is at least more correct than what was there.

quodlibet/docs/development/plugins.rst
lazka: We've removed the "_" function from the builtins some time ago and you now have to import it explicitly: ``from quodlibet import _``. But I guess it makes sense to keep the example simple and not involve translations.

declension:
Agh, too late I guess :slightly_smiling_face:

lazka:
oops :P sorry

PR #3141, opened 2019-03-05
Add new Banshee import plugin

Phidica:
Check-list
----------

 * [x] ~~I've added / updated documentation for any user-facing features.~~ (Not applicable)

Merging this pull request would add a new plugin that imports song information from the database of the [Banshee](http://banshee.fm/) music player, to improve the quality of life for users migrating from Banshee to QL. It was modelled after the pre-existing Rhythmbox importer and works similarly to match the URI stored in the Banshee db to songs QL has added. We import the ratings, play/skip counts, date of most recent play, and date added to library.

The Banshee database is SQLite, so we simply use the standard Python `sqlite3` libary. No new dependencies are added.

I have written and included a test program for the importer.

The plugin is no more complex than the Rhythmbox importer, and, as Banshee has ceased further development, would require very little future maintenance.

### DIFF
--- a/quodlibet/quodlibet/ext/events/bansheeimport.py
+++ b/quodlibet/quodlibet/ext/events/bansheeimport.py
@@ -28,12 +28,61 @@ class BansheeDBImporter:
         self._changed_songs = []
 
     def read(self, db):
-        """Iterate through the library and search for data to import for
-        each song
+        """Iterate through the database and import data for songs found in
+        the library
         """
 
-        for song in self._library:
-            print("filename: %s" % song["~filename"])
+        # use the Row class for extracting rows
+        db.row_factory = sqlite3.Row
+
+        # iterate over all songs in the database
+        for row in db.execute("SELECT * FROM CoreTracks"):
+            try:
+                filename = uri2fsn(row["Uri"])
+            except ValueError:
+                continue
+
+            song = self._library.get(normalize_path(filename))
+            if not song:
+                continue
+
+            has_changed = False
+
+            if row["Rating"] is not None:
+                try:
+                    # banshee stores ratings as integers up to 5
+                    value = row["Rating"] / 5.0
+                except ValueError:
+                    pass
+                else:
+                    song["~#rating"] = value
+                    has_changed = True
+
+            if row["PlayCount"] is not None:
+                # summing play counts would break on multiple imports
+                song["~#playcount"] = row["PlayCount"]
+                has_changed = True
+
+            if row["SkipCount"] is not None:
+                song["~#skipcount"] = row["SkipCount"]
+                has_changed = True
+
+            if row["LastPlayedStamp"] is not None:
+                value = row["LastPlayedStamp"]
+                # keep timestamp if it is newer than what we had
+                if value > song("~#lastplayed", 0):
+                    song["~#lastplayed"] = value
+                    has_changed = True
+
+            if row["DateAddedStamp"] is not None:
+                value = row["DateAddedStamp"]
+                # keep timestamp if it is older than what we had
+                if value < song("~#added", 0):
+                    song["~#added"] = value
+                    has_changed = True
+
+            if has_changed:
+                self._changed_songs.append(song)
 
     def finish(self):
         """Call at the end, also returns number of songs with data imported"""
--- a/quodlibet/quodlibet/ext/events/bansheeimport.py
+++ b/quodlibet/quodlibet/ext/events/bansheeimport.py
@@ -14,11 +14,11 @@ from senf import uri2fsn
 from quodlibet import _
 from quodlibet import app
 from quodlibet import qltk
-from quodlibet import config
+from quodlibet import ngettext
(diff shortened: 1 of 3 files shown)


quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+    try:
+        importer.read(db)
+    except sqlite3.OperationalError:
+        msg = _("Specified Banshee database is malformed or missing")
+        WarningMessage(parent, BansheeImport.PLUGIN_NAME, msg).run()
+    except Exception:
+        util.print_exc()
+        importer.finish()
+        msg = _("Import Failed")
+        # FIXME: don't depend on the plugin class here
```
Phidica: This line is the same in [the Rhythmbox plugin](https://github.com/quodlibet/quodlibet/blob/3c3b5d200c985ae043ee5554a2552035428bfd31/quodlibet/quodlibet/ext/events/rbimport.py#L115). I don't understand what it means

quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+        ErrorMessage(parent, BansheeImport.PLUGIN_NAME, msg).run()
```
Phidica: This line is the same in [the Rhythmbox plugin](https://github.com/quodlibet/quodlibet/blob/3c3b5d200c985ae043ee5554a2552035428bfd31/quodlibet/quodlibet/ext/events/rbimport.py#L115). I don't understand what it means

quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+
+        count = len(self._changed_songs)
+        self._library.changed(self._changed_songs)
+        self._changed_songs = []
+        return count
+
+
+def do_import(parent, library):
+    db_path = expanduser(BansheeImport.USR_PATH)
+    db = sqlite3.connect(db_path)
```
lazka: This can also raise OperationalError for weird paths ("/" for example), I'd move it into the try/except below.

quodlibet/quodlibet/ext/events/bansheeimport.py
lazka: It notes that the do_import() will break if the BansheeImport class gets renamed/changed, which might not be obvious. You can leave it

quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+        WarningMessage(parent, BansheeImport.PLUGIN_NAME, msg).run()
+    except Exception:
+        util.print_exc()
+        importer.finish()
+        msg = _("Import Failed")
+        # FIXME: don't depend on the plugin class here
+        ErrorMessage(parent, BansheeImport.PLUGIN_NAME, msg).run()
+    else:
+        count = importer.finish()
+        preamble = "Successfully imported ratings and statistics for "
```
lazka: This wont get translated this way, you have to include it in ngettext both times.

quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class BansheeImport(EventPlugin):
+    PLUGIN_ID = "bansheeimport"
+    PLUGIN_NAME = _("Banshee Import")
+    PLUGIN_DESC = _("Imports ratings and song statistics from Banshee.")
+    PLUGIN_ICON = Icons.DOCUMENT_OPEN
+
+    DEF_PATH = "~/.config/banshee-1/banshee.db"
+    USR_PATH = DEF_PATH
```
lazka: Why would the DB be in another place?

quodlibet/quodlibet/ext/events/bansheeimport.py
```diff
@@ -0,0 +1,176 @@
...
+        label = Gtk.Label(label=_("_Database path:"), use_underline=True)
+        label.set_alignment(0.0, 0.5)
+        grid.attach(label, 0, 0, 1, 1)
+
+        entry = UndoEntry()
+        entry.set_hexpand(True)
+        entry.set_text(BansheeImport.DEF_PATH)
+
+        def path_activate(entry, *args):
+            path = entry.get_text()
```
lazka: needs `path = text2fsn(entry.get_text())`

quodlibet/quodlibet/ext/events/bansheeimport.py
Phidica: The directory depends on the XDG settings of the user, though it defaults to here. We could instead read the XDG settings and infer the location that way, but perhaps it could have changed. I don't see a reason *not* to leave this as an option to the user

Phidica:
Any more comments on this?

PR #3151, opened 2019-03-23
player: Make external volume cubic by default

### LINKED ISSUE #3125: Volume % Difference in GUI vs Terminal
Might not be a "bug" so to say, but there's a discontinuity in the way volume % is calculated in the user interface vs in the terminal.

The user interface seems to be perceptually exponential. Going from 5-50% is as going from silent to slightly less silent, with almost all 'real' volume control being in the 60/70-100% range depending on system loudness. Applies to slider bar, too.

Terminal commands process/display volume like most other things, á la Pulse/ALSA, most web controllers and other music players, etc.

volume

frestr:
Check-list
----------

Different parts of the codebase have gradually been changed from using a linear volume scale to a cubic one instead (see 9217dfb5d117dfdf5013cd84028db0d5e703a93d, 44aa244408aef31236a22e145dfc31565b14ca06, 1f3bb3280b93d1715469c8fb608f380a93128f3d, #3098)

This commit introduces some consistency by making a cubic volume the default. This will also fix the CLI and some plugins erroneously modifying the internal/linear volume directly.

Fixes #3125

There is a slight risk of breakage when changing the volume semantic like this, but it seems like the remaining parts of the codebase using `player.volume`  (that are not touched by this commit) are plugins that also should use the cubic volume interface.

It seems to work fine with the testing I've done, but I might have overlooked something, so feedback is welcome.

### DIFF
--- a/quodlibet/quodlibet/ext/events/mpdserver/main.py
+++ b/quodlibet/quodlibet/ext/events/mpdserver/main.py
@@ -270,7 +270,7 @@ class MPDService(object):
     def setvol(self, value):
         """value: 0..100"""
 
-        self._app.player.volume_cubic = value / 100.0
+        self._app.player.volume = value / 100.0
 
     def repeat(self, value):
         self._options.repeat = value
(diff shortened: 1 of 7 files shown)


lazka:
lgtm

jplatte:
I was so confused when my music wasn't audible anymore despite having the volume set to 50% with a custom script / desktop integration. Good to know I can simplify that script now by dropping the cubic <> linear conversion!

Issue #3149, opened 2019-03-20
Scrolling down on time display seeks forward, not backward as expected


xhc1:
Steps to reproduce:

1. start play
2. scroll with mouse on waveform
     then
3. scroll with mouse on basic display time 
------------------

Actual Output:

when scrolling **up**:
on waveform: it goes **forward** in seek
BUT on display time: it goes backward in seek

when scrolling **down**:
on waveform: it goes **backward** in seek
BUT on display time: it goes forward in seek

Which version of Quod Libet?
>  4.2.1

Which operating system 
> fedora 29

PR #3161, opened 2019-04-05
Fix waveform scroll  #3149

Koci0:
Fixed inconsistent scrolling on waveform seekbar as described in #3149 . Now scrolling DOWN on waveform seekbar seeks forward and scrolling UP seeks backwards. The behaviour is the same as scrolling on basic display time.
Another way to fix that would be to change scrolling behaviour of basic display time but (to me) scrolling down to seek forward seems more natural.


frestr:
Thanks for the PR.

It seems like it's more common that scrolling down seeks backward -- at least that's the behavior in the other players I've tested (vlc, mpv, rhytmbox, clementine). So I think it would make more sense to change the time display instead.

Issue #927, opened 2015-03-14
Store/retrieve song ratings from Musicbrainz


lazka:
Original [issue 927](https://code.google.com/p/quodlibet/issues/detail?id=927) created by hawke@hawkesnest.net on 2012-03-07T18:42:23.000Z:

Musicbrainz allows users to store their ratings for songs. It would be nice if Quod Libet would be able to use (store and retrieve) this information.

Details: http://musicbrainz.org/doc/XML_Web_Service/Version_2#ratings

I would expect that it would:
1. on song load or program launch, if the track has a recording MBID, lookup the user’s rating for it. User’s MB rating should override QL’s rating.
2. When the rating changes, it should be sent back to MB.

Things to watch out for: 
- Musicbrainz ratings scale 0–100
- Some submission queuing may be in order, especially if the MB service is unavailable (e.g. no internet connection).

LoveIsGrief:
Hi, this is a feature I would like to work on. I assume the best way to go about this would be with a plugin, but I would like to make sure:

 - should I write a new plugin?
 - should I modify the current plugin?
 - would you rather have this a basic functionality of Quod Libet?

Cheers

declension:
Hi

A new plugin is probably the quickest way to get this done. There are various ways to do HTTP / API calls internally, none is ideal. Perhaps have a look at the requests usage, or the async GIO versions (e.g. in Soundcloud browser) 

If there's any real overlap with existing plugins (I can't see much currently), or if #963 happens then it can be refactored / extended accordingly, but that hasn't happened for ages so...

LoveIsGrief:
Thanks for the response.

I think I'd then start with #963, since I don't really like working with UI stuff. If I do, I'd most likely copy code from the example rating plugin that adds its own rating column, but I honestly am against. It's quick, but dirty.

bassstorm:
@declension
PR is merged, but the ticket is open for a reason? Sorry, just trying to help with clean up :)

declension:
Yes - thanks, didn't have correct trigger text I guess. Closing...

PR #3180, opened 2019-05-23
First version of Musicbrainz Sync plugin that sends ratings

LoveIsGrief:
Check-list
----------

         New plugin so no unit tests
         There doesn't seem to be documentation on the existing plugins
         ~~Lag introduced when rating and the plugin is activated~~ Background task fixed that

For now all the plugins does is send ratings and doesn't retrieve them. The submission
 is not a fast connection so there's a lag between the click 
 and the star rating being updated in the GUI.
That would surely be solved with an async call. For a v0.2 that might be good?

Not sure what to do with errors. For now nothing is done and a dialog
 will popup warning the user of the occurred error.

TODO:

 - [ ] Error handling?
 - [x] Async submission of rating?

quodlibet/quodlibet#927 Store/retrieve song ratings from Musicbrainz


LoveIsGrief:
Running `./setup.py update_po` as described [in the guide](https://quodlibet.readthedocs.io/en/latest/translation/howto.html#translating) makes many more changes than I made, mostly with the paths  (removes `../`). I assume the script hasn't been run in a while?

Therefore, I don't know what to do about the translations other than delete them.

LoveIsGrief:
I'm unable to use Github's "request review" feature, so I hope it's OK that I tag @declension and @lazka here for that :slightly_smiling_face:

quodlibet/ext/events/musicbrainzsync.py
```diff
@@ -0,0 +1,115 @@
...
+            BRAINZ_APP, VERSION, "loveisgrief@tuta.io"
+        )
+        musicbrainzngs.auth(
+            plugin_config.get("username"),
+            plugin_config.get("password")
+        )
+
+    def plugin_on_changed(self, songs):
+        ratings_dict = {
+            song[ATTR_BRAINZ]: int(song[ATTR_RATING] * 100)
```
declension, line 52: Hmm, I didn't think `[]` for internal tags worked actually, but clearly I've forgotten. Nonetheless normally we use call semantics i.e. `song(ATTR_RATING)` as more processing sometimes happens, plus nobody likes `KeyError` :) )

quodlibet/ext/events/musicbrainzsync.py
```diff
@@ -0,0 +1,115 @@
...
+            for song in songs
+            if ATTR_BRAINZ in song and ATTR_RATING in song
+        }
+        if len(ratings_dict) > 0:
```
declension, line 56:
```suggestion
        if ratings_dict:
```
perhaps?

quodlibet/ext/events/musicbrainzsync.py
```diff
@@ -0,0 +1,115 @@
...
+            call_async_background(
+                musicbrainzngs.submit_ratings,
+                Cancellable(),
+                callback=lambda *args: (),
```
declension, line 60: I guess `None` doesn't work here?

quodlibet/ext/events/musicbrainzsync.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class MusicBrainzSyncPlugin(EventPlugin):
+    PLUGIN_ID = "musicbrainzsync"
+    VERSION = VERSION
+    PLUGIN_NAME = _("MusicBrainz Sync")
+    PLUGIN_DESC = _("Syncs the rating of a song with music brainz.")
+
+    def __init__(self):
+        super(MusicBrainzSyncPlugin, self).__init__()
```
declension, line 40:
```suggestion
        super().__init__()
```

quodlibet/ext/events/musicbrainzsync.py
```diff
@@ -0,0 +1,115 @@
...
+        musicbrainzngs.set_rate_limit()
+        musicbrainzngs.set_useragent(
+            BRAINZ_APP, VERSION, "loveisgrief@tuta.io"
```
declension, line 43: Might that want to be the user's email, or the default QL one (we use this for ratings actually)

declension:
Generally LGTM! The msys2 is a known problem now and currently disabled, so I think if you rebase / merge from master you'll get a green build here

quodlibet/ext/events/musicbrainzsync.py
LoveIsGrief: Unfortunately not because of https://github.com/quodlibet/quodlibet/blob/4b8a99f0dde0ee87d38739edb228d37ce93f8665/quodlibet/util/thread.py#L104

quodlibet/ext/events/musicbrainzsync.py
LoveIsGrief: `KeyError` shouldn't happen since it's in a loop that checks if the attribute is in the song. I applied your suggestion anyway :slightly_smiling_face:

quodlibet/ext/events/musicbrainzsync.py
LoveIsGrief: Applied

quodlibet/ext/events/musicbrainzsync.py
LoveIsGrief: Opted for the default QL on.

declension:
Thanks!

LoveIsGrief:
Thank you too! Contributing was a nice experience :+1:

Issue #2158, opened 2016-12-21
No space before and after "/" in my language


scootergrisen:
I see the source strings for translations uses " / " with space before and after the "/".
In my language i would prefer to have no space.
I can do this for some strings in the file but i see more of this in the software like:
"Title / Version"
"Album / Disc Subtitle"

In my language i would like:
"Titel/version"
"Album/disk-undertitel"

lazka:
That would need to be changed for all languages. Why do you need it?

scootergrisen:
I think without space is more correct in my language and how we translate in other software in my language.
Perhaps you can add a string like "%s / %s" in the translation and i can translate that to "%s/%s" in my language and use that when you need to join strings.

scootergrisen:
Strange the readme use different style:
`Ex Falso / Quod Libet - A Music Library/Editor/Player`
Maybe that is perfectly correct i don't know i just like to use the same style when i translate.

declension:
Yeah I'd say we should fix that, the slashes are a bit abused IMO, or at least: confusing...

lazka:
The ` / ` separator is used for joining tied tag previews. Their values get joined with `, `. Would `, ` make more sense?

scootergrisen:
I don't know i haven't used the software enough to know.

PR #3192, opened 2019-06-22
Minor QuodLibet / ExFalso subtitle change (see #2158)

CreamyCookie:
Minor QuodLibet / ExFalso subtitle change (see #2158)

### DIFF
--- a/quodlibet/README
+++ b/quodlibet/README
@@ -1,5 +1,5 @@
-Ex Falso / Quod Libet - A Music Library/Editor/Player
-=====================================================
+Ex Falso / Quod Libet - A Music Library / Editor / Player
+=========================================================
 
 Quod Libet is a music management program. It provides several different
 ways to view your audio library, as well as support for Internet radio and
--- a/quodlibet/docs/index.rst
+++ b/quodlibet/docs/index.rst
@@ -1,4 +1,4 @@
-.. title:: A Music Library/Editor/Player
+.. title:: A Music Library / Editor / Player
 
 .. image:: images/paned.png
     :align: right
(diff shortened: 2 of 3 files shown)


CreamyCookie:
Weird, I didn't change the line endings. Maybe we should add a [`.gitattributes` file](https://help.github.com/en/articles/dealing-with-line-endings#per-repository-settings) to make sure consistent line endings are used (and files like *.nist keep crlf)

CreamyCookie:
I'm out of ideas how to fix the line endings. (Have tried `git add --renormalize`, copying the CRFL version into the editor here etc.) Any ideas?

lazka:
Not sure. Make sure "core.autocrlf" isn't enabled (it's not the default) and that your editor doesn't change line endings.

CreamyCookie:
core.autocrlf was set to input but supposedly [`.gitattributes` overrides an individual's core.autocrlf](https://help.github.com/en/articles/dealing-with-line-endings)

I've also tried from the command line (while IDE is closed). I've checked the file has CRLF and Git responds that there is nothing to commit, so CRLF is used.

I have no created a fresh clone (with my branch selected) and CRLF is used. Seems like Git's diff viewer is bugged. (sigh..)

Can this be merged?

lazka:
> I've also tried from the command line (while IDE is closed). I've checked the file has CRLF and Git responds that there is nothing to commit, so CRLF is used.

I think .gitattributes will change things when you check them out, so you will not see what's in the actual repo.

Anyway, I've squashed things in the branch. Feel free to hit the merge button if it still looks ok to you.

CreamyCookie:
Looks good. Thanks!

PR #3196, opened 2019-06-23
Closes #3179: Tracks without track number are now sorted before first track (Same for disk numbers)

### LINKED ISSUE #3179: Bug sorting tracks without tracknumbers
Steps to reproduce
------------------
Try sorting an album containing a mix of songs with/without tracknumbers by Artist or Album.

Expected Output
---------------
I would expect all songs in an album without tracknumbers to be sorted before all songs in the album with tracknumberd.

Actual Output
-------------
(See first screenshot below) The screenshot shows the ordering with a custom sort order, but the result is the same sorting on Album or Artist. It seems like the lack of a tracknumber is being considered equivalent to a tracknumber of 1 for the purposes of sorting, so the ordering of the tracks falls back to being determined by title instead. Changing the title of the tracknumber-less track to one alphabetically before that of the one with a tracknumber causes it to be sorted first.

Test System
-----------
4.2.0, but it doesn't appear to have been fixed in 4.2.1.

### Which operating system 
Linux Mint 18.3

Additional Information
----------------------
[The incorrect sort order](https://i.imgur.com/Tzbxlqp.jpg)

As you can see, there is no extraneous metadata to muck things up; the tracks match on everything except title and tracknumber (or lack thereof).
[Metadata 1](https://i.imgur.com/inPzv4u.png)
[Metadata 2](https://i.imgur.com/31gEOrp.png)

CreamyCookie:
before:
before fix

now:
before after fix

### DIFF
--- a/quodlibet/quodlibet/formats/_audio.py
+++ b/quodlibet/quodlibet/formats/_audio.py
@@ -136,7 +136,7 @@ class AudioFile(dict, ImageContainer):
             self[key] = value
 
     def __song_key(self):
-        return (self("~#disc", 1), self("~#track", 1),
+        return (self("~#disc", 0), self("~#track", 0),
             human(self("artistsort")),
             self.get("musicbrainz_artistid", ""),
             human(self.get("title", "")),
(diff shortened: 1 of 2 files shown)


CreamyCookie:
I guess this is small enough to be merged without code review.. though in general I think it's better to get one look from someone who knows more about any possible side effects that might have to be addressed. Also, code reviews are pretty cool / useful.

PR #3205, opened 2019-07-04
Cutom Commands: Make description row translatable

Mellthas:
Make the description row in the custom-commands window translatable. The word _Command_ was not translatable, as can be seen in the attached screenshot:

custom-commands

I think it would be even nicer to remove the word _Command_ from that row altogether, and just display the actual command there. It should be clear from the context what it is. So it would be, e.g.:

**Compress files**
file-roller -d <~filename>

instead of:

**Compress files**
Command: "file-roller -d <~filename>"

I can do another PR, if this is desired.

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
+++ b/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
@@ -126,7 +126,7 @@ class Command(JSONObject):
                 or "playlistindex" in self.pattern)
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


CreamyCookie:
@Mellthas Yeah, I think removing is the better option. Also less confusing if there are quotes in the command.

declension:
FYI I think when I first did the UI, I had several fields in the display not just command, but it seemed neater without. I agree that removing entirely would look better now, so let's do that then @Mellthas  (thanks)

PR #3207, opened 2019-07-04
Small string fixes

Mellthas:
Fixes incomplete parentheses and spaces missing between sentences in tooltips in the MQTT Publisher plugin.

For the spaces: Is the fix done like this okay, or would it be better to combine the tooltip strings into one for each (duplicating the _Accepts QL patterns …_ text)?


declension:
Thanks!

PR #3208, opened 2019-07-04
Consistent simple tooltips wrt. period at the end

Mellthas:
Fix some instances of simple tooltips ending in a period. They shouldn't.

Regarding tooltips using multiple sentences, see #3209.


declension:
Thanks!

CreamyCookie:
@Mellthas I think the `.po` files should be updated too, since otherwise the translations would not be found (which would be a shame and possibly cause translators to redo their work).. unless I'm missing something and this isn't how it works.

https://github.com/quodlibet/quodlibet/search?l=Gettext+Catalog&q=Library+directory+the+server+connects+to

Anyway, thanks for the commits!

Mellthas:
Good point. Yes, the changed strings will appear untranslated until the `po` file is updated by a translator. PR incoming.

PR #3211, opened 2019-07-05
Restarting

### LINKED ISSUE #2376: Add restart functionality
The crash reporter has a button for exiting the program in a "controlled" manner. What we'd actually want there is a button to restart the program.

I guess something like registering an atexit handler which execv itself.

blimmo:
Fixes #2376. In addition to adding a restart button on the crash reporter I've also added a restart button to the file menu above Quit.

On windows this solution has the annoying behaviour of the old process appearing to exit and the new process just using the same console. What I've read seems to suggest that on unix it'll replace the process nicely though and there doesn't seem to be a good way of doing that on windows.

I think what I've got for communicating when to restart is ok though I'm open to suggestions if anyone has anything they'd prefer.

### DIFF
--- a/quodlibet/quodlibet/_main.py
+++ b/quodlibet/quodlibet/_main.py
@@ -59,6 +59,9 @@ class Application(object):
     is_quitting = False
     """True after quit() is called at least once"""
 
+    restart = False
+    """True if the program should restart after quitting"""
+
     @property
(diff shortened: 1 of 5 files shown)


lazka:
Thanks, I made some changes in a followup commit: 62cdd6a877435103

PR #3215, opened 2019-07-08
Adapt po files after tooltip punctuation fixes in #3208

Mellthas:
See https://github.com/quodlibet/quodlibet/pull/3208#issuecomment-509030661.

For each instance of the tooltip strings changed in #3208, I updated the source and the translation string (made their ending punctuation match). This way, the strings will appear translated again after `build_mo`. Some languages had none of the strings translated, so they weren’t touched.


CreamyCookie:
Thanks!

Issue #3209, opened 2019-07-04
Should multi-sentence tooltips end in a period?


Mellthas:
Most do, some don't. This is especially apparent in the prefs window on master, see the attached screenshots:

![prefs-tooltip-1](https://user-images.githubusercontent.com/5852189/60688069-2cd4f800-9eb3-11e9-9a3f-32274abf1e12.png)
![prefs-tooltip-2](https://user-images.githubusercontent.com/5852189/60688070-2cd4f800-9eb3-11e9-8c30-f1473f9bb517.png)

Both tooltips are worded similarly. The first one does not end in a period, the second does. It would be nice to be consistent here, as it is also confusing when translating those strings.

declension:
Thanks for the report. I think yes they should (and some are my fault I imagine, sorry)

Mellthas:
I think my preference would be slightly in favor of no ending period (coming from my German language intuition), but either way looks good to me.

I created #3216 which adds an ending period to all the multi-sentence tooltips I could find which have none.

Mellthas:
There are now two competing PRs: #3217 and #3218. The latter one fixes just the tooltip discussed here.

CreamyCookie:
#3218 is merged now. Are there any more multi-sentence cases that need to be addressed or should we close this?

Mellthas:
Not that I am aware of. This was mostly about what to do in general anyway. I think we can close.

PR #3216, opened 2019-07-09
Consistent ending punctuation for multi-sentence tooltips

Mellthas:
See discussion in #3209. The second commit tries to retain existing translations, similar to #3215.


Mellthas:
I did a forced push to update the commits, but the old ones are still shown here. I'll close and open a new PR.

Issue #1509, opened 2015-03-15
Test "written style"


lazka:
Original [issue 1509](https://code.google.com/p/quodlibet/issues/detail?id=1509) created by engored on 2014-12-25T15:15:28.000Z:

I was just checking if the usual way to end PLUGIN_DESC is with or without period ('.') and I got a crazy idea about having an automated test that checks some basic style and punctuation usage.

Some current data:
$ grep -Pzo 'PLUGIN_DESC = _([^)]+.\W)' -rl quodlibet/ext/ | wc -l
says 46 plugins have their descriptions end with a dot and
grep -Pzo 'PLUGIN_DESC = _([^)]+[^.]\W)' -rl quodlibet/ext/ | wc -l
says 23 plugins - no dot.
(The results are approximate, these regexes are too primitive for parsing python strings)

I think it would be nice to have the UI look solid and plugins window is a low hanging fruit. Two problems though:
1. Obviously changing old messages will lead to missing translations and translators having to do extra work (trivial, but still). Is this a good idea to do this just for this case with plugin description? If not, how about more stuff (see question 2)? When is a good time to do such changes (e.g. after major release)?
2. It seems there are test_po.py and code quality tests, but no user-facing strings tests. So if I were to start writing such tests, where would I start? In addition to plugin descriptions always ending with dot, they would also check for extra/missing whitespace, capitalization of certain terms, etc. I guess they can be optional, just like pyflakes and pep8 tests (so the test suite wouldn't break on simple typos), but can be used once in a while to discover and fix typos and such in a batch. Any ideas?

PR #3217, opened 2019-07-09
Consistent ending punctuation for multi-sentence tooltips

Mellthas:
See discussion in #3209. The second commit tries to retain existing translations, similar to #3215.

### DIFF
--- a/quodlibet/po/cs.po
+++ b/quodlibet/po/cs.po
@@ -1789,8 +1789,8 @@ msgstr "Nastavení"
 
 #: ../quodlibet/ext/events/mqtt.py:51
 #, python-format
-msgid "Accepts QL Patterns e.g. %s"
-msgstr "Akceptuje QL vzory, např. %s"
+msgid "Accepts QL Patterns e.g. %s."
+msgstr "Akceptuje QL vzory, např. %s."
 
 #: ../quodlibet/ext/events/mqtt.py:57
 msgid "MQTT Publisher"
@@ -2284,7 +2284,7 @@ msgstr "Přehrává se:"
 
 #: ../quodlibet/ext/events/telepathy_status.py:124
 #, python-format
-msgid "Status text when a song is started. Accepts QL Patterns e.g. %s"
+msgid "Status text when a song is started. Accepts QL Patterns e.g. %s."
 msgstr ""
 "Text statusu při začátku přehrávání. Je možné použít QL vzory jako např. %s."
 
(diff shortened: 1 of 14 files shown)


Mellthas:
`test_po.py` now fails for `ru.po`, because the translation does not end in a period (`test_original_punctuation_present`). It is only a single sentence though, so I think a legitimate difference from the source string.

CreamyCookie:
I think lines ending with a pattern is an edge case where I personally prefer no period since it looks a bit too noisy. (And a new user might wonder if the dot is part of the pattern.)

`Accepts QL Patterns e.g. <~artist~title>` 
`Accepts QL Patterns e.g. <~artist~title>.`

Also, in the above case isn't that change inconsistent with the previous one to remove periods for simple texts (#3208)?

> test_po.py now fails for ru.po, because the translation does not end in a period

Shouldn't it not also end in a period?

Mellthas:
@CreamyCookie, that is a point, I thought about that, too. Maybe better to just leave these alone. Then the changes would boil down to just the one tooltip in `prefs.py`. I’ll open a competing PR :-).

> Also, in the above case isn't that change inconsistent with the previous one to remove periods for simple texts (#3208)?

No, as this string is only used in combination with two tooltip strings (see line 137 and 141):

```
_("Status text when a song is started.") + ' ' + _ACCEPTS_PATTERNS
_("Text when a song is paused.") + ' ' + _ACCEPTS_PATTERNS
```

Mellthas:
> Shouldn't the translation not also end in a period?

The translation in question is:

`Список разделителей (через пробел), которые будут использоваться для разделения тегов`

which Google translates back to:

`List of delimiters (separated by spaces) that will be used to separate tags`

Which is a bit shorter than the original, and wouldn't use a period at the end (right?):

`A set of separators to use when splitting tag values in the tag editor. The list is space-separated.`

declension:
@Mellthas That Russian sounds like a direct translation of the old version of that string (IIRC)

Mellthas:
I added another commit which reverts the change to the Russian PO file, to make the tests pass.

Mellthas:
For the future, it might be good to have a look at the PO tests. The ending-punctuation test (`test_original_punctuation_present`) is only run for the Russian and German PO files. I looked up the origins of the test, #1509 has some background on the difficulty of enabling it for more languages. It might make sense to make these checks optional, but allow them to be run for any language. Using poedit, I get a warning for an ending punctuation mismatch, but I can decide to ignore that. The checks are more comprehensive, so e.g. it also warns on an ending period in the translation if the source string does not have one.

If I find some time, I can take a look at that.

CreamyCookie:
> It might make sense to make these checks optional, but allow them to be run for any language. 

@Mellthas I don't know enough about the translation tools, but that sounds like a good idea. Anyway, I'm closing this. (The alternative was merged.)

Mellthas:
Thanks. Just for the reference, Translate Toolkit (https://github.com/translate/translate) looks quite promising. It has the ability to check PO files with a multitude of filters (e.g. http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/pofilter_tests.html#endpunc for ending punctuation).

PR #3218, opened 2019-07-10
Prefs: Fix ending punctuation in tooltip

Mellthas:
Alternative to #3217, only covering the one tooltip in the prefs window.

### DIFF
--- a/quodlibet/po/he.po
+++ b/quodlibet/po/he.po
@@ -5296,7 +5296,7 @@ msgstr "שמירת שינויים בתגים ללא בקשת אישור בעת �
 #: quodlibet/qltk/prefs.py:608
 msgid ""
 "A set of separators to use when splitting tag values in the tag editor. The list "
-"is space-separated"
+"is space-separated."
 msgstr ""
 "סדרת תווי הפרדה לשימוש בעת פיצול ערכים בעורך התגים. ערכי הרשימה מופרדים בפסיקים"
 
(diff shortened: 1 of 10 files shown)


quodlibet/po/he.po
```diff
@@ -5296,7 +5296,7 @@ msgstr "שמירת שינויים בתגים ללא בקשת אישור בעת 
 #: quodlibet/qltk/prefs.py:608
 msgid ""
 "A set of separators to use when splitting tag values in the tag editor. The list "
-"is space-separated"
+"is space-separated."
 msgstr ""
 "סדרת תווי הפרדה לשימוש בעת פיצול ערכים בעורך התגים. ערכי הרשימה מופרדים בפסיקים"
```
CreamyCookie: Looks good, though I think this (translated) line should also end in a period.

Mellthas: Probably yes, but as Hebrew is right-to-left, I didn't want to mess up the translation. I think it is safer to leave as-is.

CreamyCookie: Oh.. good point. Totally didn't think of that.

CreamyCookie:
Thanks!

quodlibet/po/he.po
lazka: Best to just leave the original string as is in the future, so it will be marked as "fuzzy" by gettext and the translator sees that the string needs an update

PR #3219, opened 2019-07-10
Fix punctuation of another simple tooltip

Mellthas:
Missed in #3208.

### DIFF
--- a/quodlibet/po/da.po
+++ b/quodlibet/po/da.po
@@ -4641,7 +4641,7 @@ msgid "Tag expression"
 msgstr "Tag-udtryk"
 
 #: ../quodlibet/qltk/data_editors.py:373
-msgid "Tag expression e.g. people:real or ~album~year."
+msgid "Tag expression e.g. people:real or ~album~year"
 msgstr "Tag-udtryk. F.eks. people:real eller ~album~year."
 
 #: ../quodlibet/qltk/data_editors.py:374
--- a/quodlibet/po/pt.po
+++ b/quodlibet/po/pt.po
@@ -4726,7 +4726,7 @@ msgid "Tag expression"
 msgstr "Expressão de etiqueta"
 
 #: quodlibet/qltk/data_editors.py:371
-msgid "Tag expression e.g. people:real or ~album~year."
+msgid "Tag expression e.g. people:real or ~album~year"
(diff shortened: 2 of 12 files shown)


declension:
LGTM

quodlibet/po/da.po
```diff
@@ -4641,7 +4641,7 @@ msgid "Tag expression"
 msgstr "Tag-udtryk"
 
 #: ../quodlibet/qltk/data_editors.py:373
-msgid "Tag expression e.g. people:real or ~album~year."
+msgid "Tag expression e.g. people:real or ~album~year"
 msgstr "Tag-udtryk. F.eks. people:real eller ~album~year."
```
CreamyCookie: I think the period was missed

quodlibet/po/pt.po
```diff
@@ -4726,7 +4726,7 @@ msgid "Tag expression"
 msgstr "Expressão de etiqueta"
 
 #: quodlibet/qltk/data_editors.py:371
-msgid "Tag expression e.g. people:real or ~album~year."
+msgid "Tag expression e.g. people:real or ~album~year"
 msgstr "Expressão de etiqueta. Por exemplo, people:real or ~album~year."
```
CreamyCookie: another one in "Expressão de etiqueta..." line

quodlibet/po/da.po
Mellthas: It looks like two sentences in the translation, so the translator might want to retain the ending punctuation.

quodlibet/po/pt.po
Mellthas: See above

quodlibet/po/da.po
Mellthas: But given all the uncertainties that arise, it might indeed be better in the future to not touch the PO files, and let the translators decide what to do; see https://github.com/quodlibet/quodlibet/pull/3218#discussion_r302978599.

CreamyCookie: Ah, I see, the period after udtryk is surprising. 

> touch the PO files, and let the translators decide what to do; see #3218 (comment).

True. I wasn't aware about the fuzzy matching thing. That said, it still requires extra work from the translators and given that that work might only happen very rarely, I think updating in the very obvious cases still makes sense. But yeah, that's not the case here.

CreamyCookie:
Thanks!

Issue #2156, opened 2016-12-21
Perhaps move Advanced Preferences and Change Language to Preferences


scootergrisen:
In Plugins i see two plugins that can't be disabled:
Advanced Preferences
Change Language

Perhaps better to have these things under File > Preferences.
Also the string "Change Language" does not appear to be translatable.

lazka:
The advanced preferences plugin contains things which 99% of users should not change. Think "about:config" in firefox or similar. Not sure about the "Change language" plugin.. We've only had _one_ request for it over the time. If it turns out to be desired by more people we can move it.

>  Also the string "Change Language" does not appear to be translatable.

Did you run `./setup.py update_po`. That string was just added yesterday.

scootergrisen:
Ok that might be way. No i have not run the update yet. I will try when i have translated all the strings.

declension:
Are we OK to close this issue then?

scootergrisen:
I just don't see why there are two plugins that can't be disabled.
Are the thing about plugins not that they can be enabled/disabled (installed/removed)?
If they are on all the time and not possible to remove then i don't think its correct to call it a plugin.

I would move those things under File > Preferences.
But if you don't agree then yes this can be closed.

CreamyCookie:
Hmm.. I kinda see the issue. It seems a bit weird that there's an always enabled plugin that's used for rare setting. I guess it would be possible to add an Advanced tab to the Preferences, which contains a button with a warning and a button (like in `about:config`): <kbd>I accept the risk</kbd>

It could then also include the language settings. (Removing it from the plugins would also have the minor advantage of making some space. Although there are other ideas for making it look less crowded.)

declension:
I've updated the prefs themselves in 35060ccf3f5434d23cf53c87aec99fa88ff15011 and during / before / now, the more I use it, the more it seems that it belongs in the main prefs hidden perhaps like @CreamyCookie suggests (remembering Quod Libet's meaning... :wink: ), and like other apps often do. 

Any more thoughts on this?

PR #3221, opened 2019-07-14
Ignoring characters in queries.

### LINKED ISSUE #3198: Ignore the character "-" on search
The proposed change
-------------------
On any search input fields ignore the character "-" during search.

Benefits of this change
-----------------------
**1.** 
Most if not any music application I know ignores the "-" char during search. This makes it possible to search a song like this: `<Artist> - <Songname>`. 

In my case I am copy pasting very often. Therefore I need to manually remove the "-" sign every single time, which starts to annoy me enough to write this feature request.

**2.** 
The proposed change also prevents users from getting search results not matching the expectation.

Lets say I search `<Artist> - <Songname>`.

The results are:
`<Artist>, <Songname> - Radio Edit`
`<Artist>, <Songname> - Club Edit`

I get all results containing "-" somewhere in the text itself. 

But I wanted to find the original song. This is annoying since it requires another search. I have a huge library, this every time takes a few seconds.

Any issues this might cause
---------------------------
Don't see any.
Usability isn't decreased since this is the expected behaviour in my opinion.

blimmo:
Closes #3198 

I've erred on the side of caution and only done the ignoring on text queries.

I've also gone for a default of no ignoring since I think it's easier to notice that you want to ignore a character than to notice you don't (after having a query inexplicably not work for example).

### DIFF
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -112,6 +112,9 @@ INITIAL = {
         # "global" filter for SearchBar
         "background": "",
 
+        # characters ignored in queries
+        "ignored_characters": "",
+
         # album list
         "albums": "",
 
--- a/quodlibet/quodlibet/config.py
+++ b/quodlibet/quodlibet/config.py
@@ -1,7 +1,7 @@
 # Copyright 2004-2008 Joe Wreschnig
 #           2009-2017 Nick Boultbee
 #           2011-2014 Christoph Reiter
-#                2018 Peter Strulo
+#           2018-2019 Peter Strulo
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/quodlibet/quodlibet/ext/events/advanced_preferences.py
+++ b/quodlibet/quodlibet/ext/events/advanced_preferences.py
@@ -1,5 +1,6 @@
-# Copyright 2015 Christoph Reiter
-#        2016-17 Nick Boultbee
+# Copyright 2015    Christoph Reiter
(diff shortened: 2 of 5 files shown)


CreamyCookie:
Looks good (though I don't know enough about the query system to be sure), but is this really something that should be in the main preferences. I think this would fit better into the [advanced preferences](https://github.com/quodlibet/quodlibet/blob/30293621f466cbc0076ed2db703702ef0234a853/quodlibet/quodlibet/ext/events/advanced_preferences.py).

Maybe the query thing also needs tests. (Does it work with queries like `artist=test-test` if `-` is ignored? If not, would that be confusing?)

blimmo:
Wait... There's an advanced preferences?! XD

Yeah putting it there sounds like a good idea, I'll move it there when I have time.

I avoided tests mostly because I wasn't really familiar with the organisation of them and I wasn't sure if this was really a big enough change to warrant them but I can add probably add those too.

CreamyCookie:
Thanks! 

> Wait... There's an advanced preferences?! XD

Yeah, it's a bit hidden since it's currently implemented as a plugin. (If #2156 gets implemented that might change but that will likely happen anytime soon, if ever.)

CreamyCookie:
Looks good to me.

(the one test fail is again caused by IRC connection issues, so unrelated to this pull request)

declension:
> (the one test fail is again caused by IRC connection issues, so unrelated to this pull request)

BTW this should now be fixed (#3241) so meeting from `master` might help here

declension:
I've retried that job nonetheless

CreamyCookie:
@declension Nice, thanks! :)

PR #3232, opened 2019-07-29
Closes #3185: Don't modify command when using parameters

### LINKED ISSUE #3185: Custom commands plugins: Parameter value is replaced in the Command after first run
Steps to reproduce
------------------
Using the `Custom Commands` plugin, I created a command which should ask for the value of a parameter each time it is run. However, after the first run, the value of the parameter is textually replaced in the command, so that subsequent runs won't ask for it anymore and will use the parameter value from the 1st run.

Here's the command I defined:

When I ran it for the first time:

Here's how the command looked like after running it once:

Expected Output
---------------
QL shouldn't have changed the command after running it once. It should have left it like it was defined:

Actual Output
-------------
After running the command once:

QL changed it to:

Test System
-----------

Latest master (97c2f0dd3)

### Which operating system 
Xubuntu 18.04.02

CreamyCookie:
This closes #3185. Currently the command is permanently modified the first time a user parameter is supplied.

### DIFF
--- a/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
+++ b/quodlibet/quodlibet/ext/songsmenu/custom_commands.py
@@ -94,7 +94,7 @@ class Command(JSONObject):
         if playlist_name:
             print_d("Playlist command for %s" % playlist_name)
             template_vars["PLAYLIST"] = playlist_name
-        
+
         actual_command = self.command.format(**template_vars)
         print_d("Actual command=%s" % actual_command)
         for i, song in enumerate(songs):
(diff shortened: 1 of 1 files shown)


CreamyCookie:
Hmm.. another test that failed because of an IRC connection issue. (Can we make it so an IRC connection failure doesn't cause the tests to fail?)

> Your job ran 3060 tests in py.test with 0 failures
> 
> IRC connection refused by irc.oftc.net:6667
Attempt to post message to channel quodlibet on irc server irc.oftc.net:6667 failed

Issue #3135, opened 2019-02-18
New small translations


underailton:
https://gist.github.com/underailton/71031225d0e10e6542af13e8d50fa3c1

PR #3233, opened 2019-07-29
Added Portuguese/Brazil translations

CreamyCookie:
This is mostly copied from #3135 posted by @underailton. ("mostly" because I merged it with the current version.)

Not sure about, but if I understand it right the changes can be used under GPL 2 with a pseudonym (since @underailton didn't add his real name). 

If that's the case, either the real name or his user name have to be added to the notice before this is merged.


CreamyCookie:
(The test failed because of a connection error AFAICS: Failed to connect to codecov.io port 443: Connection timed out)

PR #3235, opened 2019-07-30
Closes #946: Rename cover filename option

### LINKED ISSUE #946: Rework preferences
Original [issue 946](https://code.google.com/p/quodlibet/issues/detail?id=946) created by reiter.christoph on 2012-03-18T23:48:03.000Z:

Human title case:

Move the extended version into the plugin, leave a basic one in the core that gets used for performer roles. Move the option to the plugin.

Scan Directories:

Make it a proper list with add/delete
Add a new tab for tagging (?)

Force image filename:

It's a bit all or nothing.. so maybe &quot;prefer iamge filenames&quot; instead.

Color search term:

Make the non complex search black, and remove the option.

Search after typing:

Maybe add to the search bar context menu instead.

Split on:

Add a revert to default button

Visible columns:

Support patterns somehow.
Possible solution: Leave the checkboxes. Show some text info about how many more columns are defined and provide a way to edit them. Either through a new window, or a default closed expander at the bottom showing a list

CreamyCookie:
Label changed from fixed to preferred image filename(s). Also updated the tooltip and used it for both label and entry.

### DIFF
--- a/quodlibet/quodlibet/qltk/prefs.py
+++ b/quodlibet/quodlibet/qltk/prefs.py
@@ -318,10 +318,11 @@ class PreferencesWindow(UniqueWindow):
             hb = Gtk.HBox(spacing=3)
 
             preferred_image_filename_tooltip = _(
-                "The album art image file to use when available "
-                "(supports wildcards)")
+                "The album art image file(s) to use when available "
+                "(supports wildcards). If you want to supply more "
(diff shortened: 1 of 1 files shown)


declension:
LGTM

CreamyCookie:
@declension Thanks for the review! I'll squash and merge this after #3236.

PR #3236, opened 2019-07-31
Closes #1769: Allow multiple entries for cover files

### LINKED ISSUE #1769: Allow multiple entries for selecting artwork (cover) file
in util/cover/built_in at line around 75, if force_filame is set, QL picks up the exact name specified, but it seems obvious for me that "Folder" or "folder" is quiet the same thing.
Shouldn't QL understand this againts of letting the user manually correct the capital letter  ?

Just an idea.

Also, if force_filame is not set and if you have 2 jpg's, namely "cover" and "folder", folder wins (see around line 139 in the same file.) but both got the same number of points.
Now image that "cover" is the big picture and "folder" what it should be, a  thumbnail, you get the little picture.
What if QL would read the picture size and let the winner be the biggest picture ?

CreamyCookie:
Allows multiple entries in preferred cover art filename (Preferences > Browsers > Album Art). Items at start of the comma-separated list have a higher priority than those at end.

Additionally with this change, the normal cover "search" algorithm will start, if the no preferred files were found.

With this change names containing `,`, or spaces at start or end cannot be directly specified anymore, but that shouldn't be a common use case. Of course, any feedback regarding this and the whole change in general is welcome.

Not perfect, but I'll do the user facing changes in #3235. (label rename in settings)

### DIFF
--- a/quodlibet/quodlibet/util/cover/built_in.py
+++ b/quodlibet/quodlibet/util/cover/built_in.py
@@ -1,5 +1,6 @@
 # Copyright 2013 Simonas Kazlauskas
 #      2015-2018 Nick Boultbee
+#           2019 Joschua Gandert
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


CreamyCookie:
@lazka Hey! Anything to improve before a merge, or a reason for not merging?

If `,` is a problem, we could allow backslash escaping or use an alternative UI element that supports entering multiple items.

Issue #3131, opened 2019-02-17
Permanent playlist overwrites and erasures should not be so aggressive and transparent.


will-ca:
The proposed change
-------------------

If you click any of the tabs to sort your view by Title, Artist, Album (etc.) while viewing a playlist, the displayed order of songs will be changed with no clear way to revert it to the original playlist's order, and if you then click and drag any tracks to try to fix it, the order of the playlist itself and all associated files is promptly also changed permanently to the sorted order, even though it's not immediately clear and there is no indication or precedent that doing so will affect anything other than the displayed order.

Additionally, if a file is missing or has been renamed, it just gets deleted from the playlist with no confirmation. (A *notification* shows up, but by then it's too late.)

And if the music library's directory has been relocated or renamed, *all* files from *all* playlists get erased permanently from all playlist files in `~/.quodlibet`, with no backups, and with neither confirmation nor indication communicating such a large change. (I just lost multiple playlists, including one with over four hundred songs, because of this.)

None of this should happen, and especially not so easily.

Benefits of this change
-----------------------

Playlists are a great way of organizing your music library. They can be used to tell a story by connecting songs in a sequence, they can help with sorting and finding songs based off of mood or social setting, and they can bear significant sentimental significance as well when either they or their constituent songs are connected to personally important events.

In addition to all this— or perhaps because of it— they can also represent a significant amount of work spent by the user in manually creating and maintaining them.

*Quod Libet should not Irreversibly delete or modify user-created and heavily curated data without any kind of clear prompt.*

Some proposed solutions
---------------------------

The primary potential challenge is in finding an elegant way to change this behaviour.

For the issue with the order being overwritten, I suggest creating a stronger interface-level distinction between "playlist order" and "display order". In every other part of the Quod Libet interface, the way the display is sorted is completely separate from and has no bearing on the actual contents of the material being sorted (like albums' track listings), so the behaviour in the playlist browser is inconsistent, confusing, and troublesome without any explicit indication.

The saved playlist order could be exposed to the user through the creation of a new `~playlistpositions` tag for applicable tracks, generated from `~/.quodlibet/playlists` to parallel the already-existing `~playlists` tag. The value of this tag per each playlist should be displayed in its own column as the default sort key in the Playlists browser.

[...]

Issue #3105, opened 2019-01-06
Grey out songs if they can't be played. Never remove them in any list.


DonRichie:
The story
-------------------
I just moved my Music folder to another place. I changed the path in the settings and made quodlibet rescan the library. As I did that all my playlists became empty, the songs were removed. 

My music is on a network share. I also noticed when the share is offline quodlibet shows me empty playlists. After the share is online again the files come back.

The sum of the described baheviors left me kind of worried quodlibet removed something out of my playlists without my knowledge. It feels like any network outage might cause problems.

I see I might have used the library-path settings dialog wrong, but as a user I never thought quodlibet would wipe out my playlists.

The proposed changes
-------------------

* Grey out songs if they can't be played.
* Never remove any song out of any list without the user confirming it or knowing what he is doing.

Please consider changing the behavior when a file doesn't exist and quodlibet tries to play it. The usual way other programs handle that is greying out a file in the list if it can't be played. 

The change would be good for any kind of list in quodlibet. Not only playlists. I think in the music library a song is simply skipped if it doesn't exist. Greying out songs there would be better aswell.

Benefits of this change
-----------------------

It prevents users from losing their playlists. In my opinion the proposed change also fits better to the users expectation so quodlibet feels more reliable.

Any issues this might cause
---------------------------

Don't see any.

The good news
---------------------------
I got a backup. All fine on my side.

urielz:
Great idea!

will-ca:
The playlist files don't seem to store any metadata. They only save the filepath.

With no cached metadata and without access the original file and its `id3` tags, you wouldn't be able to display the Title, Artist, Album (etc.) information for even a greyed out track. You could display the raw path, but that'd be ugly and it wouldn't fit in the column-based display.

I strongly agree with the premise though. Please don't randomly erase user-curated data as your default behaviour.

declension:
See #1010 too

alanpaulprice:
Playlists could make use of `#EXTINF`:

```
#EXTINF:294,Attitude - Alien Ant Farm
/home/media/user/drive/Music/Alien Ant Farm - ANThology/Attitude.mp3
```

It would provide something more presentable than the raw path.

declension:
> Playlists could make use of #EXTINF:

The problem is we don't use m3u (or any structured data) for playlists (hence #1122)

derek-v-s:
[...]

Issue #1122, opened 2015-03-15
Add XSPF support for playlist import and export


lazka:
Original [issue 1122](https://code.google.com/p/quodlibet/issues/detail?id=1122) created by shadowfax5124 on 2013-02-20T22:50:53.000Z:

While the m3u plugin suffices, I prefer the xspf. Are ther any plans to support it?

declension:
There's also now JSPF: http://www.xspf.org/jspf/

ibrokemypie:
No news for 2 years? Really would love metadata based playlists rather than file based...

declension:
:heavy_check_mark:  I've got an XSPF-based playlist browser PoC working for QL 

Obviously, this is a deeper integration than just import and export support, but (and see discussion on #518), this could provide other benefits that now seem worth exploring.

### Things to think about
Even assuming it's something that fully works, and we'd want, then maybe have these first:
* [ ] Warning / documentation this is happening
* [ ] Ample thought into any knock-on effects in the remote "API"
* [ ] Automation around migration (see below)  

### Migrating users to XSPF-based playlists
1. Migrate all users to XSPF-based once, on upgrade of QL (either as a planned migration or just on first load of any non-XSPFs)
2. Read both formats, but save as XSPF only. Not many of the benefits (see #1010, #3105 etc) though until playlists _all_ have metadata (this is another open question)
3. Do 2., then 1 on the next (major?) release.
4. Have a preference system for "preferred playlist format"

### Unification of playlist support
 * It seems that m3u / pls should perhaps just be another supported format eventually that can be read natively if dropped into the directory (related: #174) 
* Then maybe remove plugins and expose UI to export or "add" in any format "natively"

declension:
@lazka / @frestr any opinions?

declension:
(See PoC in [`xspf-support`](..tree/xspf-support) branch)

lazka:
What's the advantage of using XSPF to store things compared to the current way?

declension:
> What's the advantage of using XSPF to store things compared to the current way?

#### TL; DR 
It's more about the things this can then unlock (#518, #1010, #3105 / #3131 / #708)

#### Longer version 
Not much, for now (other than this issue of course). 

[...]

Issue #1010, opened 2015-03-14
Songs are forcibly removed from playlist on refresh


lazka:
Original [issue 1010](https://code.google.com/p/quodlibet/issues/detail?id=1010) created by _deepfire@feelingofgreen.ru on 2012-07-20T10:30:02.000Z:

This applies to all versions of QL, up to, at least 2.2.1.

The misbehaviour can be summarised as the general QL's policy of dropping an entry from a playlist, whenever it is unable to locate the corresponding file.

While this could be desirable for some people, there are others, who absolutely loathe this.

This is aggravated by the fact, that sometimes QL fails to locate the file, despite it being present -- which leads to gratuitious lossage.

The proposed solution is to provide an optional mode, whereby playlist entries for missing files will not be deleted, but rather presented as such.

CreamyCookie:
Duplicate of #1733 as far as I can see.

PR #3242, opened 2019-08-12
XSPF support (#1122)

declension:
Check-list
----------

From #1122:

...this can then unlock (#518, #1010, #3105 / #3131 / #708)
...
But because this format allows us to store metadata as well as the filename, we can have a chance of saving files have gone away / moved, e.g. intelligently "finding" missing playlist items by artist / title / album (currently) but presumably by say MBIDs or labelid in the future (it's extensible, or offers identifier which perhaps we could use here)

I guess a few other things, e.g. comments / URLs / images for playlists would then be natively supported if extending the Playlist Browser... but not sure how useful these are.


quodlibet/quodlibet/util/collection.py
```diff
@@ -595,38 +612,148 @@ def from_songs(cls, dir_, songs, library=None):
...
+                        % (node.text, self.name, self.path))
+            for node in tree.iterfind('.//track'):
+                location = node.find('location')
+                path = location.text.replace('\n', '').replace('\r', '')
+                if path in library:
+                    self._list.append(library[path])
+                elif library and library.masked(path):
+                    self._list.append(path)
+                else:
+                    # TODO: handle missing playlist items (#3105, #729, #3131)
```
CreamyCookie: I wonder how this would be implemented.. The playlist could be loaded, ignore invalid entries and then merge in the changes, right?

quodlibet/quodlibet/util/collection.py
```diff
@@ -595,38 +612,148 @@ def from_songs(cls, dir_, songs, library=None):
...
+        playlist.append(track_list)
+        tree = ElementTree(playlist)
+        ET.register_namespace('', XSPF_NS)
+        tree.write(self.path, encoding="UTF-8", xml_declaration=True)
+        if self._last_fn != self.path:
+            self._delete_file(self._last_fn)
+            self._last_fn = self.path
+
+    @classmethod
+    def text_el(cls, name: str, value: any) -> Element:
```
CreamyCookie: minor thing, but I think for consistency this should also have "element" in the name.

quodlibet/quodlibet/util/path.py
```diff
@@ -104,17 +104,18 @@ def filesize(filename):
...
 
 
-def escape_filename(s):
+def escape_filename(s, safe=''):
     """Escape a string in a manner suitable for a filename.
 
     Args:
-        s (str)
+        s (str) The string to convert
+        safe (bytes) A string of
```
CreamyCookie: This is missing a description. (a string of ...)

quodlibet/tests/test_util_collection.py
```diff
@@ -619,3 +636,37 @@ def test_masked_handling(self):
...
+        super().setUp()
+        self.temp = mkdtemp()
+        self.temp2 = mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.temp)
+        shutil.rmtree(self.temp2)
+
+    def test_from_songs(self):
+        pl = XSPFBackedPlaylist.from_songs(self.temp, NUMERIC_SONGS)
```
CreamyCookie: Shouldn't this also use `self.Playlist` like in `TFileBackedPlaylist`

quodlibet/tests/test_util_collection.py
```diff
@@ -529,9 +532,10 @@ def test_remove_fully(self):
 
 
 class TFileBackedPlaylist(TPlaylist):
```
CreamyCookie: I think `TFileBackedPlaylist` and `TPlaylist` should start with `Test` too (instead of `T`), to be consistent with `TestXPSFBackedPlaylist`

quodlibet/quodlibet/util/collection.py
```diff
@@ -595,38 +612,148 @@ def from_songs(cls, dir_, songs, library=None):
...
+                    self._list.append(path)
+                    library.mask(path)
+        except ET.ParseError as e:
+            print_w("Couldn't load %s (%s)" % (self.path, e))
+
+    @classmethod
+    def filename_for(cls, name: str) -> str:
+        # Manually do *minimal* escaping, to allow near-readable filenames
+        for bad, good in cls._SAFER.items():
+            name = name.replace(bad, quote(good, safe=''))
```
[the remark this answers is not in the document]
declension: yep, that was a mistake, that I fixed.. then forgot to push... :man_facepalming:

quodlibet/quodlibet/util/collection.py
declension: Good shout, I'll update that

quodlibet/tests/test_util_collection.py
declension: Yes, but _that_ discussion... I just counted 299 such classes already in the codebase...

declension: I'll change this back then though, it's a fair point about consistency

quodlibet/tests/test_util_collection.py
declension: Good point, in fact even better not sure why that method's still there as the base version now handles it

quodlibet/quodlibet/util/collection.py
declension: Who knows - but it all feels possible

quodlibet/quodlibet/util/collection.py
```diff
@@ -535,56 +544,64 @@ def __str__(self):
...
-                pass
-        raise ValueError("Couldn't create playlist of name '%s'" % base)
+        name = "%s %d" % (base, attempt) if attempt else base
+        fn = cls.filename_for(name)
+        last_err = None
+        try:
+            return cls(dir_, fn, library, validate=True)
+        except ValueError as e:
+            if attempt < 1000:
+                return cls.new(dir_, name, library, attempt + 1)
```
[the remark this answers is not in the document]
declension: > couldn't that recursive call not theoretically lead to a RuntimeError (maximum recursion depth exceeded

Well, hence the safety net (1000). The stack should be within that (I just tried and maxed out at 3000, which TIL is [configurable by design](https://docs.python.org/3/library/sys.html#sys.getrecursionlimit)).

> doesn't this do to a lot of repeated checks of dir_. (assert .. and if not (dir_ and os.path.realpath(dir_)):) A loop seems better..

Yeah, maybe it is. Been doing too much non-Python maybe :grin: 

I'll revert it when I get a bit of time...

declension:
@CreamyCookie @lazka / others would you mind testing this branch locally? 

I'd like to finally merge it in (just keeping it alive is a hassle - see above...), but need more confidence that it works well enough on other real-world setups.

Also, what to do about [versioning](https://quodlibet.readthedocs.io/en/latest/changelog.html)? At what point do we say a major version changes? (side note: I imagine we're due a Python version upgrade at some point, 3.9 is out now, Ubuntu LTS is 3.8 now, I'll raise a ticket)

quodlibet/browsers/playlists/main.py
```diff
@@ -145,6 +152,8 @@ def cell_data(self, col, cell, model, iter, data):
             return
         self.__last_render = markup
         cell.markup = markup
+        if isinstance(playlist, XSPFBackedPlaylist):
+            markup += " <small><tt>[XSPF]</tt></small>"
```
CreamyCookie, line 156: Not a fan of this. Adds a lot of noise to the interface.

Screenshot_20200908_203705

CreamyCookie (commented):
Rest LGTM. 

Works perfectly so far. No noticeable performance impact on my end.

quodlibet/browsers/playlists/main.py
declension: Yeah, actually this was more for my debugging, I'll remove it. Thanks!

quodlibet/util/collection.py
```diff
@@ -596,38 +613,148 @@ def from_songs(cls, dir_, songs, library=None):
...
-        new_name = super(FileBackedPlaylist, self)._validated_name(new_name)
-        basename = self.quote(new_name)
-        path = os.path.join(self.dir, basename)
+        new_name = super()._validated_name(new_name)
+        path = os.path.join(self.dir, self.filename_for(new_name))
         if os.path.exists(path):
             raise ValueError(
-                    _("A playlist named %s already exists.") % new_name)
+                _("A playlist named %(name)s already exists at %(path)s")
+                % {"name": new_name, "path": path})
```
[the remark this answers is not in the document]
declension: Yeah, though new of course is now f-strings which are better than both IMO (See #3305 actually)

docs/guide/browse/playlists.rst
```diff
@@ -14,15 +14,30 @@ Choose the *Playlists* browser by clicking on *View* -> *Playlists*. The
...
+The names are partially URL-encoded - that is, they can and should be URL-unencoded,
+but efforts are taken to make them human-friendly names where reasonable.
+
+As before, these playlists are editable but it's recommended to leave them to Quod Libet.
+
+Quod Libet < 4.4
+~~~~~~~~~~~~~~~~
+
+Playlists are stored as plain text files, with the names URL-encoded.
+Whilst these can be edited, it's recommended to leave it to Quod Libet.
```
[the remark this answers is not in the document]
declension: Definitely should put it in `NEWS.rst`, but I've often thought we should.

That said RTD is very good at versioning so perhaps we can just put this in (with a :new: symbol perhaps)?

Aside: I thought there was somewhere about tags that mentioned a step change in QL, can't find it. There's https://quodlibet.readthedocs.io/en/latest/guide/faq.html?highlight=newer#where-does-quod-libet-store-all-its-metadata though

quodlibet/util/collection.py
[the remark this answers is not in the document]
CreamyCookie: True, though f-strings will not work with translatable text since the variable are interpolated immediately.

CreamyCookie (approved):
LGTM

Meriipu:
Maybe squash it to not make bisecting more painful though.

declension:
@Meriipu yep definitely squashing this one :+1:

declension:
@lazka happy if I merge this one (finally)?

PR #3245, opened 2019-08-15
serbian translation for QL

leipero:
Check-list
----------

Adds translation for Ql, tested via ./setup.py test --to-run PO.<mylang> (not in this repository).

### DIFF
--- a/quodlibet/po/sr_RS.po
+++ b/quodlibet/po/sr_RS.po
@@ -1,0 +1,6809 @@
+# Serbian translations for PACKAGE package
+# Преводи на српски језик за пакет „PACKAGE“.
+# Copyright (C) 2019 THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# Automatically generated, 2019.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2019-08-13 06:59+0200\n"
+"PO-Revision-Date: 2019-08-15 10:07+0200\n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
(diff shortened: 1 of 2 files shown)


piotrdrag:
It would be great if these files could be renamed to `sr.po` and `sr@latin.po`, as that’s the usual convention.

leipero:
> s

I've looked at examples of en_CA, and en_GB, as well as zh_CN and zh_TW, generally for most translations I saw that format is used, the reason for it is because same as in other two examples, there are more sr_XX variations (specifically for BiH, CRO, MN mostly) depending on location with slight (but for someone important) differences, hence I've choosen that format.

However, I don't have any objections for renaming .po files, however would work well for QL devs/contributors, their choice.

piotrdrag:
No (free) software I know of has separate translations for sr_XX variants (which I presume is mostly a lack of manpower issue that is not going to change any time soon.) Using just sr would guarantee the users of all these locales have translations that they can at least mostly understand.

leipero:
> No (free) software I know of has separate translations for sr_XX variants (which I presume is mostly a lack of manpower issue that is not going to change any time soon.) Using just sr would guarantee the users of all these locales have translations that they can at least mostly understand.

Without intent for discussion, if I remember correctly, there were at least few projects that follow same principle for translation files, and all of those locales are listed under locales in any GNU/Linux distribution for example. I don't really know how locales work exactly, do they follow main language for specific locale if any variation is available or not, so I will assume you have valid point for using just sr.po for locale, and as I said, it doesn't really make any difference to me, I just wanted to follow GNU/Linux standard for locales that is present in all distributions AFAIK, for example, you have multiple es_XX (Spanish), en_XX as you already know and any language that have variations for specific country it is used in.
My starting assumption was that locale daemon is clever enough to 'realize' that it should use any available yy_XX locale for any yy locale that is used and have no available match, however, I could be completely wrong.

PR #3251, opened 2019-08-24
Added New Internal "~library" Tag.

Eoin-ONeill-Yokai:
This PR creates a new internal tag, `~library`, which allows users to filter their music by library scan directory. This effectively treats each scan directory as a separate library or music collection.

Quodlibet already allows users to add multiple scan directories and people who are using this feature probably already have their music organized in a way that represents separate logical "libraries." In other words, if their music comes from different places, it's likely to represent a separate collection.

While you could achieve something similar using an explicit, external `collection` tag applied to every song in your library, that's both inconvenient and unnecessary if you already have your songs organized into logical collections on your file system. This can be inferred from the fact that they have been imported from separate scan directories. 

For example, say you have your main music library as `~/Music/Personal Collection` which contains all of your favorite artists. You might also have other folders in your library like `~/Work/Compositions` which contains music that you have made, or `~/Music/Soundtracks` which contains a collection of soundtracks. This is kind of how my personal music collection is organized. With this PR, those different locations will be stored using an internal tag named `~library`, which would have the following values:

```
Personal Collection
Compositions
Soundtracks
```

If the user wants, they can then create a new pane in the panned browser that allows them to filter their music by the associated library or use the search bar.

### DIFF
--- a/quodlibet/formats/_audio.py
+++ b/quodlibet/formats/_audio.py
@@ -59,7 +59,8 @@ NUMERIC_ZERO_DEFAULT = {"~#skipcount", "~#playcount", "~#length", "~#bitrate"}
 NUMERIC_ZERO_DEFAULT.update(TIME_TAGS)
 NUMERIC_ZERO_DEFAULT.update(SIZE_TAGS)
 
-FILESYSTEM_TAGS = {"~filename", "~basename", "~dirname", "~mountpoint"}
+FILESYSTEM_TAGS = {"~filename", "~basename", "~dirname", "~mountpoint",
+                   "~library"}
 """Values are bytes in Linux instead of unicode"""
 
 SORT_TO_TAG = dict([(v, k) for (k, v) in TAG_TO_SORT.items()])
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -720,7 +720,7 @@ class FileLibrary(PicklingLibrary):
             return False
 
         # first scan each path for new files
-        paths_to_load = []
+        elements_to_load = []
         for scan_path in paths:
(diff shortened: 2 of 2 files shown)


Eoin-ONeill-Yokai:
Hey everyone, I'm still interested in having this merged at some point. I think with the 3 pane browser it's pretty useful to have all of the libraries available for filtering my music collection. Here's how this looks on my current system setup.

quodlibet-panel

Which represents the following scanned libraries:
quodlibet-settings


This also allows me to add a Downloads directory and the like to track individual folders. Perhaps this tag could be a togglable option in the library preferences window? Just let me know if there's something I can do to improve this and make it fit the greater project. I have been testing this for about a year now and it seems to work pretty consistently on my end.

Eoin-ONeill-Yokai:
> So it's just to ping people, if any are still following it ! Especially @Eoin-ONeill-Yokai why did you close that ? It's look nice :)
> Cheers

I was thinking about reopening this MR after I had to do a pretty extensive rebase. However, this MR was open for quite a while without any activity or merge so I'm not sure if there's a demand for this particular feature. However, I might open a new version of this again soon.

ShuffleWire:
Go for it, then, I would be an happy user (well, and my mom would also, actually !). I do think this feature is a very nice touch and would help many people !

declension:
I'd say the best way to go here now is to open an _issue_ to discuss the feature request here - that helps when it comes reviewing PRs (hence the note in the template).

In advance of that though I'd say this looks quite reasonable - just some naming considerations  maybe

ShuffleWire:
I'm creating the issue, because I wish to see this merged, it would be a very good improvement, and the main part of the work is already done !

PR #3282, opened 2019-11-29
Import/export plugin: accept full filenames when renaming

frestr:
Maybe this fixes the issue with azure pipelines?


frestr:
So the problem was that `mkstemp` generated a shortened path in Windows (`vssadministrator` became `vssadm~1`) which was later normalized by the import plugin, making the test comparison fail.

Adding the normalization right after `mkstemp` also fixed the test for macOS, nice.

lazka:
Nice, thanks!

PR #3300, opened 2020-01-25
Zsh completion

### LINKED ISSUE #3299: Install zsh completer to site-functions
The proposed change
-------------------
The zsh completion file should be installed to `$PREFIX/share/zsh/site-functions` rather than `$PREFIX/share/zsh/vendor-completions`.

Benefits of this change
-----------------------
vendor-completions is a Debian-ism that many distros and other OSes don't support. site-functions is added to the default `$fpath` by zsh, so it should just work for all users.

Any issues this might cause
---------------------------
Debian-style distro maintainers will likely want to update their packages to move the completion file back to vendor-completions.

phy1729:
Check-list
----------

 * (N/A) Unit tests have been added where possible
 * (N/A) Performance seems to be comparable or better than current `master`

Fixes #3299 .
Installs the zsh completer to the more universal site-functions rather than vendor-completions.
Also cleanup an "array" in the completer to follow idiomatic style with a real array,

### DIFF
--- a/quodlibet/NEWS
+++ b/quodlibet/NEWS
@@ -6,6 +6,7 @@
 Packaging Changes:
   * Various installed files have been renamed
     ("exfalso" -> "io.github.quodlibet.ExFalso") to work better with Flatpak
+  * zsh completion file installation location changed to site-functions
 
 
 .. _release-4.2.1:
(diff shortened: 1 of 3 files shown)


lazka:
Thanks!

PR #3306, opened 2020-02-11
Add Poetry support 🆙

### LINKED ISSUE #3239: Switch to Poetry from Pipenv for self-contained builds
The proposed change
-------------------
* Replace the defunct `Pipfile` with a working `pyproject.toml` from [Poetry](https://poetry.eustace.io/), *and the lock file* 
* Avoid installing `quodlibet` from OS when building from source - this is unusual / risky but "should work"
* Update contributing docs for new devs. The aim is a single line to install all dependencies and pin to a Python version.
* Build more things from Poetry directly in CI, to validate this setup. Should have done this for Pipenv anyway...

Benefits of this change
-----------------------
* Our Pipenv setup is unusable currently (Just tried it, missing lots of dependencies)
* Adding these causes dependency resolution errors, a problem many [have complained about](https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/).
* Personally, I've been using Poetry significantly for > 6 months and some but very few issues. At $WORK, have rolled it out to ~10 devs across the three major OSes without significant issue.
* At some point, Poetry might supersede `setup.py` for a more integrated experience, but this proposal is not addressing packaging and distribution.

Any issues this might cause
---------------------------
None really, as this will remain complementary to the current setups, for now.

declension:
* First pass at a fully configured Poetry file
 * Also the current lockfile from this
 * Also ignore common virtualenv dirs
 * TODO: switch some CI over, maybe
 * TODO: some docs around developer experience

This fixes #3239

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,48 @@
+[tool.poetry]
+name = "quodlibet"
+version = "4.3.1"
+description = "a music library, tagger, and player"
+authors = ["Joe Wreschnig", "Michael Urman", "Steven Robertson",
+           "Christoph Reiter", "Nick Boultbee", "(many others)"]
+maintainers = ["Christoph Reiter"]
+homepage = "https://quodlibet.readthedocs.io/"
+documentation = "https://quodlibet.readthedocs.io/"
+repository = "https://github.com/quodlibet/quodlibet"
+keywords = ["audio", "music", "player", "tags", "gtk"]
+license = "GPL-2.0-or-later"
+
+packages = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+]
+
+
+[tool.poetry.dependencies]
+python = "^3.5"
+mutagen = "^1.44"
+senf = "^1.4"
+feedparser = "^5.2"
+# Or is this a dev dependency?
+pyGObject-stubs = "*"
```
lazka: yeah, should be dev

PR #3309, opened 2020-02-23
ext/inhibit: Add option to inhibit suspend

### LINKED ISSUE #2665: Prevent system suspend while playing
Currently, there is a plugin which disables the screensaver while playing. It also prevents the computer from going to sleep. However, I think there should be an option to prevent the system from going to sleep without disabling the screensaver. I do want the screensaver to become active, my user session being locked and my monitor go into sleep mode to save energy. I just don't want the whole system to sleep, so that playing always continues until I stop it myself.

antigone-xyz:
ext/inhibit: Add option to inhibit either

- screensaver (default, as before) or
- suspend (new).

Resolves #2665.

Check-list
----------

    ext/inhibit: Add option to inhibit suspend (as requested through #2665)


lazka:
Looks good, thanks

PR #3310, opened 2020-02-23
IRadio - do station updates in background

### LINKED ISSUE #3281: Iradio: UI freezes until connection times out
Steps to reproduce
------------------
add the following url for a new radio station:
http://146.186.221.181:8000/listen.pls

whatever audio that was playing continued to play;
however Libet locks up, no response to any input.
had to be terminated. 

4.2.1

### Which operating system 
kubuntu 19.10

declension:
* Do fetching of Internet Radio stations in a new Thread but using GLib idles to schedule UI dialogs etc if necessary
 * Change a few things to Exceptions to mix the view and logic code less...
 * Also, use a basic `Task` to indicate some progress. Not very pretty though. TODO: umm, make nicer?
 * Better logging
 * Fix bug around (attempted) re-addition of existing favourite stations
 * `ParseM3U` and `ParsePLS` functions finally renamed :)

This fixes #3281


lazka:
Is the Task interface thread safe? It looks like that calls straight into gtk.

declension:
> Is the Task interface thread safe? It looks like that calls straight into gtk.

Oops just saw this. Hmm not sure.

declension:
@lazka any ideas on ensuring this is thread-safe? Could just ditch the Task stuff entirely... or some kind of `threading` datastructure to push back I guess?

declension:
OK - hopefully this version handles it correctly. Tested locally a lot and seems to have no issues

PR #3328, opened 2020-03-23
win_installer: pass options to build script on env switch

gkarsay:
Check-list
----------

The Windows installer build script spawns itself, if it's not in the MINGW console. It doesn't pass the original options, though, and builds only from master. Any git tag given as an option is lost.

I have discovered this in my copy of your build script at https://github.com/gkarsay/parlatype/blob/master/build-aux/win32/build.sh and tested it with my copy.


lazka:
Thanks. (it probably should have been `"$@"` to catch all future changes, but this works)

PR #3365, opened 2020-05-03
Fix setting pane mode

### LINKED ISSUE #3367: Column mode resets on restart
Steps to reproduce
------------------
Start Quod Libet and change the browser's column mode to wide or columnar. Afterwards, close and restart Quod Libet.

Expected Output
---------------
Quod Libet should restore the column mode set previously.

Actual Output
-------------
Quod Libet resets the column mode to small.

Test System
-----------

4.3.0

### Which operating system 
Fedora 32

Additional Information
----------------------
When setting the column mode, the config file contains something like this:
```
pane_mode = ColumnMode.COLUMNAR
```

This is due to the fact that ColumnMode is an enum and casting it to a string returns the enum name. However, the config code expects an int. This seems to have been introduced by e73e007fe57317249752a6cbab83f0adfc0bf9e6.

michaelkuhn:
Check-list
----------

When setting `pane_mode` via the GUI, the config file contains something like this:
```
pane_mode = ColumnMode.COLUMNAR
```

This is due to the fact that ColumnMode is an enum and casting it to a string returns the enum name. However, the config code expects an int. Manually setting `pane_mode` to an int in the config file fixes the problem, so we just cast it to an int before setting the config value.

Fixes #3367

### DIFF
--- a/quodlibet/browsers/paned/prefs.py
+++ b/quodlibet/browsers/paned/prefs.py
@@ -62,7 +62,7 @@ class ColumnModeSelection(Gtk.VBox):
             selected_mode = ColumnMode.WIDE
         if self.buttons[2].get_active():
             selected_mode = ColumnMode.COLUMNAR
-        config.set("browsers", "pane_mode", selected_mode)
+        config.set("browsers", "pane_mode", int(selected_mode))
         self.browser.set_all_column_mode(selected_mode)
 
 


declension:
Thanks. Could you file (or link to) a bug for this problem?

michaelkuhn:
I could not find an open issue about this, so I created #3367.

lazka:
lgtm

Issue #3393, opened 2020-06-12
macos: bundle.sh can't clone


smammy:
Steps to reproduce
------------------
1. Build quodlibet per [macos bundle instructions](https://github.com/quodlibet/quodlibet/tree/master/dev-utils/osx_bundle#creating-a-bundle).
2. Run `./bundle.sh` as instructed.

Expected Output
---------------
```
('Warning, library not available in any prefix:', '/usr/local/opt/gdbm/lib/libgdbm.6.dylib')
('Warning, library not available in any prefix:', '/usr/local/opt/gdbm/lib/libgdbm.6.dylib')
objc[66429]: Class GNotificationCenterDelegate is implemented in both /Users/sam/Files/Projects/Music/quodlibet/dev-utils/osx_bundle/_build/._launcher.app/Contents/Resources/lib/libgio-2.0.0.dylib (0x1046ac480) and /Users/sam/Files/Projects/Music/quodlibet/dev-utils/osx_bundle/_home/jhbuild_prefix/lib/libgio-2.0.0.dylib (0x104935480). One of the two will be used. Which one is undefined.
[thousands of lines, removed]
Listing '/Users/sam/Files/Projects/Music/quodlibet/dev-utils/osx_bundle/_build/QuodLibet.app/Contents/Resources/lib/python3.7/xmlrpc'...
hdiutil: attach: WARNING: ignoring IDME options (obsolete)
hdiutil: attach: WARNING: ignoring IDME options (obsolete)
```

Command exit staus is 0 and a DMG file is produced in `_build`.

Actual Output
-------------
```
('Warning, library not available in any prefix:', '/usr/local/opt/gdbm/lib/libgdbm.6.dylib')
('Warning, library not available in any prefix:', '/usr/local/opt/gdbm/lib/libgdbm.6.dylib')
objc[93697]: Class GNotificationCenterDelegate is implemented in both /Users/sam/Files/Projects/Music/quodlibet/dev-utils/osx_bundle/_build/._launcher.app/Contents/Resources/lib/libgio-2.0.0.dylib (0x108617480) and /Users/sam/Files/Projects/Music/quodlibet/dev-utils/osx_bundle/_home/jhbuild_prefix/lib/libgio-2.0.0.dylib (0x10889b480). One of the two will be used. Which one is undefined.
[thousands of lines, removed]
[...]

PR #3394, opened 2020-06-12
macos: bundle.sh: clone from ../.. rather than .., fixing #3393.

smammy:
Check-list
----------

    * #3394
    * (not applicable)
    * (not applicable)
    * (not applicable)

`dev-utils/osx_bundle/bundle.sh` is broken since 227c1e7f9fcf due to repo
reorganization. This patch changes the source path for the `git clone` command
to fix it.


smammy:
I'm not sure what that circleci failure is all about but I don't think my change caused it.

PR #3396, opened 2020-06-19
Support for mock-confirmer for ConfirmationPrompt (and example of usage)

Meriipu:
PlaylistsBrowser now has a Confirmer parameter which is passed to dialog-creation helper functions. The dialog instantiates Confirmer and calls its run method, then compares the response-type run returned to the stored accept-RESPONSE_TYPE in Confirmer.

Also changed one more instance of a prompt to use ConfirmationPrompt
instead of a duplicated manually constructed one.

Also gave ConfirmationPrompt an icon-paramater for the ok-button (default as what it already had).

ConfirmationPrompt now uses the enum Gtk.ResponseType.YES as opposed
to a hardcoded 1.

Minor style cleanup now that line lengths of 88 characters are fine.

Check-list
----------

Steps towards making the monkeypatching discussed in #2746 unnecessary to achieve the same. See also #2453.

There should be *no* functional changes with these commits, only code-cleanliness and the possibility for a confirmer.

### DIFF
--- a/tests/test_browsers_playlists.py
+++ b/tests/test_browsers_playlists.py
@@ -199,10 +199,18 @@ class TPlaylistsBrowser(TestCase):
 
         PlaylistsBrowser.init(self.lib)
 
-        self.bar = PlaylistsBrowser(self.lib)
+        self.bar = PlaylistsBrowser(self.lib, self.MockConfirmerAccepting)
         self.bar.connect('songs-selected', self._expected)
         self.bar._select_playlist(self.bar.playlists()[0])
         self.expected = None
+
+        # Uses the declining confirmer.
+        self.bar_decline = PlaylistsBrowser(self.lib, self.MockConfirmerDeclining)
+        self.bar_decline.connect('songs-selected', self._expected_decline)
+        self.bar_decline._select_playlist(self.bar_decline.playlists()[0])
+        # Note that _do() uses self.expected, but _do() is not called by the
+        # testcase for declining the prompt. Tests fail with a shared expected.
+        self.expected_decline = None
 
     def tearDown(self):
         self.bar.destroy()
@@ -215,6 +223,12 @@ class TPlaylistsBrowser(TestCase):
         songs.sort()
         if self.expected is not None:
             self.failUnlessEqual(self.expected, songs)
+        self.success = True
+
+    def _expected_decline(self, bar, songs, sort):
+        songs.sort()
+        if self.expected_decline is not None:
+            self.failUnlessEqual(self.expected_decline, songs)
         self.success = True
 
     def _do(self):
@@ -280,17 +294,11 @@ class TPlaylistsBrowser(TestCase):
         qltk.selection_set_songs(sel, [song])
         b._drag_data_get(None, None, sel, DND_QL, None)
 
+    # deletion of playlist tracks
     def test_deletion(self):
-        def a_delete_event():
-            ev = Gdk.Event()
-            ev.type = Gdk.EventType.KEY_PRESS
-            ev.keyval, accel_mod = Gtk.accelerator_parse("Delete")
-            ev.state = Gtk.accelerator_get_default_mod_mask() & accel_mod
-            return ev
-
         b = self.bar
         self._fake_browser_pack(b)
-        event = a_delete_event()
+        event = self.a_delete_event()
         # This is selected in setUp()
         first_pl = b.playlists()[0]
         app.window.songlist.set_songs(first_pl)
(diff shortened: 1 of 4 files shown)


Meriipu:
Also `get_top_parent(parent)` appears to be called both in `ConfirmationPrompt` and in the superclass that eventually also gets called, resulting in it being called twice. 

I removed the call to it in `ConfirmationPrompt` in favour of the one that happens in `qltk.Message`

Meriipu:
ok maybe I understand now

```
class PlaylistsBrowser():
    # alternatively always passed in by instancer without a default
    def __init__(self, ..., confirmer=ConfirmationPrompt):
        self.confirmer = confirmer
    ...
    def function_with_promptable_action(self):
        # this is the helper-function for this particular actions prompt
        response = confirm_remove_playlist_dialog_invoke(..., self.confirmer)
        if response:
            # go through with the promptable action
```

```
def confirm_remove_playlist_dialog_invoke(parent, playlist, confirmer):
    ...
    title = (_("Are you sure you want to delete the playlist '%s'?")
             % escape(playlist.name))
    ...
    # confirmer is usually ConfirmationPrompt, so this usually creates a ready-to-run dialog
    # in case of a mock-confirmer, this might be an empty class that eats and discards the args,with a simple true/false run-method
    dialog = confirmer(parent, title, description, ok_text, ok_icon)
    prompt = dialog.run()
    response = (prompt == ConfirmationPrompt.RESPONSE_INVOKE)
    return response
```


```
# example of mock-confirmer
class MockConfirmerTrue(object):
    def __init__(self, *args):
        pass
    def run(self):
        return True # or False
```

Meriipu:
How is this second commit (with the added tests)? 

The default parameter of ConfirmationPrompt may or may not be desirable. I also monkeypatch the mock-confirmer in in the actual unit test to avoid having to do a whole new setup with a slightly different mock-confirmer (one that returns GtkResponsetype.CANCEL instead of GtkResponsetype.YES).

Meriipu:
so contingent on this here being what was intended, this pull request should be done now (in case it was not clear)

of course not that there is any rush considering I let it lie for a good 2 years aha

declension:
@Meriipu I think there's some good stuff in here, and a few open questions / nit-picks. 

If you're happy to fix it up perhaps we can aim to get it merged and fix / improve any extras in master / separate MRs?

Meriipu:
Sure. 
I could duplicate that one test construction to avoid any monkeypatching whatsoever as a start.

Meriipu:
Is there a way to change the test setup for one particular test case? I would like to initialize the playlists browser with two separate confirmers in the accepting and in the declining test cases.

declension:
> Is there a way to change the test setup for one particular test case? I would like to initialize the playlists browser with two separate confirmers in the accepting and in the declining test cases.

If I understand what you mean, can't you just use:
```
second_browser = PlaylistsBrowser(self.lib, another_confirmer())
```

Meriipu:
Well now I feel a bit silly. That makes it way cleaner. 
I had some issues with self.expected leading to test init feiling but not sure why. It disappears if I give the declining case its own expected attribute + _expected method.

Meriipu:
Is the change from `get_top_parent(parent)` to `parent` correct? 
in `quodlibet/qltk/msg.py`, `ConfirmationPrompt` inherits from `WarningMessage` which inherits from `Message`. The `__init__`-method of message replaces its parent-parameter with `get_top_parent(parent)`, such that unless there is a difference between calling `get_top_parent(parent)` more than once, it gets called twice (in current master):

https://github.com/quodlibet/quodlibet/blob/master/quodlibet/qltk/msg.py#L18-L24
https://github.com/quodlibet/quodlibet/blob/master/quodlibet/qltk/msg.py#L81-L89

I mentioned it earlier here https://github.com/quodlibet/quodlibet/pull/3396#issuecomment-646587145 but just to be explicit

declension:
> Is the change from get_top_parent(parent) to parent correct?

I think so, seems like a mistake / oversight to call it twice

Meriipu:
Should be ready again now, either way.

tests/test_browsers_playlists.py
```diff
@@ -322,10 +357,38 @@ def fns(songs):
...
+        ev.type = Gdk.EventType.KEY_PRESS
+        ev.keyval, accel_mod = Gtk.accelerator_parse("Delete")
+        ev.state = Gtk.accelerator_get_default_mod_mask() & accel_mod
+        return ev
+
     @staticmethod
     def _fake_browser_pack(b):
         app.window.get_child().pack_start(b, True, True, 0)
 
+    class MockConfirmerAccepting(object):
```
declension, line 372:
```suggestion
    class MockConfirmerAccepting:
```

tests/test_browsers_playlists.py
```diff
@@ -322,10 +357,38 @@ def fns(songs):
...
+
+        RESPONSE_INVOKE = Gtk.ResponseType.YES
+
+        def __init__(self, *args):
+            pass
+
+        def run(self, *args):
+            return self.RESPONSE_INVOKE
+
+    class MockConfirmerDeclining(object):
```
declension, line 382:
```suggestion
    class MockConfirmerDeclining:
```

tests/test_browsers_playlists.py
```diff
@@ -322,10 +357,38 @@ def fns(songs):
...
+
     @staticmethod
     def _fake_browser_pack(b):
         app.window.get_child().pack_start(b, True, True, 0)
 
+    class MockConfirmerAccepting(object):
+
+        RESPONSE_INVOKE = Gtk.ResponseType.YES
+
+        def __init__(self, *args):
```
declension, line 376: This can go, no?

tests/test_browsers_playlists.py
```diff
@@ -322,10 +357,38 @@ def fns(songs):
...
+            pass
+
+        def run(self, *args):
+            return self.RESPONSE_INVOKE
+
+    class MockConfirmerDeclining(object):
+
+        RESPONSE_INVOKE = Gtk.ResponseType.YES
+
+        def __init__(self, *args):
```
declension, line 386: This too?

tests/test_browsers_playlists.py
Meriipu: I think it might be needed unless the implementation is changed
The Confirmer is passed values needed for the gtk dialog in the function confirm_remove_playlist_dialog_invoke

declension (approved):
LGTM!

declension:
Thanks!

Issue #3318, opened 2020-03-06
Quodlibet 4.2 italian translation


dprimon:
Hi, I've fully translated the quodlibet-4.2 branch of the repository. I would like to push the changes with git but I have not figured out how to do it. So here is the link to the .po file: https://gist.github.com/dprimon/280de406b20d2cd65db499a669b17028.
Cheers.

dprimon:
Up

lazka:
Thanks!

dprimon:
I pulled the changes from master; I noticed that I am not listed as a translator in credits.

lazka:
The credits are hardcoded here: https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/const.py

I only update them from time to time (and have forgotten with the last release actually...)

dprimon:
Thanks for you reply.
I have another question. I added support to accellerators to some plugins tabs in my local branch named 4.2-italian. How can I push that branch?

lazka:
* Fork this repo (I see you have an empty repo named quodlibet, you might need to delete that first)
* In you local repo add a new remote (`git remote add fork https://github.com/dprimon/quodlibet.git`)
* Change your branch to `4.2-italian`, and `git push fork`
* Then in the github UI create a PR.

dprimon:
Thank you. I've made a pull request as you described but it's still not being merged. Can you update the const.py to include my name?

lazka:
The pull requests breaks the test suite, so it can't be merged as is. I'll have a look.

lazka:
Also it includes code changes which should go into a stable branch. I'll rebase it onto master

lazka:
Merged your changes in #3405

PR #3405, opened 2020-07-03
Update italian translation

lazka:
See https://github.com/quodlibet/quodlibet/issues/3318#issuecomment-653529886


dprimon:
Thank you!

PR #3438, opened 2020-09-15
Move to Python 3.7 (#3433)

### LINKED ISSUE #3433: Move to newer Python 3.6 / 3.7
The proposed change
--------------------

* Change the _minimum_ version to Python 3.6 or 3.7, and update any builds accordingly

Python 3.5 is long outdated now, and 3.6 and 3.7 bring quite a few benefits (3.9 is current latest at time of writing)

Benefits of this change
-----------------------
* Better typing support, including local variables (not a fan, but hey)
* [f-strings](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498), which are amazingly useful IMO
* The excellent [Pathlib](https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol), which we could use everywhere, but would be a pretty mammoth effort.
* Maintained / non-deprecated versions are important

Any issues this might cause
---------------------------
* Might need to drop oldest Ubuntu / Debian support, though actually even [Buster is 3.7](https://wiki.debian.org/Python), 18.04 is on 3.6
* Windows should be fine AIUI

declension:
* Update versions in docs
* Update Poetry Python version, and lint file
* Fix some more version ranges a bit
* Add missing optional deps
* Re-lock dependencies within ranges
* Fix / update some documentation around versions and package names that was a little confusing
* TODO: work out CI / OS versions testing 

This closes #3433

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
       run: |
         sudo apt update
         sudo apt install -y \
-          dbus-x11 \
+          libdbus-glib-1-dev \
           gettext \
           gir1.2-gst-plugins-base-1.0 \
           gir1.2-gtk-3.0 \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -128,10 +128,12 @@ jobs:
           python3-dev \
           python3-pip \
           python3-venv \
+          python3-wheel \
           xvfb
-
         sudo python3 -m pip install poetry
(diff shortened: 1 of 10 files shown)


lazka:
This likely needs bumping: https://github.com/quodlibet/quodlibet/blob/ae68185a6b81f8d477d9dd56aa89ad50701657d0/quodlibet/const.py#L58

declension:
Thanks!

Issue #3442, opened 2020-09-17
Remote socket keeps dying after some time


LoveIsGrief:
Steps to reproduce
------------------

Unfortunately couldn't find step to reproduce. After running for some time, the remote control socket (`~/.quodlibet/control`) just isn't there anymore.

Expected Output
---------------

- socket should be restarted
- or error message should be shown
- or error log / stack should be printed

Actual Output
-------------

- debug logs have no stacktrace and no output even mentioning the remote socket
- application has to be restarted to remake socket

Test System
-----------

### Which version of Quod Libet?

4.3.0

### Which operating system 

openSUSE Leap 15.2

### If it's audio-related, what back-end?

Additional Information
----------------------

LoveIsGrief:
I haven't grokked unix file-descriptors and pipes yet, but it seems like [`io_add_watch`](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/util/fifo.py#L262) doesn't trigger when the pipe is deleted. The pipe can just disappear and no action will be taken to recreate it.

I'll try and search around to see how that's generally handled, but `poll()`, `select()` and all that jazz is pretty new to me.

LoveIsGrief:
So, I wanted to use `pynotify` to listen for deletions of the FIFO, but opening it with `O_NONBLOCK` hasn't allowed me to do so thus far. So I'm looking at the reason for the use of that flag.

My intuition tells me that [this line](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/util/fifo.py#L42) might be the actual culprit for FIFO deletion. It's not clear to me if multiple writers can be used at the same time on the FIFO and if an exception is raised should it not be, thus triggering the deletion of the existing and functional socket. That still has to be confirmed with a test.

Regardless, I think recreating the FIFO on deletion would be a good way to make similar bugs less noticeable in the future. It would be prudent to add more logs anyway when the FIFO is deleted.

declension:
Thanks for the investigation! And yes that line has a definite smell, with all the swallowed exceptions there.

How do you even find out the socket is gone, if there's nothing in the logs? Presumably the application (or interfacing application) exhibits behaviour that indicates this? Or are you just observing the file itself?

If QL _does_ change behaviour, we could probably detect, trap the exception and recreate if missing. If not, then harder but we could at least check it on schedule at some frequency. +1 to more logs around it too.

LoveIsGrief:
> How do you even find out the socket is gone

[...]

PR #3448, opened 2020-10-02
unix: Don't unlink FIFO on write failure

LoveIsGrief:
Check-list
----------

If one were to think of the relationship between writer and reader as client and server, a write deleting the reader's FIFO is the equivalent of a client killing the server's listening socket.
For that reason the client won't unlinking the FIFO when an error happens.

I haven't been able to track down exactly how the error is triggered that enters the try/catch, but it happens once in a while.
A second PR could monitor the control FIFO for when it's deleted by some external actor and recreate it, but that's out of the scope of this PR.

Related to #3442 - Remote socket keeps dying after some time

### DIFF
--- a/quodlibet/util/fifo.py
+++ b/quodlibet/util/fifo.py
@@ -35,19 +35,11 @@ def _write_fifo(fifo_path, data):
     assert isinstance(data, bytes)
 
     # This will raise if the FIFO doesn't exist or there is no reader
+    fifo = os.open(fifo_path, os.O_WRONLY | os.O_NONBLOCK)
     try:
-        fifo = os.open(fifo_path, os.O_WRONLY | os.O_NONBLOCK)
+        os.close(fifo)
     except OSError:
-        try:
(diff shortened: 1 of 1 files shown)


lazka:
I think the motivation was that if the "server" is stuck this wouldn't allow starting another instance of QL. We've lots of bug reports for that back then.

lazka:
Maybe adding a `from quodlibet.util.print_exc` in there might help to figure out the cause?

LoveIsGrief:
I added the log line and it seems like the FIFO is deleted somewhere else, but I can't find where. It's a quite frustrating bug that I think I'll work around it for now by adding a DBUS interface in an extension with the commands not covered by MPRIS2.

As for this PR, what would you think of making the unlink optional? By that I mean that before the unlink is done, a dialog pops up for the user asking them if they would like to replace the running instance.

LoveIsGrief:
Closing in favor of #3455

PR #3456, opened 2020-10-04
Tag Editor: Add smart replacer for colon delimiters

### LINKED ISSUE #3451: Renaming: add smart conversion of colon delimiters
The proposed change
-------------------
A new checkbox in the rename files window, with high precedence

It would allow conversions of colons (and semicolons maybe) when used in tags (typically titles) to delimit parts.
Suggested / default replacement is ` - ` as it's near-universal and safe on all filesystems. 

### Examples
Affected:
* `ii: allegro.flac` -> `ii - allegro.flac` (instead of `ii_ allegro.flac` from existing Windows-safe converter)
* `Mozart ; Requiem in D minor.ogg` -> `Mozart - Requiem in D minor.ogg`

Whilst _not_ affecting things that don't seem to be delimiters as much:
 *  `Nu:Tone - mix.m4a` -> `Nu:Tone - mix.m4a`
 *  `Elastica - 2:1.mp3` -> `Elastica - 2:1.mp3`

Benefits of this change
-----------------------
Nicer filenames produced than just with the Windows-incompatible-removal option

Any issues this might cause
---------------------------
Not many so long as it's optional

declension:
* Detect `Part: name` in tags and replace with a friendlier ` - ` (avoiding excess underscores for Windows etc)
 * Add some tests
 * Add support for tooltips in these (TODO: more tooltips...)
 * Some autoformatting and related updates

This closes #3451

### DIFF
--- a/quodlibet/qltk/_editutils.py
+++ b/quodlibet/qltk/_editutils.py
@@ -92,10 +92,11 @@ class FilterCheckButton(ConfigCheckButton):
     __gsignals__ = {
         "preview": (GObject.SignalFlags.RUN_LAST, None, ())
         }
+    _tooltip = None
 
     def __init__(self):
         super().__init__(
-            self._label, self._section, self._key)
+            self._label, self._section, self._key, tooltip=self._tooltip)
         try:
             self.set_active(config.getboolean(self._section, self._key))
         except:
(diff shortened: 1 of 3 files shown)


lazka:
lgtm

PR #3464, opened 2020-10-10
Support feedparser 6.0

lazka:
This removes the pre-parse check that was using lots of private API
that is now gone. Ideally this should be implemented using urllib
if it's still required.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.7"
 mutagen = "^1.44"
-feedparser = "^5.2"
+feedparser = "^5.2 || ^6.0"
 pycairo = "^1.19"
 pygobject = "^3.34.0"
 # Optional (mainly plugins)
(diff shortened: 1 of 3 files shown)


lazka:
@declension

pyproject.toml
```diff
@@ -18,7 +18,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.7"
 mutagen = "^1.44"
-feedparser = "^5.2"
+feedparser = "^5.2 || ^6.0"
```
declension, line 21: woah, you can do that? What does it mean - choose whichever works?

declension (approved):

pyproject.toml
lazka: good question, there is only one case documented for specifying the Python version: https://python-poetry.org/docs/dependency-specification/#multiple-constraints-dependencies

It means either of those constrains with npm/composer, and poetry seems to understand it as well.

declension: TIL, thanks :)

lazka: Just checked: When you create a dist with `poetry build` it converts it to `feedparser>=5.2,<7.0` in the generated setup.py.

Issue #3469, opened 2020-10-10
CI: flatpak build fails on lack of pip


declension:
Steps to reproduce
------------------
Push something (working) to master or a branch

Expected Output
---------------
No errors

Actual Output
-------------
```
Run ./.ci/flatpak-test.sh
/usr/bin/python3: No module named pip
Error: Process completed with exit code 1.
```

Example builds
---------------------

* https://github.com/quodlibet/quodlibet/runs/1235914975
* https://github.com/quodlibet/quodlibet/runs/1235609780?check_suite_focus=true

PR #3472, opened 2020-10-11
Disable flatpak CI for now (works around #3469) 

declension:
Seems to be broken on every commit, so better to disable and investigate separately


lazka:
weird, I tested it locally before pushing the flatpak update. I'll check tomorrow. Feel free to merge either way.

declension:
Thanks. 

Aside: I think we should probably fix the double-build thing (see this PR as an example - both `pull_request` and `push` trigger the exact same builds, so we're doubling our usage (and getting queued a lot))

PR #3475, opened 2020-10-14
Make ID3 Replaygain ALL_CAPS

### LINKED ISSUE #3228: TXXX frames are normalized to lowercase but that conflicts with replaygain spec
Steps to reproduce
------------------

Scan example.mp3 with `r128gain` (a minimal and standards-compliant Replaygain tagger). If you run `mutagen-inspect` on the resulting file you will see `TXXX=REPLAYGAIN_*_GAIN` and `TXXX=REPLAYGAIN_*_PEAK`. According to the [spec](https://wiki.hydrogenaud.io/index.php?title=ReplayGain_2.0_specification#ID3v2) this is correct. The case of the tag matters. However, when I edit example.mp3 with Ex Falso and save, the TXXX replaygain tags have been converted to lowercase. A decent amount of players on Linux recognize this but some do not. music on console, GoneMad Music Player are two examples.

Expected Output
---------------

I would expect TXXX frames in id3v2 to not be lowercased. I don't really see the point in doing this as no other tag editor or music player considers TXXX frames to be case insensitive. It is best to not mess with the user's tags unless they explicitly ask for it.

Test System
-----------

Quod Libet 4.2.1
Ex Falso 4.2.1

### Which operating system 

Arch Linux

declension:
When writing ReplayGain tags in ID3, make sure they are in all caps to comply with the spec (apparently)

Fixes #3228

### DIFF
--- a/docs/guide/renaming_files.rst
+++ b/docs/guide/renaming_files.rst
@@ -10,7 +10,7 @@ Quod Libet allows you to rename files based on their tags. In some cases
 you may wish to alter the filename depending on whether some tags are
 present or missing, in addition to their values.
 
-Quod Libet uses _pattern_ syntax for this interaction.
+Quod Libet allows _pattern_ syntax for this interaction, as well as plain text.
 A pattern is some text surrounded by angle brackets, typically containing tags.
 
 A common pattern might be::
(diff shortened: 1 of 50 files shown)


declension:
Not sure what's going on with MacOS here

PR #2685, opened 2018-01-03
fix bug with relative file names in enqueue and enqueue-files

pfps:
Here are the changes to fix the bug with relative file names in the cli arguments enqueue and enqueue-files.  I've also figured out how to do some testing of the commands.

PR #3478, opened 2020-10-18
Fix relative filename queue bug (imported from #2685)

### LINKED ISSUE #2653: bug in enqueue cli command

declension:
Thanks to @pfps

This fixes #2653


declension:
Hmm, @pfps any idea about those MacOS test failures? Seems like something isn't completely absolute somewhere?

pfps:
That's what stymied me when I put in my original PR.  I may be able to see what is going on on my work Mac but it will take at least a few days.

declension:
OK, full circle :laughing: Thanks though, and as this is a public branch feel free to copy it back to your fork or just send me the patch and I'll update

pfps:
I tried running the tsts on my Linux laptop, which exposed an issue with the tests.  Some tests were failing because flake8 was not installed.   There should be something that requires that flake8 is installed, so that the tests are even attempted - just like what happens with python3-pytest.

pfps:
And I'm also getting an error from gettext
```
>                       raise GettextError(stderr)
E                       gdist.gettextutil.GettextError: xgettext: warning: a fallback ITS rule file '/usr/share/gettext-0.21/its/metainfo.its' is used; it may not be in sync with the upstream
E                       xgettext: warning: a fallback ITS rule file '/usr/share/gettext-0.21/its/metainfo.its' is used; it may not be in sync with the upstream
```

This is on a Fedora machine running the master branch.

pfps:
And there are a couple of warnings that should be easy to fix:

```
================================================ warnings summary =================================================
quodlibet/util/importhelper.py:95
  /home/local/SoftwareDownloads/quodlibet/quodlibet/util/importhelper.py:95: DeprecationWarning: Deprecated since Python 3.4. Use importlib.util.find_spec() instead.
    loader = importlib.find_loader(fullname, [path])

/usr/lib64/python3.8/site-packages/dbus/decorators.py:330
  /usr/lib64/python3.8/site-packages/dbus/decorators.py:330: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
    args = inspect.getargspec(func)[0]

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================= 1 failed, 4282 passed, 52 skipped, 2 warnings in 68.52 seconds ==========================
```

Phil305:
@pfps are you trying those tests on Fedora 33 by any chance or 32?

I'm on Fedora 33, and I'm getting that same **gdist.gettextutil.GettextError** error.
I'm gonna take a stab at fixing the *gettext* errors over the weekend, but wondering if you might have any idea what it means?

declension:
@Phil305 what version of gettext are you using?

I think it might be too new, we use `0.19.x` in CI (also locally for me on Ubuntu 20.10 actually) etc

Phil305:
I've got gettext version `0.21-3`.
Unfortunately on Fedora 33, that seems to be the oldest version available, so I can't downgrade...

I'm gonna try to update the failing test so it works with this version of gettext.             

The errors start in the file `tests/test_po.py` in the test `test_potfile_format`.

Specifically the `_create_pot` function in `gdist/gettextutil.py` is getting an error returned from gettext via a subprocess call to `xgettext --from-code=utf-8 --add-comments --files-from=/tmp/tmpneehtcyq --directory=/home/user/projects/quodlibet --output=/tmp/tmpz3j6_knj.pot --force-po --join-existing`

I'm hoping making the test work with `gettext-0.21.x` doesn't cause the tests to fail on other OS...

declension:
Closing this, will need a fresh start, or if someone cares please LMK and we can reopen this

PR #3506, opened 2020-11-30
Support moving library folders (scandirs)

### LINKED ISSUE #1939: Move Library Path
I have stored all my music files on an SMB share in the local network. Since the music collection is pretty large and the network pretty slow, scanning the library over my wifi network literally takes forever. So my idea was to first plug in the drive directly via USB, let QuodLibet scan all my files, put the drive back to the network and then adjust the path to the music files afterwards.

Since there doesn't seem to exist a library migration plugin or something the like, I tried to achieve this by directly manipulating the song list of QuodLibet. What I did was that I took the `~/.quodlibet/songs` file and replaced all occurrences of `/Old/Path/` with `/New/Network/Path` and also changed the scan directory in `~/quodlibet/config` to match the new library path.

This doesn't seem to do the trick. Whenever I open QuodLibet after editing the above files, the song list seems to be empty. Also, I get a `~/.quodlibet/songs.notvalid` file, so QuodLibet seems to remember the old file paths somewhere.

What have I missed? Is the whole approach possibly stupid?

QuodLibet 3.6.1 on OS X 10.11.4 El Capitan

declension:
* Add a `move_dir()` (generator) method to `FileLibrary`
 * ...and to `Librarian` likewise
 * Add a button in Preferences -> Library to select an individual scandir to migrate
 * This then calls a Task that runs in the background
 * Add some tests around this
 * Add a little bit of documentation around this
 * TODO: more real-world testing, but it consistently seems to work for me locally, with fairly simple scenarios though

 This fixes #1939

### DIFF
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -16,7 +16,6 @@ least useful but most content-agnostic.
 import os
 import shutil
 import time
-from os.path import realpath
 from pathlib import Path
 from typing import Set, Optional, Generator
 
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -16,26 +16,23 @@ least useful but most content-agnostic.
 import os
 import shutil
 import time
-from os.path import realpath
-from pathlib import Path
-from typing import Set, Optional, Generator
+from typing import Set, Optional
 
 from gi.repository import GObject
-from senf import fsn2text, fsnative, expanduser
 
 from quodlibet import _
+from quodlibet import formats
+from quodlibet import util
 from quodlibet.formats import (MusicFile, AudioFileError, load_audio_files,
-                               dump_audio_files, SerializationError, AudioFile)
-from quodlibet.query import Query
+                               dump_audio_files, SerializationError)
 from quodlibet.qltk.notif import Task
+from quodlibet.query import Query
 from quodlibet.util.atomic import atomic_save
 from quodlibet.util.collection import Album
 from quodlibet.util.collections import DictMixin
-from quodlibet import util
-from quodlibet import formats
 from quodlibet.util.dprint import print_d, print_w
-from quodlibet.util.path import (unexpand, mkdir, normalize_path, ishidden,
-                                 ismount)
+from quodlibet.util.path import unexpand, mkdir, normalize_path, ishidden, ismount
+from senf import fsn2text, fsnative
 
 
 class Library(GObject.GObject, DictMixin):
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -856,10 +856,8 @@ class FileLibrary(PicklingLibrary):
                 path = Path(key)
                 if old_path in path.parents:
                     # TODO: more Pathlib-friendly dir replacement...
-                    new_key = normalize_path(key.replace(str(old_path),
-                                                         str(new_path),
-                                                         1),
-                                             canonicalise=False)
+                    new_key = key.replace(str(old_path), str(new_path), 1)
+                    new_key = normalize_path(new_key, canonicalise=False)
                     if new_key == key:
                         print_w(f"Substitution failed for {key!r}")
                     if self.move_song(song, new_key):
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -837,10 +837,12 @@ class FileLibrary(PicklingLibrary):
           4. Move audio files: old_root -> new_path
 
         """
-        old_root = realpath(expanduser(old_root))
-        new_root = expanduser(new_root)  # type: ignore
-        if not Path(fsn2text(new_root)).is_dir():
-            raise ValueError(f"{new_root!r} is not a directory")
+        old_root = Path(realpath(expanduser(old_root)))
+        new_root = Path(expanduser(new_root))  # type: ignore
+        if not old_root.is_dir():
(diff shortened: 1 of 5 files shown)


quodlibet/library/libraries.py
```diff
@@ -823,6 +824,74 @@ def remove_masked(self, mount_point):
...
+            raise ValueError(f"Destination {new_path!r} is not a directory")
+        print_d(f"{self._name}: checking {len(self.values())} item(s) for {old_path!r}")
+        missing: Set[AudioFile] = set()
+        changed = set()
+        total = len(self)
+        if not total:
+            return
+        with Task(_("Library"), _("Moving library files")) as task:
+            yield
+            for i, song in enumerate(list(self.values())):
```
CreamyCookie, line 853: That `list` call is necessary, because of `self.move_song`, right?

quodlibet/library/libraries.py
```diff
@@ -823,6 +824,74 @@ def remove_masked(self, mount_point):
...
+                task.update(i / total)
+                key = normalize_path(song.key)
+                path = Path(key)
+                if old_path in path.parents:
+                    # TODO: more Pathlib-friendly dir replacement...
+                    new_key = normalize_path(key.replace(str(old_path),
+                                                         str(new_path),
+                                                         1),
```
CreamyCookie, line 861: I think the `key.replace` part would look better on a new line, but that's just my preference for limiting multi line statements.

tests/test_library_libraries.py
```diff
@@ -113,7 +114,7 @@ def setUp(self):
     def test_add(self):
         self.library.add(self.Frange(12))
         self.failUnlessEqual(self.added, self.Frange(12))
-        del(self.added[:])
+        del (self.added[:])
```
CreamyCookie, line 117: Another `del()` case

tests/test_library_libraries.py
```diff
@@ -342,7 +343,7 @@ def test_rename_changed(self):
 
     def test_tag_values(self):
         self.library.add(self.Frange(30))
-        del(self.added[:])
+        del (self.added[:])
```
CreamyCookie, line 346: I think these braces should be removed. (`del` is a statement, not a function)

tests/test_library_libraries.py
CreamyCookie, line 117: I think these braces should be removed. (`del` is a statement, not a function)

CreamyCookie:
LGTM (though I have no knowledge of the Library classes)

tests/test_library_libraries.py
declension: Done (and a few more)

quodlibet/library/libraries.py
declension: :+1:

declension:
@CreamyCookie did you try moving any library paths locally?

declension:
OK, tried again locally twice - some weird stuff around re-scanning and then masking new that took ages (but worked fine eventually). Anyway think that was me re-adding my old masked path.

Second time it was as previous times (fast, playlists / metadata all fine). 

Seem to be no objections so merging now and we can file issues as normal if they come up. Thanks @CreamyCookie for reviewing :+1:

CreamyCookie:
@declension Ah sorry, was really busy. Happy new year! 

I hadn't tried, but I have tried now (from `master`, using a completely fresh install without any existing `.quodlibet` files, `mutagen` and `feedparser` installed via pip3), and it seems something is going wrong.. 

Here's the [log](https://gist.github.com/CreamyCookie/d7741e1be5635901fd78b18665b8782b) and what I did:

1. Start Quodlibet for first time
2. The new user dialog asks me to add a library folder, and I do so
3. Switch to Album List view
4. Scan for new files
5. Close and restart QL
6. Scan again. Nothing happen
7. Close and restart QL
8. Try rebuild library in settings but it doesn't help.

Now, whenever I start QL, the albums quickly appear and disappear in the Album List view.

declension:
Thanks for testing @CreamyCookie.

Did you move the files?

Perhaps it's not obvious enough - but you have to have _already_ moved the audio files (or perhaps just after I guess) - i.e. quodlibet won't do it for you (this I deemed _way_ too invasive for my liking).

CreamyCookie:
Oh, I see! I think in that case a message dialog would be useful. It could also include the warning about the backup which is currently shown after selecting a directory.

btw. I just moved it and it works. :+1: 

---

It seems there's no check for whether a scan directory is selected, so this error comes up when just pressing "Move" after opening the settings dialog. I think making the button disabled before a row is selected would also be reasonable.

```python-traceback
      File "/home/anon/Desktop/quodlibet/quodlibet/qltk/scanbox.py", line 118, in __move
        base_dir = rows[0][0]
    IndexError: list index out of range
```

declension:
> I think in that case a message dialog would be useful. It could also include the warning about the backup which is currently shown after selecting a directory.

Cool. Did you try the very latest master - I added a warning / confirmation message there (including backup actually), but I'll update it there perhaps

CreamyCookie:
@declension I just checked and my local last commit is the same (https://github.com/quodlibet/quodlibet/commit/de362593832aceceb68075edfbe01c9724a43008). 

> I added a warning / confirmation message there

That's the one I was referring to when I talked about showing the dialog before asking for a directory (and as I mentioned, adding a message that the user has to manually move the file). Although I see now that I could have been more clear about that in my comment.

declension:
OK makes sense; anyway I'm updating it now

Issue #3467, opened 2020-10-10
Use monospace font for search bar?


declension:
The proposed change
-------------------
Switch the search bar(s) to use a monospaced (fixed-width) font, as they're basically code, rather than free text.

Benefits of this change
-----------------------
* Much easier to read when using anything apart from basic searches, e.g. `artist=/.+ and the .+/` rather than "artist=/.+ and the .+/" etc... or even just helping make quotes stand out more: `album = 'Master of Puppets'c`
* Indicates to users more naturally that this _isn't_ just free text

Any issues this might cause
---------------------------
* Inconsistent look-and-feel
* Annoyances?

PR #3517, opened 2020-12-06
Restyle the search query

### LINKED ISSUE #3515: Query colours are not consistent across themes
Steps to reproduce
------------------
Change query text from plain to invalid to valid
Repeat in both dark and light themes

Expected Output
---------------
Vaguely equally visible across themes
Good colour (hue) spacing between green and red, similar saturation, and equivalent brightness (given sensitivity of eyes)  

Actual Output
-------------
Hard to see greens and reds in dark theme compared to the detault text colour, but looks a bit ugly in light theme too.

Test System
-----------

4.4

declension:
* Use monospace by default (#3467)
 * ...but controlled by prefs boolean, which is now exposed in advanced prefs :up:
 * Validators: use colours that adapt according to the default (from theme) (#3515) by mixing them together. Makes for some interesting colours but I definitely prefer it, and it's now configurable too.
 * Increase the font size by default. I tried _many_ looks here and despite initial weirdness this default feels like a good balance between usability and awkwardness, especially given how important the searches are in QL...
 * But make _this_ configurable too, with another advanced prefs linking, so that users can change this back
 * TODO: some kind of docs around this?

### Light
Screenshot from 2020-12-06 16-24-51
Screenshot from 2020-12-06 16-24-48

### Dark
Screenshot from 2020-12-06 16-24-31
Screenshot from 2020-12-06 16-24-24

### Light
Screenshot from 2020-12-05 19-48-13
Screenshot from 2020-12-05 19-47-51
Screenshot from 2020-12-05 19-47-45

### Dark
Screenshot from 2020-12-06 16-22-34
Screenshot from 2020-12-06 16-14-03
Screenshot from 2020-12-06 16-13-53

Closes #3515, #3467

### DIFF
--- a/quodlibet/qltk/entry.py
+++ b/quodlibet/qltk/entry.py
@@ -1,5 +1,6 @@
 # Copyright 2005 Joe Wreschnig, Michael Urman
 #           2011, 2012 Christoph Reiter
+#           2020 Nick Boultbee
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -10,7 +11,7 @@ import math
 
 from gi.repository import Gtk, GObject, Gdk, Gio, Pango
 
-from quodlibet import _
+from quodlibet import _, config
 from quodlibet.qltk import is_accel, add_fake_accel
 from quodlibet.qltk.x import SeparatorMenuItem, MenuItem
 from quodlibet.qltk import Icons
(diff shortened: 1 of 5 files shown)


declension:
Any objections @lazka @Meriipu @CreamyCookie @frestr (and others)?

lazka:
The default color mixing looks nice. Not a fan of all the other stuff (monospace, font size, config options), but if you feel strongly about it..

quodlibet/qltk/entry.py
```diff
@@ -238,30 +239,46 @@ class ClearEntry(UndoEntry, ClearEntryMixin):
...
+        return col
+
+    def _set_color(self, _widget, validator):
+        def mix(src: Gdk.RGBA, dest: Gdk.RGBA, ratio: float):
+            ratio = min(1.0, max(0.0, ratio))
+            inv = 1.0 - ratio
+            return Gdk.RGBA(inv * src.red + ratio * dest.red,
+                            inv * src.green + ratio * dest.green,
+                            inv * src.blue + ratio * dest.blue,
+                            inv * src.alpha + ratio * dest.alpha)
```
CreamyCookie, line 271: I think this would be better in some kind of Gdk / color utils module.

CreamyCookie (commented):
The colors look very nice!

The contrast in your new (dark theme) screenshot is far better, and now meets the WCAG AA contrast requirement: https://webaim.org/resources/contrastchecker/?fcolor=EF9FA5&bcolor=343434 

The old one did not: https://webaim.org/resources/contrastchecker/?fcolor=B80303&bcolor=343434

(WCAG: I only recently learned about the [Web Content Accessibility Guidelines](https://en.wikipedia.org/wiki/Web_Content_Accessibility_Guidelines). Basically it provides requirements for the contrast between colors of different elements, among many other things. Tools for checking whether the WCAG contrast requirement is met are built into Firefox and Chrome, for example.)

----

On one hand, I feel like monospace makes sense for a query language. And it also might make it easier to catch errors with missing braces or other characters.

On the other hand, going by the screenshots, the font size and type change seems to make the interface look more cluttered / noisy. (Could be that it's just the font size. I'm not sure.) Also, and this might be a weird point, I think people associate monospace fonts with tech stuff, and so the search bar might look (even) less approachable to non-tech people and so they might experiment less with queries.

I think it's probably better to have them disabled by default.

quodlibet/qltk/entry.py
declension: Done!

CreamyCookie (commented):
:+1:

declension:
Thanks all

There are a few monospace fans on #3467 and I think I prefer it (slightly) to use, though perhaps not to look at so leaving the option there for personal taste. 

If you want a pathological QL search case for variable-width fonts, try:`|(/l./,/:/)` :scream: 

So I'll switch the defaults back to normal size and font and proceed :+1:

PR #3529, opened 2021-01-03
Library rescan: Yield and signal less often

declension:
* Also align with 100 to prevent lots of weird (_101 files changes_) messages
* 10k tracks full rescan would have 100 signals and 2000 yields. Maybe 50 and 500 is enough, especially given faster machines now.


lazka:
fine with me. note that you trade laggy UI vs overall performance with this.

declension:
Thanks, and yes. My thinking was that, 15 years later, machines are now faster (but humans not) the overall lag change should be lower than it looks.

lazka:
test

PR #3548, opened 2021-01-29
Change None to empty string to fix TypeError in missing.py

### LINKED ISSUE #3547: @(missing) query trigger errors in the Album List Browser
Steps to reproduce
------------------
write **@(missing)** query in the Album List Browser while the Missing Query plugin is enabled

Expected Output
---------------
query doesn't trigger an error

Actual Output
-------------
TypeError: 'NoneType' object is not subscriptable

Test System
-----------

commit 08360e8ee3462270c7c4d31dca987855219853ca

### Which operating system 
Ubuntu 20.04.1 LTS

Additional Information
----------------------
This bug is my fault, should have been more careful the first time I tried to fix this issue, I will do a pull request ASAP.

Ironbalden:
Fixes #3547

Check-list
----------

 * [ x] There is a linked issue discussing the motivations for this feature or bugfix
 * [ x] Performance seems to be comparable or better than current `master`

None value is problematic for the Album based browsers, using an empty string solves the issue and doesn't change behaviour

### DIFF
--- a/quodlibet/ext/query/missing.py
+++ b/quodlibet/ext/query/missing.py
@@ -20,7 +20,7 @@ class MissingQuery(QueryPlugin, PluginConfigMixin):
     usage = "<b><tt>@(missing: artist)</tt></b>"
 
     def search(self, data, body):
-        val = data.get(body.strip() if body else None, None)
+        val = data.get(body.strip() if body else '', None)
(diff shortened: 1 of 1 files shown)


declension:
Thanks!

Issue #3549, opened 2021-01-31
Plugin to copy / move songs from right-click menu


mkalinski:
The proposed change
-------------------

Add to songs' right-click menu options to copy or move files directly from song browsers. The option to delete files is already there.

I created a plugin that adds this functionality. The file system operations are executed in a worker thread and reported via text log in a new window.

Benefits of this change
-----------------------

Song browsers offer filtering and sorting options that are absent from ordinary file managers. Right now, the only way to move and copy songs is through "show in file manager", which usually just shows the first file of the selected ones, and offers no music-aware filtering.

Also, if a song file is moved via a file manager, quodlibet will "lose track" of it and forget QL-specific data, like ratings, play counts, or playlist membership. When using the proposed plugin to move a song, the data is kept.

Any issues this might cause
---------------------------

Performance: the operations are performed in a different thread, and most of the time is spent on IO operations (GIL is released), so it should not affect performance.

Migrations: none needed.

davidSchlosser:
One can use the song 'Edit tags' menu, 'Rename files' tab to move songs to another new or existing directory. I do that often, with multiple songs, and like that it gives me access to tag values as part of the directory & filename. It doesn't copy though -  don't have a use-case for copying files.

Ironbalden:
@davidSchlosser copy music files around is extremely useful to prepare an usb stick for the car or an sd for a music player, you can do that with playlists but are an extra step, sometimes you just want to stuff the stick with random songs and get over with it. Fact is, you can do that with drag and drop, at least under linux, so it's kind of a redundant feature, but hey more options are never bad.
For actually moving files, 'Rename Files' is powerful enough for basically every realistic case I can think of, maybe we can discuss if it can be beneficial to make it more visible.

mkalinski:
You're right about "Rename Files". I somehow never thought about using full paths of directories there. It's described in the documentation too, so that's on me.

Drag and drop I would have never discovered, and the documentation doesn't seem o mention the effect of copying the file when dropped into an external file browser.

Regardless, it seems that the functionality I wanted from this plugin is technically already covered. I'm not sure now if it's worth it to include the plugin, but like you said, it's more options, and in one place.

declension:
We should fix Drag-n-drop and document this (and moving files with renaming) a bit better.... so converting this issue into that

petitminion:
[...]

PR #3550, opened 2021-01-31
Add FileOperationsPlugin

mkalinski:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix: #3549
 * [ ] I've added / updated documentation for any user-facing features: individual plugins don't seem to have a dedicated documentation section

This change is adding a songsmenu plugin that provides two menu options: `Copy` and `Move`.

These options open a file chooser to select destination, and then copy or move the selected songs to the destination.

Moving / copying is done in a separate thread.

A window is opened with a progress bar, log of each file copied / moved and any exceptions that might occur, and a button to abort the copy / move.

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -119,6 +119,7 @@ quodlibet/ext/songsmenu/duplicates.py
 quodlibet/ext/songsmenu/editplaycount.py
 quodlibet/ext/songsmenu/embedded.py
 quodlibet/ext/songsmenu/exact_rating.py
+quodlibet/ext/songsmenu/file_operations.py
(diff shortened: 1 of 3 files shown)


declension:
Did you try doing via  _Custom Commands_ ? It seems like it should be doable

mkalinski:
@declension It's technically doable via Custom Commands but it's cumbersome. The destination can only be passed as parameter, which needs to be manually typed as text input every time.

There are good alternatives to this plugin proposed in #3549 though.

declension:
Closing in favour of #3459

Issue #3551, opened 2021-02-01
Support Pipewire output


BryanQuigley:
The proposed change
-------------------
Add Pipewire option like exists for JACK audio

Benefits of this change
-----------------------
Let's users use Pipewire directly if they want to for audio. The plan is for Pipewire is eventually going to replace both Jack and PulseAudio.

Any issues this might cause
---------------------------
The transition from Jack/PulseAudio to Pipewire is in it's early stages. Volume/Output/Input management is much easier with existing pulseaudio tools at the moment.

declension:
Doesn't Pipewire provide a drop-in Jack server replacement though?

BryanQuigley:
Yes, PipeWire can handle apps that use ALSA, JACK, and PulseAudio transparently.  (I've only tested pipewire-pulse with quod libet personally though and it just worked).  This is about giving an option to use PipeWire directly.

declension:
I'm now not sure how this will work - Gstreamer itself doesn't look like it'll support pipewire directly [AFAICT from plugins](https://gstreamer.freedesktop.org/documentation/plugins_doc.html?gi-language=python) (and QL depends on Gstreamer)

declension:
Closing unless this changes or I've missed something in which case we can reopen

PR #3552, opened 2021-02-01
Adds Pipewire Sink

BryanQuigley:
Uses the same strategy uses if you want to use Jack instead
of PulseAudio. Config option set to disabled by default.

Check-list
----------

 * [X] There is a linked issue discussing the motivations for this feature or bugfix #3551
 * [] Unit tests have been added where possible
 * [TODO] I've added / updated documentation for any user-facing features.

### DIFF
--- a/quodlibet/player/gstbe/util.py
+++ b/quodlibet/player/gstbe/util.py
@@ -37,6 +37,8 @@ class AudioSinks(Enum):
     JACK = "jackaudiosink"
     """from plugins-good"""
 
+    PIPEWIRE = "pipewiresink"
+
     WASAPI = "wasapisink"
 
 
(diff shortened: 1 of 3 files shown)


BryanQuigley:
Hmm.. it always works at the start.. but something makes it stop working..  Need to investigate further.

BryanQuigley:
I somehow broke Pipewire on my system, but still think this may be useful in the future.

declension:
I think one easy fix is the extra newline:
> AssertionError: /home/runner/work/quodlibet/quodlibet/quodlibet/player/gstbe/util.py:78:1: E302 expected 2 blank lines, found 1

quodlibet/player/gstbe/util.py
```diff
@@ -120,6 +134,9 @@ def sink_options():
         if config.getboolean("player", "gst_use_jack") and jack_is_running():
```
declension, line 134: Might this not be triggered if Pipewire is Jack-compatible?

declension:
It might be that a drop-down for sink (e.g. Default / Jack / Pipewire) is now sensible, I was thinking that before already...

BryanQuigley:
I think it won't be long before PipeWire makes sense as the default (both next versions of Fedora and Ubuntu may have it).  Having said that, I don't have PipeWire or Quod Libet with PipeWire working as reliably as I'd like _yet_.

BryanQuigley:
Fallback doesn't work as I expected - and from what I can some of the gstreamer bits in pipewire were just considered not ready yet per https://bugzilla.redhat.com/show_bug.cgi?id=1884260. I do plan to revisit this when it's more stable though.

PR #3553, opened 2021-02-02
Docs: Fix typos

freddii:
fixed typos  in code comments

### DIFF
--- a/tests/test_browsers_iradio.py
+++ b/tests/test_browsers_iradio.py
@@ -170,7 +170,7 @@ class TIRFile(TestCase):
         assert all(ret)
         assert self.received, "No stations received from %s" % url
         assert len(self.received) > 100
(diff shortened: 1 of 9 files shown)


tests/test_browsers_iradio.py
```diff
@@ -170,7 +170,7 @@ def cb(data):
         assert all(ret)
         assert self.received, "No stations received from %s" % url
         assert len(self.received) > 100
-        # TODO: some more targetted tests
+        # TODO: some more targeted tests
```
declension, line 173: This one is actually debatable / locale-dependent ([wiktionary targetted](https://en.wiktionary.org/wiki/targetted)) but seems reasonable to use the more common :+1:

declension:
Thanks!

PR #3558, opened 2021-02-14
Create GitHub Actions automated testing for Fedora 33 (resolves #3542)

### LINKED ISSUE #3542: Build automation for Fedora
The proposed change
-------------------
Some automation is added for building on Fedora-like systems (ideally Github Actions CI)

Benefits of this change
-----------------------
Contributors on these systems have fewer problems

Any issues this might cause
---------------------------
GH actions only support Ubuntu, so might need to be Docker-in-Docker....

Phil305:
Check-list
----------

 * [x ] There is a linked issue discussing the motivations for this feature or bugfix

Automation is added for building on Fedora 33 in Github Actions CI, to make it easier for Fedora users to contibute ( i.e. ..me :^]  )

### DIFF
--- a/.github/workflows/fedora33/Dockerfile
+++ b/.github/workflows/fedora33/Dockerfile
@@ -1,0 +1,42 @@
+FROM fedora:33
+
+# DNF dependencies required before poetry can install virtualenv dependencies
+RUN dnf -y install \
+      poetry \
+      cairo-devel \
+      gcc \
+      python3-devel \
+      dbus-devel \
+      gobject-introspection-devel \
+      cairo-gobject-devel
+
+# Dependencies for tests
+RUN dnf -y install \
+      Xvfb \
+      # Contains 'GdkPixbuf' namespace (used by 'gi') \
+      gtk3 \
+      # Contains 'Gst' module \
+      gstreamer1 \
+      # Contains GStreamer replaygain plugin \
+      gstreamer1-plugins-good \
+      # SVG support (for 'Timage_support.test_create_pixbuf' test) \
+      librsvg2 \
+      dbus-x11
+
+RUN dnf clean all
+
+# Create user (required for write access tests in 'tests/test_formats__audio.py' to pass)
+ARG HOST_USER_ID=5555
+ENV HOST_USER_ID ${HOST_USER_ID}
+RUN useradd -u $HOST_USER_ID -ms /bin/bash user
+
+# Required to prevent pytest from running into permissions errors when creating
+# cache files in GITHUB_WORKSPACE as 'user'
+ENV PYTHONDONTWRITEBYTECODE 1
+
+# required for translation tests that use 'gettext'
+ENV LANG C.UTF-8
+
+# run tests
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/workflows/fedora33/action.yml
+++ b/.github/workflows/fedora33/action.yml
@@ -1,0 +1,6 @@
+# action.yml
+name: Fedora 33
+description: Run tests for Fedora 33
+runs:
+  using: docker
+  image: Dockerfile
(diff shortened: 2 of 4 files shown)


Phil305:
Not sure what the situation is with the *msys2* test... I didnt touch that...

declension:
Yes that mys2 is consistently failing now, `setuptools` has gone for some reason, I'll raise an issue

.gitignore
```diff
@@ -60,3 +60,4 @@ Pipfile.lock
 # Common virtualenvs
 .venv/
 venv/
+tags
```
declension, line 63: Is this intentional?

.github/workflows/fedora33/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM fedora:33
+
+# install dependencies
+RUN dnf -y install \
+      https://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm \
```
declension, line 5: Can we add a comment why we're using these (slightly) non-standard sources?

.github/workflows/fedora33/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+      dbus-x11 \
+      glib2-devel \
+      gobject-introspection-devel \
+      gstreamer1 \
+      gstreamer1-libav \
+      gstreamer1-plugins-{bad-\*,good-\*,base} \
+      gtk3 \
+      poetry \
+      python3-devel \
+      python3-feedparser \
```
declension, line 22: I think most of these we can / should get from Poetry?

declension (commented):
Looking good, couple of comments

declension:
Update: `msys2` seems fine now.

.gitignore
Phil305: I use Ctags  with Vin to jump to function definitions, it makes a **tags** file, so I dropped the exclusion in my **.gitignore**.

I can remove it though if may cause problems, or if it shouldnt be committed.

.github/workflows/fedora33/Dockerfile
Phil305: Yeah no prob, ill put which required packages are in those repos

.github/workflows/fedora33/Dockerfile
Phil305: I believe I installed all of those packages for specific errors I got when I was trying to run **poetry install**

I'll double check though

.github/workflows/fedora33/Dockerfile
Phil305: Ends up a lot of them weren't necessary;

Seems there was some Docker weirdness happening on my end (I've been testing in standalone containers, since pushing to Github all the time to test got... old, quick)

.gitignore
Phil305: Yeah, I'm gonna go ahead and remove **tags**
Just in case some day a valid **tags** file is created, i dont want anyone to have to suffer by accident :)

.github/workflows/fedora33/Dockerfile
Phil305: I've been using the RPM Fusion repos locally since reading one of the Fedora guides for getting Audio/Video working:

https://docs.fedoraproject.org/en-US/quick-docs/assembly_installing-plugins-for-playing-movies-and-music/

Ends up I didnt need it here (**gstreamer1-plugins-good**, have the *replaygain* plugin that one of the tests uses, and it's in the official Fedora 33 repo)

.github/workflows/fedora33/Dockerfile
Phil305: You were right.

I'm cut down on the imports by alot now!

I tested this via Docker image directly, with the **nektos/act** tool, and it's passing in GH Actions,
so hopefully these tests will prove to not be flaky

declension:
Thanks!

Issue #3559, opened 2021-02-14
External plugins not showing in MacOS / plugin dev page


BraveSentry:
I tried to include an external plugin by creating ~/.config/quodlibet/plugins and putting the plugin.py file in that folder (see [here](https://github.com/quodlibet/quodlibet/issues/3426#issuecomment-778738322). That worked in Linux Mint, but not on the Hackintosh on the same machine. Is the location for plugins different on Mac? If so, it should say so on [the plugin dev page](https://quodlibet-frestr.readthedocs.io/en/latest/development/plugins.html).

One related minor issue: The links at the bottom of [the plugin dev page](https://quodlibet-frestr.readthedocs.io/en/latest/development/plugins.html) are broken.

declension:
We've pretty much deprecated the plugins-in-home-dir some time ago, when plugins got moved into the main module. 

It does still work in some cases (clearly) but it's always recommended to run from source if you're developing plugins now.

I'll update the docs around this actually, though I don't know anything about how the MacOS build bundles the plugins now. 

:information_source:  those last links (to plugin dev page) are not the official build of the docs, probably why the link is dead.

Issue #3426, opened 2020-09-04
Sync ratings across OSs on a multiboot machine


BraveSentry:
Hey,
I have a laptop with a Hackintosh, a Windows 10 and a Linux Mint installation. I have installed Quod Libet on all of them, and all three are pointed to the same disk partition with my music on it.

Now I rely heavily on ratings when it comes to sorting my music and I´m looking for a way to sync ratings between all three systems.

I tried the "Update Tags in Files" plugin but then can´t find a way to import these tags on the other OSs.

I also tried the "Export Metadata" plugin, but that gives me one .tags file for each song, with no possibility to do that to multiple files at once, let alone import the tags in the other OSs.

When using the "migrate metadata" plugin, the behaviour seems erratic to me: I mark a couple of tracks, press "copy", then change one songs rating, then mark the same tracks again, then paste, and the rating changes for multiple songs, and not necessarily to the rating it had, when I copied the metadata.

Any suggestions on what else I could try? Or ways I could make the things I have already tried work?

CreamyCookie:
Ratings are stored in QL database, and while they can be stored in the files itself (I do), I think reading them isn't automatic and time intensive, so that you wouldn't want to do that every time you start QL.

I think one way that could work is copying the `.quodlibet/songs` file (QL database of songs) to an external exFAT-formatted disk?  (If you want to try, you likely need to install the format: `sudo apt-get install exfat-fuse exfat-utils`)

Then when you start, you'd need to mount the disk, copy (and replace) the `songs` file to the platform-specific QuodLibet directory. On exit, you'd copy (and replace) the `songs` file on the external disk.

I don't think it would work well to keep the whole QL directory on an external disk (e.g. file paths in playlists differ between platforms), but I could be wrong.

Please make backup of all relevant files before trying this.

CreamyCookie:
Ratings are stored in QL database, and while they can be stored in the files itself (I do), reading them isn't automatic and checking every single file would take too much time.

Sadly, the songs database doesn't seem to be platform independent, so copying that doesn't seem to be an option.

The best way I can think of is via plugins:

1. Download my auto_update_tags_in_files.py plugin: https://gist.github.com/CreamyCookie/60f007490486d09a174a87c0f7486954 (and change it's config to your liking)
2. Modify it so that it writes a file `{song filename}.changed` in the directory of the album/song whenever ratings are written to their files (in `_write_tags_to_files`)
[...]

PR #3562, opened 2021-02-16
Update plugins.rst to include locations on MacOS.

BraveSentry:
I´ve included the location of plugins for older versions of QL on MacOS. This info would have saved me some headache regarding  #3426 and #3559. The locattion still works for me and this information might be helpful for others as well.


declension:
Thanks!

PR #3568, opened 2021-03-02
Made tag scan multithreaded.

ewtoombs:
This was done to reduce the latency effects of network filesystems. I have
observed about a 3x speedup in the scan rate. That's over wifi. If I was
scanning over a wide area network, I would expect the speedup to be many times
that.

Check-list
----------

### DIFF
--- a/quodlibet/library/libraries.py
+++ b/quodlibet/library/libraries.py
@@ -18,6 +18,8 @@ import shutil
 import time
 from pathlib import Path
 from typing import Set, Optional, Generator
+from threading import Thread
+from queue import Queue
 
 from gi.repository import GObject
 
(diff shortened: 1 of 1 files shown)


lazka:
This assumes that MusicFile() is thread-safe, which it isn't. I think it calls into logging and config (any more?), so both of those need to be checked/marked for thread safety first.

ewtoombs:
> This assumes that MusicFile() is thread-safe, which it isn't. I think it calls into logging and config (any more?), so both of those need to be checked/marked for thread safety first.

It isn't? It seemed to work OK when I tried it. I knew it reads some globals, but ... oh wait. Oh, no, if somebody *else* modifies the config, MusicFile()'s read could break. God damn it.

Well, I don't know enough about the code base to be able to answer this question of thread safety, then.

lazka:
yeah, Python is very forgiving in that area because of the GIL, und config changes are rare.

CreamyCookie:
@ewtoombs Thanks for the PR. I think this would be a cool addition (if MusicFile is made thread safe).

declension:
OK, unmergeable for now then I guess. Re-open / repush if thread-safe fixes are found.

PR #3596, opened 2021-03-29
Provide empty mpris:artUrl if cover art is not present

luk1337:
Check-list
----------

This fixes an issue on KDE connect app that causes previous
cover art to get shown whenever currently playing song has
no artwork but the previous one does.


luk1337:
hm, maybe i should fix it in kde connect...

luk1337:
https://invent.kde.org/network/kdeconnect-android/-/blob/master/src/org/kde/kdeconnect/Plugins/MprisPlugin/MprisPlugin.java#L270-312 looks like they are falling back to cached values here, will try to fix it upstream.

PR #3607, opened 2021-04-10
Update universe (but lock)

declension:
* Lock to various new minor versions
 * Update range for Mutagen (we really do want a new version I'd say)
 * Upgrade MyPy, it's bringing a lot of new ~~pain~~ help
 * Actively exclude vendored packages from MyPy with `exclude`
 * Fix a weird senf import now detected as a problem
 * Fix version to 4.4.1-pre (I think this is right?)


lazka:
lgtm

Issue #1090, opened 2015-03-15
Improve initial user experience


lazka:
Original [issue 1090](https://code.google.com/p/quodlibet/issues/detail?id=1090) created by nick.boultbee on 2012-12-26T11:49:38.000Z:

Having just tested a fresh installation on Windows, I saw that QL appears unfriendly &quot;out of the box&quot;. 

This is also the impression a colleague gave me, and only after I'd described some of the more powerful features did he say he'd go back and have another look - i.e. the user interface presentation / defaults had put off even a very technical user.

Some initial ides:
- Increase the default height and width. Screens have got much bigger in ~8 years.
- Make a simpler browser the default (e.g. album list or search): looks cleaner too.
- Add Artist to the column list by default - was quite surprised to not see it there.
- Enable some &quot;core&quot; plugins by default (Album Art, Replay Gain, Browse Folders, MusicBrainz lookup etc)
- Provide better notifications of missing Python modules. On enabling plugins, I found 13 errors due to unavailable modules, despite not wanting most of these plugins anyway. But DBus stands out here particularly.
- Disable *nix-only plugins for Windows.

Probably there are a few others; add here.

declension:
Not sure, but I think there's still some work to do on this.

CreamyCookie:
Idea from @xhc1: show a popup on first use (or search) about what places it will search in (artist, album, title). Users might expect the search to find songs by their filenames.

brandonkirsch:
I'm new to Quod Libet and the only thing I didn't like about my initial experience is that the `MPRIS D-Bus Support` plugin is not enabled by default.  This plugin enables multimedia playback controls for my KDE Linux desktop.

I would suggest turning it on by default for Linux desktop users.

PR #3626, opened 2021-04-15
Made avg and bav duration-weighted and added unweighted version of those

### LINKED ISSUE #3048: Track length is not considered when calculating average album rating
Simple case: album of two tracks. 1st track is `20` minutes long, rated `0.4`. 2nd track is `4` minutes long, rated `0.6`.
Currently that gives a simple `0.5` average rating for the album. But one would expect that longer piece's rating puts more influence on average (i.e. `(5*0.4 + 1*0.6) / 6 = 0.43` for above case). For example, I don't think last track should affect average [U.F.Orb](https://www.discogs.com/The-Orb-UFOrb/release/13274) album rating much.
Such an improvement in calculation would allow to have more accurate evaluation for albums with short interludes or few very long pieces (see even more... [example 1](https://www.discogs.com/Boards-Of-Canada-Geogaddi/release/24432) or [example 2](https://www.discogs.com/Goldie-Saturnz-Return/release/2349297)).

jgandert:
Check-list
----------

 * [x] Performance seems to be comparable ~~or better~~ than current `master`

Makes it so that long songs have more influence on, for example, the average rating than short ones.

Closes #3048 (which includes a discussion about this)

Reasoning with example for smoothing is [here](https://github.com/quodlibet/quodlibet/issues/3048#issuecomment-819012316).

`weight_smoothing_factor` can be set to `0` to disable smoothing, to `-1` to disable weighting (negative smooth factors don't make sense anyway) and is `10` by default

----

I've currently changed it for all numeric tags (except `length` obviously), but maybe length-weighted averaging should only apply to some of the [numeric tags](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html?highlight=discnumber#numeric-tags)

Examples

Here two examples showing how ratings have changed

### Example 1
When durations aren't that different, the average rating doesn't change much (which is good)

(from the unweighted version)

example1

```
                                                                          simple
                                      smoothed                             ratio
                                      partial             simple         partial
                          smoothed    weighted            ratio         weighted
rating      minutes   ➜   weight      rating    VERSUS    weight          rating
--------------------------------------------------------------------------------
4★ = 1.00      5.0          8.2%      0.0824                8.0%          0.0802
3★ = 0.75      4.4          7.7%      0.0577                7.1%          0.0533
4★ = 1.00      6.5          9.5%      0.0948               10.5%          0.1049
3★ = 0.75      4.2          7.4%      0.0558                6.7%          0.0503
4★ = 1.00      5.4          8.6%      0.0862                8.7%          0.0872
3★ = 0.75      3.2          6.4%      0.0477                5.2%          0.0392
4★ = 1.00      5.8          8.9%      0.0891                9.3%          0.0928
3★ = 0.75      5.8          8.9%      0.0670                9.3%          0.0698
2★ = 0.50      5.3          8.5%      0.0426                8.5%          0.0425
3★ = 0.75      6.8          9.7%      0.0725               10.9%          0.0819
3★ = 0.75      5.8          8.9%      0.0669                9.3%          0.0696
4★ = 1.00      4.0          7.2%      0.0723                6.4%          0.0641
--------------------------------------------------------------------------------
TOTAL:        62.1                    0.8350                              0.8361

old (unweighted) average rating: 0.8333
old (unweighted) weight (1/len):   8.3%

```

Used Python Console plugin to create the `Track` list for the [script](https://gist.github.com/CreamyCookie/7192358a9f3dbb32cc1ef01784509b3d) with this:

```python
", ".join([f"Track({int(s('~#length'))}, {int(s('~#rating') * 4)})" for s in songs])
```

### Example 2
Sometimes, smoothing safes us from an album being seen as a terrible album (it isn't)

(which it would be with pure duration-weights)

example2


bassstorm:
Bravo!
Verified locally, had an album with old average rating 0.08 due to 80+ silence tracks. Now it's 0.50, which is actually just right. Dream come true, thank you :)

CreamyCookie:
@bassstorm Happy to hear that :smiley:

CreamyCookie:
Here two examples showing how ratings have changed

## Example 1
When durations aren't that different, the average rating doesn't change much (which is good)

(from the unweighted version)

example1

```
                                                                          simple
                                      smoothed                             ratio
                                      partial             simple         partial
                          smoothed    weighted            ratio         weighted
rating      minutes   ➜   weight      rating    VERSUS    weight          rating
--------------------------------------------------------------------------------
5★ = 1.25      2.0         11.8%      0.1474                3.8%          0.0475
5★ = 1.25      4.0         18.5%      0.2312                7.6%          0.0949
0★ = 0.00     20.0         33.9%      0.0000               38.0%          0.0000
0★ = 0.00     26.7         35.8%      0.0000               50.6%          0.0000
--------------------------------------------------------------------------------
TOTAL:        52.7                    0.3787                              0.1424

old (unweighted) average rating: 0.6250
old (unweighted) weight (1/len):  25.0%

                                                                          simple
                                      smoothed                             ratio
                                      partial             simple         partial
                          smoothed    weighted            ratio         weighted
rating      minutes   ➜   weight      rating    VERSUS    weight          rating
--------------------------------------------------------------------------------
4★ = 1.00      5.0          8.2%      0.0824                8.0%          0.0802
3★ = 0.75      4.4          7.7%      0.0577                7.1%          0.0533
4★ = 1.00      6.5          9.5%      0.0948               10.5%          0.1049
3★ = 0.75      4.2          7.4%      0.0558                6.7%          0.0503
4★ = 1.00      5.4          8.6%      0.0862                8.7%          0.0872
3★ = 0.75      3.2          6.4%      0.0477                5.2%          0.0392
4★ = 1.00      5.8          8.9%      0.0891                9.3%          0.0928
3★ = 0.75      5.8          8.9%      0.0670                9.3%          0.0698
2★ = 0.50      5.3          8.5%      0.0426                8.5%          0.0425
3★ = 0.75      6.8          9.7%      0.0725               10.9%          0.0819
3★ = 0.75      5.8          8.9%      0.0669                9.3%          0.0696
4★ = 1.00      4.0          7.2%      0.0723                6.4%          0.0641
--------------------------------------------------------------------------------
TOTAL:        62.1                    0.8350                              0.8361

old (unweighted) average rating: 0.8333
old (unweighted) weight (1/len):   8.3%

```

## Example 2
Sometimes, smoothing safes us from an album being seen as a terrible album (it isn't)

(which it would be with pure duration-weights)

example2

quodlibet/config.py
```diff
@@ -189,6 +189,8 @@
         # (0 = disabled i.e. arithmetic mean)
         "bayesian_rating_factor": "0.0",
 
+        "weight_smoothing_factor": "10",
```
declension, line 192: Some comments like the above might be helpful - in particular what value turns it off

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
 ELPOEP = list(reversed(PEOPLE))
 PEOPLE_SCORE = [100 ** i for i in range(len(PEOPLE))]
 
+SMOOTHING_FACTOR_DEFAULT = 10
```
declension, line 51: This default seems like it's in two places now? `config.py` and here?

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
...
 
-def avg(nums):
+
+def unweighted_average(nums):
```
declension, line 54: Would you mind adding typing annotations to all new (and ideally touched) code? It helps in the long term

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
...
     """Returns the Bayesian average of an iterable of numbers,
     with parameters defaulting to config specific to ~#rating."""
     m = m or config.RATINGS.default
     c = c or config.getfloat("settings", "bayesian_rating_factor", 0.0)
-    ret = float(m * c + sum(nums)) / (c + len(nums))
-    return ret
+    return float(m * c + sum(nums)) / (c + len(nums))
+
+
+def hyperbolic_smoother(x, n=SMOOTHING_FACTOR_DEFAULT):
```
declension, line 67: I'd say (in light of above around two places for default), just make `n` mandatory, then we don't need the default here too

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
...
+    return (n + 1) * x / (n * x + 1)
+
+
+def smoothed_length_weights_and_nums(songs, key, total_length, smoothing_factor):
+    """Returns smoothed length-based weights and the songs' associated numeric values"""
+    length_based_smoothed_weights = []
+    values = []
+    for s in songs:
+        value = s(key)
+        if value == "":
```
declension, line 78:
```suggestion
        if not value:
```
Unless there's some good reason (in which case a comment would be good!)

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
...
+    return float(m * c + sum(nums)) / (c + len(nums))
+
+
+def hyperbolic_smoother(x, n=SMOOTHING_FACTOR_DEFAULT):
+    """n determines how strongly smaller values will be magnified"""
+    return (n + 1) * x / (n * x + 1)
+
+
+def smoothed_length_weights_and_nums(songs, key, total_length, smoothing_factor):
+    """Returns smoothed length-based weights and the songs' associated numeric values"""
```
declension, line 73:
```suggestion
    """:returns: smoothed length-based weights and the songs' associated numeric values"""
```
Though not sure we use RST docstrings much in this codebase actually (probably should)

quodlibet/util/collection.py
```diff
@@ -80,8 +135,15 @@ def bayesian_average(nums, c=None, m=None):
     "max": max,
     "min": min,
     "sum": sum,
-    "avg": avg,
-    "bav": bayesian_average
+    "avg": weighted_average,
+    "bav": weighted_bayesian_average,
+    "unweighted_avg": unweighted_average,
```
declension, line 140: I was thinking `uav` (or `uavg`) - seems more consistent?

quodlibet/util/collection.py
```diff
@@ -80,8 +135,15 @@ def bayesian_average(nums, c=None, m=None):
...
+    "unweighted_bav": unweighted_bayesian_average
```
declension, line 141: ...and `ubav` I guess?

quodlibet/util/collection.py
```diff
@@ -47,19 +48,73 @@
...
+        total_weight += weight
+        weighted_sum += weight * num
+
+    return weighted_sum / total_weight
+
+
+def weighted_bayesian_average(nums, weights, c=None, m=None):
+    """Returns the weighted Bayesian average of an iterable of numbers,
+    with parameters defaulting to config specific to ~#rating."""
+    m = m or config.RATINGS.default
```
declension, line 108: Probably being dim but I don't get why the default rating has an influence here?

quodlibet/util/collection.py
CreamyCookie: Happy to add them for a few functions! :)

I wonder about something. Given that our current style for multi line comments is to start on the same line, is it okay to break with this style when using RST. 

This seems far less readable:

```python
    """:param x: a float in the range [0, 1]
    :param n: determines how strongly smaller values will be magnified
    :returns: a float in the range [0, 1] with smaller values magnified
    """
```

than this:

```python
    """
    :param x: a float in the range [0, 1]
    :param n: determines how strongly smaller values will be magnified
    :returns: a float in the range [0, 1] with smaller values magnified
    """
```

quodlibet/util/collection.py
CreamyCookie: Makes sense!

quodlibet/util/collection.py
CreamyCookie: Happy to add it!

I have to say though, I'm unsure whether the readability decrease is it worth it in simple cases:

```python
def weighted_average(nums: List[float], weights: List[float]) -> List[float]:
```

vs

```python
def weighted_average(nums, weights):
```

In general though, I do like `typing` :+1:

quodlibet/util/collection.py
CreamyCookie: True. I think I missed this, because when I wrote the default constant, I hadn't yet seen / remembered that there's even a default config.

Also the code already there also did something like:

```python
config.getfloat("settings", "bayesian_rating_factor", 0.0)
```

Despite there being the following in the default config:

```json
"bayesian_rating_factor": "0.0"
```

But you're right, so, I think it makes sense to remove them there too (default arguments of `getfloat("settings", "bayesian_rating_factor")`)

quodlibet/config.py
CreamyCookie: Good point!

quodlibet/util/collection.py
CreamyCookie: This check comes from here:

https://github.com/quodlibet/quodlibet/blob/f1bd2739c77d774aa42664aa1897e08456939705/quodlibet/util/collection.py#L298

So, I'll add the comment from there:

```python
 	                # If none of the songs can return a numeric key,
	                # the album returns default
```

quodlibet/util/collection.py
CreamyCookie: It's true that a long suffix is uncommon (all other are 3), and so maybe less consistent, but I feel like in this case it's warranted. If we only have `uavg`, `avg, `ubav` and `bav`, users would have to look up the meaning (or just ignore them). With `unweighted` it's immediately clear that the other method with the same name is weighted (though not by what). Also, the longer names means users are more likely to use `avg`, which in most cases is what they want anyway.

quodlibet/util/collection.py
declension: Cool, and I agree second looks better. None of our linters care enough (currently) about these things so whatever looks good I guess

quodlibet/util/collection.py
declension: I see. `bav` might want to be `bayesian_avg` with this thinking then I guess..

quodlibet/util/collection.py
declension: I agree around the readability a bit, though have got used to it a bit more (and now use >79 chars pretty much everywhere which helps).

It's the ability to find bugs, and to document highly dynamic or confusing code for future people that I especially like here.

Not also those two inputs can be relaxed to `Iterable` here I think (and that's probably just `Number`, not necessarily float right?)

quodlibet/util/collection.py
CreamyCookie: This is copied from the previous unweighted Bayesian average. I think, after reading the Wiki article again, that the default rating is chosen, because it seems like a reasonable prior mean.. 

Simplified, but the following also helped me: https://old.reddit.com/r/explainlikeimfive/comments/bls5hq/eli5_what_is_a_bayesian_average_used_for_and_how/emr2mqj/ :smile: 

> Calculating the Bayesian average uses the prior mean m and a constant C.

@declension Thank you for the review! :smiley:

declension: lol so I'm asking questions about my own code, classic

quodlibet/util/collection.py
CreamyCookie: Actually, I'll try to add another comment that seems clearer

CreamyCookie:
Huh...  mypy is not happy:

> Incompatible types in assignment (expression has type "float", variable has type "int")

but as mentioned in this [stackoverflow post](https://stackoverflow.com/a/50928627/), PEP 484 Type Hints states:

> when an argument is annotated as having type float, an argument of type int is acceptable

quodlibet/util/collection.py
CreamyCookie: I know that feeling.  Sometimes reading code one wrote in the past is really like reading somebody else's code. And in some sense that's true, as we're changing and hopefully improving too.

quodlibet/util/collection.py
CreamyCookie: I'd agree with that, though then we have to go for `unweighted_bayesian_avg` :smile:  which is too long. I get your point though.. It's a bit arbitrary what's abbreviated. I'd actually have gone with `bavg`, but by changing that we'd break existing queries if we switched to that long version.

quodlibet/util/collection.py
CreamyCookie: > Number

Basically, yes. Have chosen `Real` though, which is just `Union[int, float]`, not the real [`Real`](https://docs.python.org/3/library/numbers.html#numbers.Real), as mypy has an open issue about these not working: https://github.com/python/mypy/issues/3186 (still tried but didn't work)

quodlibet/util/collection.py
declension: at the time I was thinking 3-letters is the main thing I guess

quodlibet/ext/events/advanced_preferences.py
```diff
@@ -114,6 +114,15 @@ def changed(entry, name, section="settings"):
                  "the ones present in the song list. Separate with \",\"")),
             text_config("settings", "rating_symbol_full", "Rating symbol (full):"),
             text_config("settings", "rating_symbol_blank", "Rating symbol (blank):"),
+            int_config(
+                "settings", "weight_smoothing_factor",
+                "Weight smoothing factor:",
+                ("This is used to smooth the length-based weights. Without smoothing, "
+                 "very long songs would dominate an average value of an album. "
```
declension, line 121:
```suggestion
                 "long songs dominate average values for an album. "
```

quodlibet/ext/events/advanced_preferences.py
```diff
@@ -114,6 +114,15 @@ def changed(entry, name, section="settings"):
...
+                 "At the same time, short good songs would have no affect on, for "
+                 "example, the average rating of an album. "
+                 "If this is -1 or less, averages are unweighted. "
```
declension, line 124:
```suggestion
                 "Set to -1 for unweighted average. "
```

maybe?

quodlibet/ext/events/advanced_preferences.py
```diff
@@ -114,6 +114,15 @@ def changed(entry, name, section="settings"):
...
+                 "If this is 0, no smoothing is applied. (Default is 10)")),
```
declension, line 125:
```suggestion
                 "Set to 0 for no smoothing (...)")),
```
Though I'm not sure what weighted-but-smoothed means actually?

quodlibet/util/collection.py
declension: Actually remind me why we're adding an `unweighted_` modifier and not just a `weighted_` version (thus leaving current functionality / queries as-is)...

Thus `avg` (remains same) and `bav` (remains same) and `wav` (or `weighted_avg` or whatever)

quodlibet/config.py
```diff
@@ -189,6 +189,15 @@
...
         "bayesian_rating_factor": "0.0",
 
+        # This is used to decrease the high and increase the small weights, which are
+        # used in the calculation of numeric function averages (e.g. ~#rating:avg).
+        # This is done, so that longer tracks have a higher impact on the average than
+        # shorter ones. For example, when calculating the average rating of an album, we
+        # want this to be the case. However, without smoothing, short tracks had nearly
+        # no effect on the average, while long ones would dominate the average.
+        # If this is -1, averages are unweighted. If this is 0, no smoothing is applied.
+        "weight_smoothing_factor": "10",
```
declension, line 199: `weight_bias` just came to mind as well... `smoothing` still makes me think of time series or other 2D data somehow

CreamyCookie: Maybe it's slightly pedantic, but `weight_bias` seems incomplete to me, as it doesn't at all reference the fact that weights are compressed / smoothed. But maybe `weight_smoothing_bias`.. hmm

I hadn't actually done this until now, but when I read the [Wikipedia article](https://en.wikipedia.org/wiki/Smoothing) for smoothing, the term actually does seem somewhat appropriate. 

> In statistics and image processing, to smooth a data set is to create an approximating function that attempts to capture important patterns in the data, while leaving out noise or other fine-scale structures/rapid phenomena. In smoothing, the data points of a signal are modified so individual points higher than the adjacent points (presumably because of noise) are reduced, and points that are lower than the adjacent points are increased leading to a smoother signal.

I guess one difference is that we're going beyond adjacent points.

> smoothing still makes me think of time series or other 2D data somehow

That's technically correct though, as the data is two dimensional (track_index and it's weight, for example: `[0.7, 0.8, 0.2]`).

quodlibet/ext/events/advanced_preferences.py
CreamyCookie: Why the `(...)`?

> I'm not sure what weighted-but-smoothed means actually

The way I see it a `weight_smoothing_factor` of `0` implies that weights aren't smoothed. 

So in an album with two songs, where song 1 is twice as long as song 2 would result in: 

* song 1: has a weight of 2/3 (twice as much as song 2)
* song 2: has a weight of 1/3

(And, of course, then we calculate the rating like this for `:avg`: `s1.rating * s1.weight + s2.rating * s2.weight`)

Or did you mean something else?

quodlibet/util/collection.py
CreamyCookie: Main reason: Provides a reasonable update to ratings (and can easily be disabled) . At the same time, ratings don't change that much.

Also, and I may have misread the thread's sentiment, but it definitely seemed like most people were in favor of making this a default. See, for example, lazka's comment: https://github.com/quodlibet/quodlibet/issues/3048#issuecomment-438558504

https://github.com/quodlibet/quodlibet/blob/bc03a5a6393ccef53b280852051f3ffee5127594/quodlibet/util/collection.py#L189-L194

(minor reason: more consistent with bitrate)

Of course, there was the caveat that you mentioned (about long songs dominating the average), which made me think of the whole smoothing thing :slightly_smiling_face:

quodlibet/ext/events/advanced_preferences.py
declension: Right, I see, thanks :+1: 

Perhaps I'm hoping for a single config number where 
0 : no weighting
1: full (linear) time-weighting (like your example)

And assume nobody would want _more_ than full time-based weighting, that seems crazy.
Is this doable?

quodlibet/config.py
declension: But we're never talking about rating track 1.25, so kinda different perhaps (like you say the adjacency thing). Anyway...

quodlibet/ext/events/advanced_preferences.py
CreamyCookie: Theoretically we could add two options: `use_duration_weights_in_average` and the existing one

---

Okay, GitHub seems buggy. For me, it's not displaying my last response at the bottom to at least one of your comment. I only see it via the ["Changed Files" tab](https://github.com/quodlibet/quodlibet/pull/3626/files), so you might have to check there for my response to the "Actually remind me why we're adding" comment you made.

quodlibet/config.py
CreamyCookie: We could also call it `communism_factor` :laughing:

CreamyCookie:
> Failed to download action 'https://api.github.com/repos/actions/checkout/zipball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.

CreamyCookie:
@declension @lazka Any thoughts about which [numeric tags](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html?highlight=discnumber#numeric-tags) should not use a length weighted average, if any? Anything that needs to be addressed for a merge?

quodlibet/ext/events/advanced_preferences.py
declension: Right found it, thanks.

I think my overall suggestion is why not leave the _explicit_ aggregators (e.g. `avg`) unchanged, so that no user query results suddenly change, but change the _default aggregator_ for given tags.

So implementation-wise, change [`NUM_DEFAULT_FUNCS`](https://github.com/quodlibet/quodlibet/blob/2b7f645bf8120db60817992801f762ff5df340e7/quodlibet/util/collection.py) to point to the new weighted average (whatever it gets called) for whichver tags we choose (possibly all of these. 

This way nobody gets surprised, unless they were using defaults which will be documented to have changed, erm probably, and full choice is still available.

Another way is to ship closer to what you have here, but with the default smoothing to be turned off, to achieve the same "least surprise" outcome.

Either way we should think about ways of describing to users (or a tour of) "what's new" (see #1090 a bit, but much more than this), this would help lots of things

declension:
> failed test is once again `Tcopool.test_pause_resume`, but there's also the following error on the msys2 platform:
> 
> > Failed to download action 'https://api.github.com/repos/actions/checkout/zipball/5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f'. Error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.
> 
> Both seem completely unrelated to this change.

Yep both happen sometimes unfortunately... The download one less often and less our fault (think it's Github actions going wrong). The other one now has #3653.

> 
> @declension @lazka Any thoughts about which [numeric tags](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html?highlight=discnumber#numeric-tags) should not use a length weighted average, if any?

I think it's all about `rating` to me. Also `bitrate` I guess. The others (times, counts) seem too unituitive for users to give a weighting to (by default)

>  Anything that needs to be addressed for a merge?

On seeing it in the flesh I still prefer a three (or four)-letter aggregation e.g. `uav` or `wav` even maybe as consistency feels important here. We should add more prominent docs around this if _that's_ the problem.
 
On which note, could you add some user guide updates I guess.

CreamyCookie:
> I think it's all about rating to me. Also bitrate I guess. The others (times, counts) seem too unituitive for users to give a weighting to (by default)

Makes sense. Might make it an option with rating and bitrate as default.

> On seeing it in the flesh I still prefer a three (or four)-letter aggregation e.g. uav or wav even maybe as consistency feels important here. We should add more prominent docs around this if that's the problem.

I think consistency is important, but I think, as always, there's a balance to be made (in this case between clarity and consistency).

That said, I guess the unweighted variants are pretty long and since the unweighted versions are likely not what users want anyway, maybe clarity isn't that important for them. 

I will change them to `uavg` and `ubav`, as the three letter alternative just seems very problematic to me. (I doubt people would recognize `av` as average.)

> On which note, could you add some user guide updates I guess.

Oh good point! Totally didn't think about that.

TODO (for me):
[ ] unweighted to u
[ ] add option with rating and bitrate as default
[ ] add to user guide

quodlibet/ext/events/advanced_preferences.py
CreamyCookie: @declension I still feel like that has the problem I described in the comment:

> At the same time, hiding good changes in the config means that many, many people will never use them.. For example, I've only now enabled bayesian averaging, despite using QL all these years, as I just didn't know why I should enable it. Other things were more important than figuring out what it is. I'm sure a lot of users are far less curious about why bayesian averaging, smoothing, or duration-weighting makes sense.

(which is also why I created #3634)

Here, if `rating:avg` stays the same, then most people will never benefit from the (small) change, as many people have already created most of their queries, or like me, they tend to use the `avg` even though it can be left out.

Of course, I get your point. There's a balance to be made between improving things for the better and some sort of backward compatibility / not surprising people. 

Hmm.. a poll would be useful in this case. (Is it okay to make the average for ratings length-weighted by default (both `~#rating:avg` and `~#rating`)? Yes, No, Just show results)

bassstorm:
So... what's with this one, gents? Hope to see it in master...

CreamyCookie:
@bassstorm Oh, man time is flying by. Not sure when I get to it, but I'm still planning on doing it eventually.

bassstorm:
No worries, I know what you mean. Was afraid to ask about update :) The result will pay for waiting!

declension:
Any updates on this MR @CreamyCookie ?

bassstorm:
Hi @CreamyCookie, I see you're not active for quite a while... hope you're doing well...

quodlibet/util/collection.py
posita: I don't know much about your use case (of if you even consider this an issue worth pursuing further), but you might want to see if [``numerary``](/posita/numerary) can help you. If [this journey](https://posita.github.io/numerary/latest/whytho/) lines up with your own, ``numerary`` could be an intermediary solution until python/mypy#3186 is fixed. (Alternatively, the techniques used therein might provide inspiration if you can't take a dependency.) Happy to [consult here](https://github.com/posita/numerary/discussions), if helpful. Apologies for the interruption if this is a distraction for you.

bassstorm:
@afontenot
I see you are quite experienced python developer. Maybe this abandoned piece of great but unfinished work also lies in sphere of your interest... Totally okay if it doesn't :)

afontenot:
> @afontenot I see you are quite experienced python developer. Maybe this abandoned piece of great but unfinished work also lies in sphere of your interest... Totally okay if it doesn't :)

I could probably look at this at some point. Could you help by identifying (a) what still needs to be done - the "TODO" in the comment above is hard to follow, and (b) help me understand where this feature is exposed in Quod Libet - is it just when [searching](https://quodlibet.readthedocs.io/en/latest/guide/browse/album.html) or is the album average rating exposed in other places in the UI?

bassstorm:
@afontenot thanks for reply!

a) Perhaps the list is incomplete, but what I see as most important is to fix tests, make the change configurable (if there any gaps) with reasonable defaults, there was a bit of discussion about function naming (however I'm not comfortable to comment that due to lack of language and math background)
b) Besides search queries, it could be used in user-built display templates (album info rendering), but that was indirectly propagated via `avg` if I'm not mistaken, not sure if deviated from that after code review and further changes...

Sorry, not too much of help, I know...

afontenot:
@bassstorm I spent about 20 minutes trying to wrap my head around the idea as it currently stands. I have the following thoughts / questions:

1. It looks like the new feature isn't exposed in the settings, you have to adjust it through the advanced preferences extension. That seems undesirable. Does it need to be fixed?
2. I think I agree with declension that the current parameterization doesn't make much sense. I think I'd probably try to set a logarithmic function such that 0 means that all tracks are treated equally (no averaging, same as the current behavior), 0.5 means the same behavior as 10, the current default, and 1 means a flat time weighted average. This should be pretty easy to do and would allow putting a slider in the setting and adjusting it independently of the current "bayesian" average. 
3. I'm puzzled about the rationale for including a smoothing function at all, actually. I can understand two different attitudes one could have toward an album rating:
 * An album rating represents how much you like playing the *whole* album. In this case, if there's a 2.5 *hour* long track that you *hate* (see example two above), then ... doesn't it just make sense to say that the album is "bad" and you don't want to play it?
 * An album rating represents how much you enjoy individual songs from the album, regardless of whether you would ever play the whole thing. In this case, doesn't the *current* behavior make the most sense, and you should just average the values of the individual songs, without worrying about the length?

Parts of this PR assume that it makes the most sense to try to split the difference between those two ways of thinking about album ratings, but it kinda seems like you end up with some of the flaws of both. Maybe it would make more sense to just let the user toggle between the two behaviors, without trying to do the (much more complicated) smooth interpolation between them?

bassstorm:
@afontenot thanks for that, I know it's not easiest PR and subject in itself.

1. I personally accept either. Not sure about internals, but probably having it just in advanced config is easier to implement. Of course, UI along with `bav` setting would work better.
2. Sounds reasonable, agree.
3. This is very controversial topic, because it touches personal preferences (but that's why we talk about 0,0..1,0 scale). Re your example, yes, but that album would be equally as bad as with _3 mins_ song you hate. Which doesn't seem right if you compare albums. There was quite long discussion in #3048. Here's [the album](https://www.discogs.com/release/113126-m%C2%B2-14id1610s) I mentioned above, it contains 85 short clips of silence (rated by me with zero obviously) that makes album scary bad, which is not true. I would see it among other albums in my queries, and at the same time low rating tracks disappear from my radar if I query tracks by genre for example. Hope that clarifies.

Agree about smoothing. I honestly tried to use bayesian option, but ended up with seeing no reals benefits for me (a hardcore listener who puts a lot into polishing personal library in the sense of tags and ratings). E.g. I don't think preference should be given to shorter tracks albums, that just doesn't make sense in music listening domain (when one deals with own ratings only).

PR #3636, opened 2021-04-19
Moved Burn CD plugin commands into custom commands plugin

### LINKED ISSUE #2165: [Burn CD plugin] configuration option to set burning application to use
it would be nice if there was a configuration option to choose from the supported burning applications (K3b, Brasero or xfburn). Instead of requiring the user to go through the extra menu to select one of those applications each time, the "Burn CD" context menu item would immediately open the user's configured choice.

CreamyCookie:
Check-list
----------

As described in #2165 and with which I agree, some plugins are basically just static versions of commands in the Custom Commands plugin. As such they're incredibly limited. Adding them to the default commands means that users can remove the ones they don't use, or learn how to add one for the program of their choice.

Comment by @declension:

> Agreed a drop-down of one program is not good UX.
> 
> Personally i believe a fair few of these plugins could be converted to Custom Command plugins, and there are a few issues / ideas for consistent UX improvements there.... (though at the loss of some specific features / config)... but i would say that...

It doesn't really make sense to have a plugin that only works with three burning applications (as is currently the case). This pull request removes the burn cd plugin and converts the commands into commands of the custom commands plugin.

I've also added a `reverse` option to the `Command` JSON object, since apparently Xfburn's arguments need to be reversed, and it could be useful for other commands (and especially if a sort argument were to be added).

Closes #2165


CreamyCookie:
Ah.. another `Tcopool.test_pause_resume` failure

declension:
> Ah.. another `Tcopool.test_pause_resume` failure

Worked on re-run

declension:
Thanks!

CreamyCookie:
@declension Nice! I think I tried 3 re-runs (unless I'm mixing things up and thinking of the other PR) :sweat_smile: 

Thanks for merging and review!

PR #3654, opened 2021-04-25
Support library watching in the core

### LINKED ISSUE #961: Support file movement events in auto library plugin
Original [issue 961](https://code.google.com/p/quodlibet/issues/detail?id=961) created by nick.boultbee on 2012-04-28T12:40:48.000Z:

When reshuffling, or even just renaming songs or their directories, playlists will receive the remove signal but will never have the &quot;new&quot; song added -&gt; playlists silently gets smaller.

With the auto-library-update plugin active, it is conceivable that it could intercept the filesystem event (including parent ones like directory renames) and scan all playlists to keep them updated.

It would probably be especially nasty if you moved your library directory but I haven't tried this..

### LINKED ISSUE #1039: Support library watching in core
Original [issue 1039](https://code.google.com/p/quodlibet/issues/detail?id=1039) created by nick.boultbee on 2012-10-11T21:37:16.000Z:

This is a placeholder for something that's been discussed for a while now - the integration of the Auto Library Update code (or at least the concept), with some kind of support or at least fail-proof mechanism for Windows / other OSes without inotify.

This may be a long-term goal, of course, but it would enable or help a few other features, e.g. Issue #961, and it feels deep enough into the library to warrant this.

### LINKED ISSUE #3042: Slow start on huge library caused by pyinotify
I have a library of about 45k tracks. Literally it takes a minute for QL to start (64 seconds from clicking app icon to seeing actual window opened). In the meantime, it's just a matter of 10 seconds for Foobar2000 with same library.
Files are stored on external USB 3.0 HDD. Manjaro Linux, KDE flavor.
I'm not sure but maybe there's a need to move library initialization routine to a separate thread or something.

declension:
* Adds ability to watch all library files natively, and much faster than the plugin
* Removes the plugin that did this
* Removes the dependency (`pyinotify`) too
* Now supports file movement events (#961) - e.g. renames, which is pretty exciting to watch (try it with a browser query on the files you're renaming)
* Also, shutdown libraries explicitly at end of QL (this wasn't happening?...)
* ~~TODO: less manual testing (of which a _lot_ has happened but only on Linux)~~

Check-list
----------
 * [x] Figure out what's happening on Windows and MacOS CI builds
 * [x]  Unit tests have been added where possible

Closes #1039
Closes #961
Fixes #3042


bassstorm:
Wow, impressive!

quodlibet/library/file.py
```diff
@@ -413,3 +418,159 @@ def remove_roots(self, old_roots: Iterable[str]) -> Generator[None, None, None]:
...
+                return
+            handler_id = monitor.connect("changed", self.__file_changed)
+            # Don't destroy references - http://stackoverflow.com/q/4535227
+            self._monitors[path] = (monitor, handler_id)
+            print_d(f"Monitoring {path!r}")
+
+    def __file_changed(self, monitor, main_file: Gio.File,
+                       other_file: Optional[Gio.File], event_type) -> None:
+        try:
+            EventType = Gio.FileMonitorEvent
```
CreamyCookie, line 453: I think this would be better above the class.

quodlibet/library/file.py
```diff
@@ -413,3 +418,159 @@ def remove_roots(self, old_roots: Iterable[str]) -> Generator[None, None, None]:
...
+            file_path = main_file.get_path()
+            if file_path is None:
+                return
+            song = self.get(file_path)
+            file_path = Path(normalize_path(file_path, True))
+            other_path = (Path(normalize_path(other_file.get_path(), True))
```
CreamyCookie, line 459: Considering that there are more than these, I think the `Path(normalize_path(path, True))` part should become a helper function.

quodlibet/library/file.py
```diff
@@ -413,3 +418,159 @@ def remove_roots(self, old_roots: Iterable[str]) -> Generator[None, None, None]:
...
+                          if other_file else None)
+            # print_d(f"Got event {event_type} from {file_path} -> {other_path}")
+            if event_type == EventType.CREATED:
+                if file_path.is_dir():
+                    print_d(f"Monitoring new directory {file_path!r}")
+                    self.monitor_dir(file_path)
+                    # Consume generator
+                    list(self.scan([str(file_path)]))
```
CreamyCookie, line 467: Maybe I'm wrong but I think I've seen this multiple times (a comment "consume generator" or similar with the list creation).

I think it would make sense to create a `exhaust_generator` function as described here: https://stackoverflow.com/a/47456679/ (and supposedly more performant than using `list(x)`)

CreamyCookie (commented):
Apart from what I mentioned and that AFAICS, it's currently enabled by default (and should be configurable, but probably disabled by default): 

LGTM, nice change :+1:

quodlibet/library/file.py
declension: Yep some clean up would be good (after this is done, it goes way further than just here)

quodlibet/library/file.py
declension: Fair point, I've switched it out now

PR #3656, opened 2021-04-26
Added ObjectListMatcher utility class

CreamyCookie:
Check-list
----------

This is needed for a plugin I'm working on, which is described in #3658. So, instead of creating a very large pull request, I thought it's probably better to split it up a bit. 

Basically this will be used when importing for matching exported metadata to the selected albums.

I can also imagine this class being useful in other situations like, for example, finding the best fitting `lrc` file for a track, or matching downloaded metadata to songs.

It's adding a new utility class, `ObjectListMatcher`, which can be used to match the elements of two lists. It does so by comparing how similar user-selected attributes of the elements are, and returns the indices of the second list in a order that best matches elements in the first list.

**Construction**
* Elements in both lists have to have the same type. Apart from that, any type is okay.
* When creating a new instance with the constructor (and not a factory method), one has to supply a dictionary that maps a function that returns a certain attribute to the weight of that attribute.
* There are factory methods for comparing objects themselves (`of_identity`), a single attribute or sequences (`for_sequence`).

**Values returned by attribute getters**
  * Numbers (`int` or `float`) will be compared by `1 - abs_delta / abs_max_delta`
  * If an attribute is neither `Sequence` nor number, it will be converted to a `str` and then:
  * Whenever a `Sequence` attribute has to be compared, [`difflib.SequenceMatcher`](https://docs.python.org/3/library/difflib.html#difflib.SequenceMatcher) is used.

**Getting the matches**
* The `get_indices` method returns the indices of the second argument in the order that they best match elements in the first argument.
* If enabled, the similarity matrix can be retrieved after that call.

### DIFF
--- a/quodlibet/util/matcher.py
+++ b/quodlibet/util/matcher.py
@@ -1,0 +1,382 @@
+# Copyright 2021 Joschua Gandert
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+from difflib import SequenceMatcher
+from operator import itemgetter
+from typing import Generic, TypeVar, Mapping, Callable, Any, Union, List, Sequence, \
+    Optional, Tuple
+
+
+T = TypeVar('T')
+Real = Union[int, float]
+AttributeGetter = Callable[[T], Any]
+AttributeGetterToWeight = Mapping[AttributeGetter, Real]
+
+SUPPORTED_NUMBER_TYPES = (int, float)
+
+
+class _MatchData:
+    def __init__(self, a_idx, a_value, b_size):
+        self.a_idx = a_idx
+        self.a_value = a_value
+        self.b_idx_to_similarity = [0 for _ in range(b_size)]
+
+        self.best_b_idx = -1
+        self.best_b_similarity = float('-inf')
+        self.second_best_b_idx = -1
+        self.second_best_b_similarity = float('-inf')
+
+        self.continue_attr_index = 0
+
+        # rarely needed since in a lot of cases there won't be conflicts
+        self._sorted_b_similarity_with_idx_pairs = None
+        self.is_fully_measured = False
+
+    def add_similarity(self, b_idx, similarity_part):
+        self.b_idx_to_similarity[b_idx] += similarity_part
+        b_idx_total_similarity = self.b_idx_to_similarity[b_idx]
+
+        if b_idx_total_similarity > self.best_b_similarity:
+            self.second_best_b_idx = self.best_b_idx
+            self.second_best_b_similarity = self.best_b_similarity
+
+            self.best_b_idx = b_idx
+            self.best_b_similarity = b_idx_total_similarity
+        elif b_idx_total_similarity > self.second_best_b_similarity:
+            self.second_best_b_idx = b_idx
+            self.second_best_b_similarity = b_idx_total_similarity
+
+    def set_to_fully_measured(self):
+        self.is_fully_measured = True
+
+        # We sort so last = best as calling pop() on a list is O(1)
+        # We negate the index here, so that the sorting is correct if the
+        # similarities of two b's are the same (lowest index last)
+        self._sorted_b_similarity_with_idx_pairs = sorted(
+            ((s, -i) for i, s in enumerate(self.b_idx_to_similarity)))
+
+    def replace_best(self):
+        if not self._sorted_b_similarity_with_idx_pairs:
+            return
+
+        self._sorted_b_similarity_with_idx_pairs.pop()
+        if not self._sorted_b_similarity_with_idx_pairs:
+            # None left, which means all others a's have better values
+            # than this one. As a result, this a will stay alone :(
+            self.best_b_idx = -1
+            self.best_b_similarity = float('-inf')
+            return
+
+        sim, idx = self._sorted_b_similarity_with_idx_pairs[-1]
+        self.best_b_idx = -idx
+        self.best_b_similarity = sim
+
+
+class ObjectListMatcher(Generic[T]):
+    """
+    Utility class that compares the objects of two lists (all the same type), and finds
+    the best matches. What is considered a good match depends on the supplied mapping,
+    which maps "attribute getter" functions to their weight.
+
+    Weights are normalized, so you can use values > 1.
+
+    A specific getter has to return objects of only one of these types:
+    - Real number (e.g. int or float) - compared by 1 - abs(delta) / max_delta
+    - Other non-Sequence - converted to str which is a Sequence and then..
+    - Sequence - compared using difflib.SequenceMatcher
+
+    The problem this solves is called the "Assignment problem".
+
+    Note that this class will not necessarily call every attribute function and thus
+    not use its weight in the calculation, as it will avoid doing unnecessary
+    calculations when the current best match for an element is undefeatable.
+
+    For example, let's say we have an attribute x with weight 0.8 and an attribute y
+    with weight 0.2. If, after checking the similarity on attribute x with all items,
+    the highest similarity is 0.8 (it was equal to the compared item) and the second
+    highest has a similarity of 0.5, then we won't check attribute y at all, as the
+    second best cannot possibly win with the remaining weight (0.5 + 0.2 < 0.8).
+
+    However, whenever there's a conflict (two elements match to one), the full
+    similarity scores of the affected elements will always be calculated."""
+
+    similarity_matrix: List[List[float]]
+    minimum_similarity_ratio: float
+
+    _matcher: SequenceMatcher
+    _attr_with_weight: List[Tuple[Callable[[T], Any], float]]
+    _b_idx_to_a_match_data: List[Optional[_MatchData]]
+
+    def __init__(self, attr_to_weight: AttributeGetterToWeight):
+        self.update_attr_to_weights(attr_to_weight)
+        self._matcher = SequenceMatcher()
+
+        self.should_store_similarity_matrix = False
+        """Whether or not to store the similarity_matrix.
+
+        You may want to enable should_go_through_every_attribute to make similarity
+        scores between different a_items more comparable. Without it, the algorithm
+        may have stopped before evaluating all attributes for one a_item."""
+
+        self.similarity_matrix = []
+        """This will be recreated whenever get_indices is called, given that if
+        should_store_similarity_matrix is True.
+
+        It's a matrix that has the same number of elements as a_items. Each of those
+        lists has the same number of elements as b_items. Those elements represent how
+        similar an a item is to each b item. So matrix[a_idx][b_idx] gives you the
+        similarity between the a item and b item at those indices. A similarity of 1
+        means that every attribute was evaluated and equal to the other."""
+
+        self.minimum_similarity_ratio = 0.0
+        """A value in the range [0, 1] that defines what similarity is required to be a
+        valid match. Setting this to a reasonable value will speed up the calculation
+        as low-quality matches will be discarded before a match conflict can even arise.
+
+        Note that this isn't using the absolute similarity for comparison, because that
+        would lead to confusing results, given that by default this class only evaluates
+        weights and attributes if there is no decisive winner (best match) yet. Instead,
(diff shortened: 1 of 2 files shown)


CreamyCookie:
Hmm.. seems like requesting reviews from multiple people causes a sort of bystander effect (diffusion of responsibility). Interesting!

bassstorm:
Actually, I tried to start, but that's a bit too complex for me. Also I don't quite understand the problem statement, sorry :)

CreamyCookie:
@bassstorm Oh, no problem! Thanks for commenting here and starting a review :smiley: 

> bit too complex for me

It is definitely somewhat complex due to all the mathematics and optimizations. 

>  the problem statement

Yeah, I could have done better with the pull request description. And some things I added a bit too late. The use of this class is to match lists of arbitrary objects (but the lists in both objects should have the same type, or at least return a value for the attribute getter).

In the import / export plugin, there's the problem of having songs and matching them to some exported tags (and other data). For example, let's see you have an album with 5 songs of low bitrate in your library, but you've recently bought the 11 track Deluxe version of that album. So, now you add that to the library, and want to get rid of the old version. However you would like to copy over ratings and play counts. 

Song titles may differ slightly, contain other capitalization or contain something like ` (Remastered)` in their `title` tag. Furthermore, the order might have changed slightly. (I have definitely seen that.) That's where the matcher is helpful. Despite the differences it can easily match the existing metadata to the new tracks.

CreamyCookie:
@declension Any idea how changing metadata like ratings can result in every song appearing twice in an album? 

Weirdly enough the duplicates disappear after restarting QL. I'm calling `background_check_wrapper_changed` on the songs, so I have really no idea what could be causing this..

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # If this hasn't gone through every attribute and weight yet, we normalize the
+        # similarity, so it's in the range of [0, 1]. Without this, minimum_similarity
+        # would be somewhat unpredictable.
+        #
+        # For example, a b_item might be undefeatable with a similarity of 0.52. If we
+        # wouldn't normalize it, it would not pass a minimum of 0.6, even if it could
+        # easily achieve that with the remaining weights.
+        if cont_attr_idx < len(self._attr_with_weight):
+            cur_total_weight = 1 - self._weight_left[cont_attr_idx]
```
declension, line 245: How do we know `self._weight_left[cont_attr_idx]` isn't 1?

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+
+T = TypeVar('T')
+Real = Union[int, float]
+AttributeGetter = Callable[[T], Any]
+AttributeGetterToWeight = Mapping[AttributeGetter, Real]
+
+SUPPORTED_NUMBER_TYPES = (int, float)
+
+
+class _MatchData:
```
declension, line 22: No type annotations here so hard to say, but is this not generic on the  values that it matches on?

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+        self.second_best_b_idx = None
+        self.second_best_b_similarity = float('-inf')
+
+        self.continue_attr_index = 0
+
+        # rarely needed since in a lot of cases there won't be conflicts
+        self._sorted_b_similarity_with_idx_pairs = None
+        self.is_fully_measured = False
+
+    def add_similarity(self, b_idx, similarity_part):
```
declension, line 39: What's this for (no docstring here)?

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+    def set_to_fully_measured(self):
+        self.is_fully_measured = True
+
+        # We sort so last = best as calling pop() on a list is O(1)
+        # We negate the index here, so that the sorting is correct if the
+        # similarities of two b's are the same (lowest index last)
+        self._sorted_b_similarity_with_idx_pairs = sorted(
+            ((s, -i) for i, s in enumerate(self.b_idx_to_similarity)))
+
+    def replace_best(self):
```
declension, line 62: Hmm, replace with what? Definitely not following any of this but maybe that's fine

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+                # a1 has no matches left or could not find a match
+                return
+
+            a2_match_data = self._b_idx_to_a_match_data[best_b_idx]
+            if a2_match_data is None:
+                self._b_idx_to_a_match_data[best_b_idx] = a1_match_data
+                return
+
+            # We have found a conflict and will now solve it.
+            a1_match_data = self._get_worse_match_data(a1_match_data, a2_match_data)
```
declension, line 263: How does this escape the `while True`?

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def _get_worse_match_data(self, a1_match_data, a2_match_data):
```
declension, line 265: Why... worse... than what? :confused:

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+            # a1 is better so replace a2 in map
+            best_b_idx = a1_match_data.best_b_idx
+            self._b_idx_to_a_match_data[best_b_idx] = a1_match_data
+
+            a2_match_data.replace_best()
+
+            # we need to find a new match for a2
+            return a2_match_data
+
+    def _finish_similarity_measures(self, a_match_data):
```
declension, line 288: What happens if you don't call finish? Can it be designed so that's impossible?

quodlibet/util/matcher.py
```diff
@@ -0,0 +1,382 @@
...
+        # figure out total similarity (without stopping) if we didn't
+        self._measure_similarity_to_find_best_b_match(a_match_data)
+        a_match_data.set_to_fully_measured()
+
+    def _measure_similarity_to_find_best_b_match(self, a_match_data):
+        continue_attr_idx = a_match_data.continue_attr_index
+        can_stop = continue_attr_idx == 0 and not self.should_go_through_every_attribute
+        attr_size = len(self._attr_with_weight)
+
+        for attr_idx in range(continue_attr_idx, attr_size):
```
declension, line 303: why not just iterate over `self._attr_with_weight` then (or use `enumerate`)?

declension:
> Hmm.. from the few times I tried, it seems like requesting reviews from multiple people causes a sort of bystander effect (diffusion of responsibility). Interesting!
> 
> Or maybe it's the feature itself that makes a request very impersonal / cold.

Sorry, I've had a look now, but yes as @bassstorm said too, I have no idea what's going on or why really, so not much I can add here.

Overall two thing that might (have) help(ed) is more docstrings / type annotations on functions and less mixing of state management and calculation - i.e. aiming for as few `foo() -> None` methods as possible, as they're harder to test / reason with.

Sorry I couldn't be more help

quodlibet/util/matcher.py
CreamyCookie: Thank you for reviewing! It is generic and I'll add the type hint.

quodlibet/util/matcher.py
CreamyCookie: I've renamed it to `replace_best_with_next_best`.

Basically we have the following situation: Two lists, a_items and b_items. An a_item is already matched to a b_item (let's say `Song 1` with song 1 (remastered)`). 

Now we measure the similarity of the next a_item and find that it's best match is the exact same b_item. We have a conflict, and we need to find out which a_item is the better match for the b_item. So we compare their similarity (scores) and for the one that loses, we replace it with their second best match.

Then we return to the `while True` loop, but now we have to check if that second best match might also be in conflict... and this repeats until there are no conflicts.

quodlibet/util/matcher.py
CreamyCookie: The `_measure_similarity_to_find_best_b_match` always goes through at least one attribute, so `cont_attr_idx` can never be 0 and per definition this can never be 1:

https://github.com/quodlibet/quodlibet/blob/b838ebcae22b0b2c63944f7663bf8e07063849cd/quodlibet/util/matcher.py#L184-L188

(before that section we have checked that no weight is 0)

quodlibet/util/matcher.py
CreamyCookie: Hmm.. not really sure how to explain without basically going through every line here. 

In short, at some point there will always be an `a` without a match (`best_b_idx is None`) or a `best_b_idx` will not have match assigned yet. And of course, if there is a conflict, one a will by definition (see `_get_worse_match_data`) lose their match, so this loop cannot run forever, since inside this loop matches are only removed (`replace_best`), or they lead to a `return`.

quodlibet/util/matcher.py
CreamyCookie: I added a docstring and changed the name to `_get_losing_match_data`. (returns the match data that lost the conflict)

Maybe that's more clear?

quodlibet/util/matcher.py
CreamyCookie: Other than inlining the code inside `_finish_similarity_measures` (and thus duplicating code).. Can't think of any. You?

quodlibet/util/matcher.py
CreamyCookie: The `continue_attr_idx` is not always `0` and we don't need the weight here.

quodlibet/util/matcher.py
CreamyCookie: Added doc string

CreamyCookie:
> i.e. aiming for as few def foo(self,...) -> None methods as possible, as they're harder to test / reason with.

But that would lead to duplication and very long functions, no? Would need an example inside the matcher code to improve something in that area.

> Sorry I couldn't be more help

No, your comments were great. Thank you!

tests/test_util_matcher.py
```diff
@@ -0,0 +1,259 @@
...
+        assert matcher.get_indices(a, b) == [2, None, 1, None], formatted_matrix(
+            matcher)
+
+        matcher.minimum_similarity_ratio = 0.9
+        assert matcher.get_indices(a, b) == [None, None, 1, None], formatted_matrix(
+            matcher)
+
+
+class Car:
+    def __init__(self, seats: int, name: str, features: List[str]):
```
declension, line 256: `@dataclass` is good for this stuff BTW

quodlibet/util/matcher.py
declension: OK cool - maybe add a comment here to say we _know_ it's not zero, for the next reader :+1:

declension (approved):
> @declension Any idea how changing ratings, added and skipcount can result in every song appearing twice in an album? (Okay, seems to only happen after selecting another and then reselecting the album)
> 
> Weirdly enough the duplicates disappear after restarting QL. I'm calling `background_check_wrapper_changed` on the songs, so I have really no idea what could be causing this.. (also happened when I was just calling `library.changed` on the songs)

No idea... sounds like something's getting added to the library rather than updated, not sure why though.


Anyway some small comments but basically LGTM (the bits that I understand)

tests/test_util_matcher.py
CreamyCookie: Oh.. somehow I thought they were added in 3.8. Will replace.

CreamyCookie:
@declension Thanks again! I will squash and merge this now :slightly_smiling_face: 


Just a note: I've added a minor optimization that makes the creation of `_sorted_b_similarity_with_idx_pairs` lazy, which will improve memory usage and performance, as often there's a good match, and the second best match has no chance. Previously lists of both `a_item`'s would have been created. Now this will only happen whenever `replace_best_with_next_best` is called (which is the only place the list is touched). 

I've also repeated the whole "compare a thousand lists of random objects" thing to check for any errors to be sure.

```python

from random import choice, randint, sample
import string

om = ObjectListMatcher.for_sequence([3, 8, 1])
letters = string.ascii_letters  + string.digits + string.punctuation


word = lambda : "".join(sample(letter, randint(1, 9)))
obj = lambda : (word(), word(), randint(-323, 333))
lst = lambda : [obj() for _ in range(randint(1, 9))]

def run_tests():
    for _ in range(1000):
        om.get_indices(lst(), lst())

start_time = dt.now()
run_tests()
print(dt.now() - start_time)

PR #3666, opened 2021-05-04
Updated Dutch translation

Vistaus:
🙂


lazka:
Thanks.

Random translation related question: If we'd switch to an online translation tool, which one would you prefer?

Vistaus:
NP :)

I very much prefer Transifex.

lazka:
Thanks!

lazka:
we've had some problems with transifex, so we now switched to weblate instead: https://twitter.com/QuodLibetApp/status/1392536335785791488

PR #3705, opened 2021-05-13
Added a unified and flexible ImportExport plugin

### LINKED ISSUE #1510: Ratings import plugin
Original [issue 1510](https://code.google.com/p/quodlibet/issues/detail?id=1510) created by reiter.christoph on 2014-12-26T18:53:09.000Z:

short braindump:
- Add a plugin to import ratings
  - Select a range of songs
  - It loads all the ratings of those songs and shows a list of usernames/emails and how many of the selected songs have a rating for each user.
  - The list allows selecting users to import ratings for them (checkboxes)
  - The list allows reordering users and selecting multiple, so that ratings from multiple sources can be merged (use rating from FOO, if not available from BAR, else set no rating..)
  - The imported ratings get saved according to the username/email set in the prefs; as is currently the case.
- Maybe add a second plugin to save ratings. This would just be the force write plugin with a different name. Maybe a good idea because the forcewrite plugin isn't obvious and we can add both plugins in a submenu to make this discoverable.

As always feedback welcome.

### LINKED ISSUE #3658: Add new plugin that combines the "Migrate Metadata" and the "Import / Export" plugins
The proposed change
-------------------
The "Migrate Metadata" and the "Import / Export" plugins do very similar and related things. They both allow you to transfer data from one set of songs to another.

Actually, they're three plugins as the plugin file (`importexport.py`) contains an Import and an Export plugin in the same file.

When the user enables the new plugin they get a submenu that contains one Import and multiple Export options.

Benefits of this change
-----------------------
* Cleaner plugin window (list of plugins)
* Instead of having to enable two (or three) plugins, there's one for both import and export
* Users will be able to export ratings and more (Closes #1439)

Any issues this might cause
---------------------------
The new plugin uses `json`, so if users have been doing something with the old exported `tags` file other than importing, they might have to add the old plugin to their plugin folder.

Tags and user track data are always exported to a cache directory, so if they need the exported files, they'll have to go there. I'm planning on adding a button that opens that directory in the file manager.

CreamyCookie:
Check-list
----------

 * [ ] Unit tests have been added where possible - Tested manually with 25 albums though

Closes #1510 
Closes #3658

Adds a new plugin that combines the functionality of the following plugins: Import, Export and Migrate Metadata (all deleted in this PR) as explained in #3658. 

Furthermore, it allows users to export and import user data like ratings, play count and in which playlists the selected tracks were stored. This makes it easy to replace albums with their lossless, extended or remastered version while keeping all the metadata..

It might also help users that use QL across systems (e.g. @BraveSentry, @bassstorm, see #3426)

**New version of album is extended** 
Screenshot_20210501_233515

(Whether this matching window is shown depends on the user's plugin config.)

Thanks to the `ObjectListMatcher`, it's possible to provide users with a reasonable default matching when that replacement version has a widely different order or somewhat different track names:

**New version has different order** 
Screenshot_20210501_234124

**Plugin Config User Interface**
Screenshot

Users can export various combinations of file names, tags and track user data:

**Context menu**
Untitled

**Export file example** ("Export User Data" chosen)
```json
[ 
    {
     "//identifier": ["Paul Kalkbrenner", "Battery Park", 1, 1, 1, 12,
                      "01 - Battery Park.flac"], 
     "//playlists": ["focus"],
     "~#playcount": 7, "~#laststarted": 1618696312, "~#skipcount": 1, 
     "~#rating": 1.0, "~#lastplayed": 1618696611, "~#added": 1610564577
    }, 
    {
     "//identifier": ["Paul Kalkbrenner", "Cylence 412", 1, 1, 2, 12,
                      "02 - Cylence 412.flac"], 
     "//playlists": ["focus"],
     "~#playcount": 3, "~#laststarted": 1618957186, "~#skipcount": 5, 
     "~#rating": 0.75, "~#lastplayed": 1618957452, "~#added": 1610564577
    }, 
    {
     "//identifier": ["Paul Kalkbrenner", "Cloud Rider", 1, 1, 3, 12,
                      "03 - Cloud Rider.flac"], 
     "//playlists": ["focus_vocals"],
     "~#playcount": 7, "~#laststarted": 1618956795, "~#skipcount": 5,
     "~#rating": 1.0, "~#lastplayed": 1618957186, "~#added": 1610564577
    }
]
```


bassstorm:
Sounds pretty cool! Will try to have a look tomorrow from at least high level (user) perspective, thanks.

bassstorm:
Massive! Gonna try it eventually for sure. Thanks @CreamyCookie!

CreamyCookie:
@bassstorm Nice. I'm looking forward to your feedback! :smiley:

BraveSentry:
I finally got around to testing this, at least between macOS and Mint, and it works like a charm. Thank you so so much for this!

I had one issue though, concerning two tracks from one specific album: Every time I tried using the plugin on one of these tracks, QL stopped responding. Running with --debug and then trying to use the plugin on that track gives me `D: 60.095: ImportExportTagsAndTrackUserDataPlugin.extract_data_and_export: None` - which is the same as what I get when using the plugin on any single file I tried. Yet I get no freezing QL on other files. Any ideas what might be wrong with this file? Or any other parameters I could run QL with, other than just --debug ?

CreamyCookie:
@BraveSentry You are welcome :)

Thanks for the feedback.

> concerning two tracks from one specific album: Every time I tried using the plugin on one of these tracks, QL stopped responding

If you want you can upload the problematic files to the private repository to which I invited you. Alternatively it would be helpful what file type it is. Maybe you could run these two commands on the files:

`operon info FILENAME_HERE`

and

`operon list FILENAME_HERE` 

Could you maybe upload these

Offerel:
Currently im using 4.4.0 of QuodLibet. The plugin is not included as of yet in the release. Is there any timeline, when this will go live?

PR #3723, opened 2021-06-11
Propagate repeat options properly in mpris2 plugin

luk1337:
Check-list
----------

This change fixes the way LoopStatus is handled, previously it wouldn't work properly if one were to set it to "Playlist".


declension:
Thanks

PR #3749, opened 2021-07-18
Set empty input shape for hints window (Fixes #1691)

### LINKED ISSUE #1691: Scrolling problems in Album List mode
Hi,

The problem is that if the mouse is moving while scrolling through the albums in Album List mode, then the scrolling stops. This gets pretty annoying since you can't scroll without moving the mouse a millimeter. As long as the mouse continues to move then absolutely no scrolling will be done.

Sometimes you need to move the mouse a bit in case scrolling has locked up or press the middle mouse key and try moving again. Overall scrolling is pretty hopelessly broken in Album List mode.

This happens only in Album List mode. Is definitely not a regression, just a bug I encountered since last December but never actually reported and just lived with it. I always compile and install quodlibet from source.
Also bug hasn't been fixed in GTK as every time Debian has updated it it's still present. I can't find if the bug is in GTK or not since I've been unable to find some similar bug elsewhere.

I can't verify if the bug happens in Windows or OSX, I can only confirm it's present under Debian unstable.

slosd:
This change makes gdk pass all events on the gtk window for "hints" to the window below (which is some quodlibet window in which the tooltip was triggered). 

Fixes #1691 on x11, wayland and win32. On macOS this doesn't do anything, so the current code which passes the events to the other window remains.

Check-list
----------

### DIFF
--- a/quodlibet/qltk/views.py
+++ b/quodlibet/qltk/views.py
@@ -39,6 +39,8 @@ class TreeViewHints(Gtk.Window):
          'enter-notify-event', 'leave-notify-event'],
         'override')
 
+    __empty_region = cairo.Region(cairo.RectangleInt())
+
     def __init__(self):
         try:
             # gtk+ 3.20
@@ -350,6 +352,11 @@ class TreeViewHints(Gtk.Window):
         self.set_transient_for(get_top_parent(view))
         set_text(label)
         self.set_size_request(w, h)
+
+        # Set region on this window for which to receive mouse events to the
+        # empty region. Mouse events will be passed to the window below the
+        # tooltip.
+        self.input_shape_combine_region(self.__empty_region)
 
         window = self.get_window()
         if self.get_visible() and window:


declension:
Thanks!
Would you mind adding a comment in the code to refer to why we're doing this, I can definitely imagine coming back to this and having no idea of why it's there...

slosd:
@declension You're absolutely right, the method name doesn't tell you very much.
Is this ok? I guess the comment now assumes a common understanding that tooltips don't want to handle any events themselves.

lazka:
After this, clicking on tooltips no longer selects the row.

slosd:
@lazka I cannot reproduce that. Any particular tooltip? I can click through tooltips and select rows in the album list browser as well as the song list.

lazka:
Every tooltip, but this is on Windows.

slosd:
I can try to debug this on Windows. Seems like the event is simply discarded and not handled in either window.

PR #3751, opened 2021-07-18
Do not start dragging from empty treeview area (#3538)

### LINKED ISSUE #3538: Duplicate playlists on DnD of empty area
Steps to reproduce
------------------
Dragging and dropping the empty area at the bottom of the list of playlists onto itself.

Expected Output
---------------
Nothing.

Actual Output
-------------
A new playlist with all the tracks of the currently selected (but not dragged) playlist is created.

Test System
-----------

b1429c65b1d91789daffd9eef5274b99b4789660

### Which operating system 
Linux with a twm-derivative window manager (focus follows the mouse cursor). Probably not relevant.

slosd:
Fixes #3538 

What's notable is that if the playlists browser used `AllTreeView` instead of `RCMHintedTreeView` the `MultiDragTreeView`, which is mixed in with it, would deal with that by checking if there is a row at the current mouse pointer position and not start dragging if there isn't.

```python
class RCMHintedTreeView(HintedTreeView, RCMTreeView, DragIconTreeView):
    """A TreeView that has hints and a context menu."""
    pass


class AllTreeView(HintedTreeView, RCMTreeView, DragIconTreeView,
                  MultiDragTreeView):
    """A TreeView that has hints, a context menu, and multi-selection
    dragging support."""
    pass
```
(from [quodlibet/qltk/views.py](https://github.com/quodlibet/quodlibet/blob/80977572a307d81016a0f7bcc55049284f609d0a/quodlibet/qltk/views.py#L1313))

Check-list
----------

### DIFF
--- a/quodlibet/browsers/playlists/main.py
+++ b/quodlibet/browsers/playlists/main.py
@@ -258,8 +258,8 @@ class PlaylistsBrowser(Browser, DisplayPatternMixin):
         ]
         targets = [Gtk.TargetEntry.new(*t) for t in targets]
         view.drag_dest_set(Gtk.DestDefaults.ALL, targets, Gdk.DragAction.COPY)
-        view.drag_source_set(Gdk.ModifierType.BUTTON1_MASK, targets[:2],
-                             Gdk.DragAction.COPY)
+        view.enable_model_drag_source(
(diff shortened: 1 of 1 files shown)


declension:
Thanks!

PR #3759, opened 2021-07-24
Update macos devenv setup documentation

mjpieters:
This covers:

- a small fix to the devenv documentation; these were moved a year ago in
227c1e7f9fcf62db724ad42dde83ecbdf63abed1.
- my own experience getting my macos dev env configured to the point where the (majority) of the test suite passes.

### DIFF
--- a/docs/development/devenv.rst
+++ b/docs/development/devenv.rst
@@ -44,6 +44,41 @@ the right environment set up.
     $ git clone https://github.com/quodlibet/quodlibet.git
     $ ./QuodLibet.app/Contents/MacOS/run <path_to_git_repo>/quodlibet/quodlibet.py
 
+On recent MacOS releases, the OS Gatekeeper will complain about the application not being recognised.
+It is easiest to just clear the `com.apple.quarantine` extended attribute from all files in the bundle
+rather than try and open each and every component that MacOS will refuse to open:
+
+::
+
+    $ xattr -rd com.apple.quarantine ./QuodLibet.app/Contents
+
+The bundle includes `pip`, so you can always install additional packages (such as `flake8`, `pytest` and
+`flaky`, which would let you run the test suite):
+
+::
+
+    $ ./QuodLibet.app/Contents/MacOS/run -m pip install flake8 pytest flaky
+    $ ./QuodLibet.app/Contents/MacOS/run <path_to_git_repo>/setup.py test
+
+If you want to run the tests with your own Python command, you'll need to install some additonal software
+and packages:
+
+::
+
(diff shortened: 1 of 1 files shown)


mjpieters:
As a side note: I was glad to have found the [quodlibet metamodule list](https://github.com/quodlibet/quodlibet/blob/3ef6292b093b848e820b6be47b5a70deb6d4c780/dev-utils/osx_bundle/modulesets/quodlibet.modules#L85-L106), to help fill in the missing gaps to get a dev environment set up where `./setup.py test` would work. A combination of `brew search`, `brew install` and `pip install pyobjc` got me there.

So, anyone wanting to contribute using a mac and wanting to run the test suite must first run:

```
brew install cairo dbus gst-libav gst-plugins-bad gst-plugins-good gst-plugins-ugly \
    gstreamer gtk-mac-integration gtk+3 libsoup pkg-config pygobject3
poetry run pip install pyobjc
```

declension:
> 
> So, anyone wanting to contribute using a mac and wanting to run the test suite must first run:
> 
> ```
> brew install cairo dbus gst-libav gst-plugins-bad gst-plugins-good gst-plugins-ugly \
>     gstreamer gtk-mac-integration gtk+3 libsoup pkg-config pygobject3
> poetry run pip install pyobjc
> ```

This is great - do you fancy adding it to the developer guide somewhere?

mjpieters:
Problem is: not all tests pass. I think it’s an issue with gstreamer plugins, but I have not the time to figure out where this is going south. I also have the test suite running in a docker container (against Ubuntu bionic), and I have enough to test the Remote / FIFO implementation.

but, I’m guessing that even a partial step is still a step forward, right? :-)

mjpieters:
> This is great - do you fancy adding it to the developer guide somewhere?

Right, I've updated the PR with additional devsetup info for macos.

With that setup, on Big Sur with Python 3.8.11 (to avoid running into [Python bug #41100](https://bugs.python.org/issue41100), only 3 test failures remain:
- `test_player_gst::TGStreamerCodecs::test_decode_all` (no wavpack or gme plugins in the brew gst plugin packages)
- `test_util::Tload_library::test_glib` (because `quodlibet.utils.load_library()` only wants to load libraries ending with `.dylib` from `sys.prefix/lib/...`, which is the poetry virtualenv, and my `glib` library is homebrew-installed)
- `test_mypy::TestMypy::test_project` (spurious, perhaps a difference in mypy versions, complaining about `quodlibet/formats/_audio.py:162` being `Optional[str]` instead of `str`).

declension:
Thanks!

Issue #3745, opened 2021-07-16
utils/fifo:FIFO._process can call _open() without required ignore_lock argument.


mjpieters:
I'm not 100% yet on how this can be reproduced, just that I've seen this happen locally when using an external process that writes to the named pipe to control QuodLibet. I've not immediately been able to recreate the circumstances, but it appears the pipe was opened and immediately closed again. I've attached the dump log at the bottom of this issue.

The traceback is helpfully short:

```
Traceback (most recent call last):

  File "GLib.py", line 669, in <lambda>

  File "/Applications/QuodLibet.app/Contents/Resources/lib/python3.7/site-packages/quodlibet/util/fifo.py", line 284, in _process
    self._open(*args)

TypeError: _open() missing 1 required positional argument: 'ignore_lock'
```

The fix is simple, just add `True` to the call:

```python
self._open(True, *args)
```

However, there is, from reading the code, another location in the same method that needs fixing; there is no `self.__open()` function, but if `data = source.read()` fails with an `OSError` that is not `EWOULDBLOCK`, `EAGAIN` or `EINTR`, that's what `FIFO._process()` will try and call.

Personally, I'd replace the `ignore_lock` positional argument with a keyword argument (`ignore_lock=False`) and use explicit `_open(*args, ignore_lock=True)` calls, but the immediate fix would be to adjust those two `_open()` calls to use the correct name and argument count.

See https://github.com/quodlibet/quodlibet/blob/66acde05600d7efa1ca214102f59af856b6cf42c/quodlibet/util/fifo.py#L265-L289, the fixed version would be:

```python
    def _process(self, source, condition, *args):
        if condition in (GLib.IO_ERR, GLib.IO_HUP):
            self._open(True, *args)
            return False

        while True:
            try:
                data = source.read()
            except (IOError, OSError) as e:
                if e.errno in (errno.EWOULDBLOCK, errno.EAGAIN):
                    return True
                elif e.errno == errno.EINTR:
                    continue
                else:
                    self._open(True, *args)
                    return False
            break

        if not data:
            self._open(True, *args)
            return False

        self._callback(data)

        return True
```

Test System
-----------

### Which version of Quod Libet?
Quod Libet 4.3.0

### Which operating system 
macos 11.4 (Big Sur)

Additional Information
----------------------

<details>
  <summary>Dump log, click to expand</summary>

=== SYSTEM INFORMATION:
Quod Libet 4.3.0
Mutagen 1.41.1
[...]

PR #3761, opened 2021-07-30
Bugfix and clean up for utils.fifo

mjpieters:
Check-list
----------

 * [ ] ~~I've added / updated documentation for any user-facing features.~~

I split this up into distinct commits to make reviewing easier:

- Add more testing around QuodLibetUnixRemote and FIFO
- Fix FIFO._process error handling
- Clean up FIFO open/process handing
- Modernize exception handling
- Final clean-up of remaining issues

This addresses #3745, and should leave the code in an easier-to-maintain
state.

### DIFF
--- a/tests/test_remote.py
+++ b/tests/test_remote.py
@@ -3,13 +3,24 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import os
+import sys
+from pathlib import Path
+from unittest import mock
+
+import pytest
+from gi.repository import GLib, Gio
 from senf import fsn2bytes, bytes2fsn
 
 from . import TestCase, skipIf
 from .helper import temp_filename
 
+import quodlibet
 from quodlibet.remote import QuodLibetUnixRemote
 from quodlibet.util import is_windows
+
+
+QLPATH = str(Path(quodlibet.__file__).resolve().parent.parent)
 
 
 class Mock:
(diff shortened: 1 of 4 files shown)


tests/test_remote.py
```diff
@@ -41,3 +46,106 @@ def test_response(self):
...
+            yield
+
+    def _send_message_remote_proc(self, msg, callback):
+        """Execute QuodLibetUnixRemote.send_message() in a child process"""
+        # Using a child process is the only way to execute send_message, as it
+        # can only be run on the main thread, and blocks waiting for the
+        # response. It can't be run in a separate thread, because it uses
+        # signals to handle timeouts and we can't run the GLib mainloop in a
+        # thread as Gtk has already been initialised on the Python main thread.
+        # Attempting to use the loop in a thread leads to segfaults.
```
mjpieters, line 67: I tried the thread approach first, but that segfaults in `gdk_event_prepare()` on macos. That's not that surprising given the explicit warnings against this in the [GDK threading documentation](https://developer.gnome.org/gdk3/stable/gdk3-Threads.html):

> GTK+, however, is not thread safe. You should only use GTK+ and GDK from the thread `gtk_init()` and `gtk_main()` were called on. This is usually referred to as the “main thread”.

Before the test is even run, `quodlibet._init()` has already called `Gtk.init_args()`.

mjpieters:
Weird, I had the tests succeed in a Docker container, locally. I'll investigate why the fedora and ubuntu tests failed on GH.

mjpieters:
Source of the test failure located (no virtualenv meant the main quodlibet package was not being found in the child process).

declension:
Thanks!

PR #3766, opened 2021-08-09
Follow-up to #3749

slosd:
Change in https://github.com/quodlibet/quodlibet/pull/3749 does not work properly on Windows, so here's a modification to enable it on non-Windows platforms only.

I originally looked through the Gdk source for [shape_combine_region](https://lazka.github.io/pgi-docs/Gdk-3.0/classes/Window.html#Gdk.Window.shape_combine_region). A comment in there suggests that the win32 implementation handles this, but turns out it behaves very differently.

### DIFF
--- a/quodlibet/qltk/views.py
+++ b/quodlibet/qltk/views.py
@@ -7,6 +7,7 @@
 # (at your option) any later version.
 
 import contextlib
+import sys
 
 from gi.repository import Gtk, Gdk, GObject, Pango, GLib
 import cairo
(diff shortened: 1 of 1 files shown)


lazka:
thanks

slosd:
@lazka This means the bug I wanted to fix (https://github.com/quodlibet/quodlibet/issues/1691) remains on all platforms but Linux. Just wanted to let you know, not sure if that means it should be reopened.

PR #3781, opened 2021-08-21
Add a keyboard shortcut for rescanning the library

marmistrz:
Check-list
----------

 * [ ] There is a linked issue discussing the motivations for this feature or bugfix. N/A
 * [ ] Unit tests have been added where possible: N/A
 * [ ] I've added / updated documentation for any user-facing features. N/A

Add a keyboard shortcut for rescanning the library

### DIFF
--- a/quodlibet/qltk/quodlibetwindow.py
+++ b/quodlibet/qltk/quodlibetwindow.py
@@ -990,7 +990,7 @@ class QuodLibetWindow(Window, PersistentWindowMixin, AppWindow):
             name="RefreshLibrary", label=_("_Scan Library"),
             icon_name=Icons.VIEW_REFRESH)
         act.connect('activate', self.__rebuild, False)
-        ag.add_action(act)
+        ag.add_action_with_accel(act, "<Primary>R")
 
         current = config.get("memory", "browser")
         try:


quodlibet/qltk/quodlibetwindow.py
```diff
@@ -990,7 +990,7 @@ def check_updates_handler(*args):
             name="RefreshLibrary", label=_("_Scan Library"),
             icon_name=Icons.VIEW_REFRESH)
         act.connect('activate', self.__rebuild, False)
-        ag.add_action(act)
+        ag.add_action_with_accel(act, "<Primary>R")
```
declension, line 993: If we want this, we should add it to documentation too I'd say

Meriipu:
What does rescanning the library actually do? Could it lead to songs being added/removed? If so accidentally pressing this key-combination might be annoying.

quodlibet/qltk/quodlibetwindow.py
marmistrz: The feature is self-documenting because Gtk will show the keyboard shortcut in the menu. We're not explicitly documenting Ctrl+B for editing bookmarks.

marmistrz:
@Meriipu rescanning the library:
* removes songs which are no longer present in the filesystem
* adds songs which were added to the filesystem

My issue is that, if I start QuodLibet with my external HDD disconnected, then I need to rescan library after connecting the external HDD to have any media listed.

PR #3782, opened 2021-08-21
Make some overlooked strings translatable

Mellthas:
See the individual commits, I think these should be pretty straightforward.

### DIFF
--- a/quodlibet/ext/query/conditional.py
+++ b/quodlibet/ext/query/conditional.py
@@ -6,7 +6,7 @@
 # (at your option) any later version.
 
 from quodlibet import _
-from quodlibet.plugins.query import QueryPlugin, QueryPluginError
+from quodlibet.plugins.query import QueryPlugin, QueryPluginError, markup_for_syntax
 from quodlibet.query._parser import QueryParser
 
 
@@ -15,7 +15,8 @@ class ConditionalQuery(QueryPlugin):
     PLUGIN_NAME = _("Conditional Query")
     PLUGIN_DESC = _("Chooses the query to match based on a condition query.")
     key = 'if'
-    usage = "<b><tt>@(if: condition-query, then-query, else-query)</tt></b>"
+    query_syntax = _("@(if: condition-query, then-query, else-query)")
+    usage = markup_for_syntax(query_syntax)
 
     def search(self, song, body):
         if body[0].search(song):
--- a/quodlibet/ext/query/missing.py
+++ b/quodlibet/ext/query/missing.py
@@ -7,7 +7,7 @@
 
 from gi.repository import Gtk
 from quodlibet import _
-from quodlibet.plugins.query import QueryPlugin
+from quodlibet.plugins.query import QueryPlugin, markup_for_syntax
 from quodlibet.plugins import PluginConfigMixin
 from quodlibet.qltk import Frame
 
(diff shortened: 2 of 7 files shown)


Mellthas:
What should I do about the test `test_terms_letter_case` complaining about `@(python: expression)`? Can the test be skipped for that line?

quodlibet/ext/query/missing.py
```diff
@@ -17,7 +17,7 @@ class MissingQuery(QueryPlugin, PluginConfigMixin):
     PLUGIN_NAME = _("Missing Query")
     PLUGIN_DESC = _("Matches songs without the given tag.")
     key = 'missing'
-    usage = "<b><tt>@(missing: artist)</tt></b>"
+    usage = "<b><tt>" + _("@(missing: artist)") + "</tt></b>"
```
declension, line 20: Is this translatable? `missing` is fixed (name of the query plugin) and `artist` is fixed (tag name)...

quodlibet/ext/query/pythonexpression.py
```diff
@@ -18,16 +18,13 @@ class PythonQuery(QueryPlugin):
     PLUGIN_NAME = _("Python Query")
     PLUGIN_DESC = _("🐍Use Python expressions in queries.")
     key = 'python'
-    usage = """
-<b><tt>@(python: expression)</tt></b>
-
-The variable <tt>s</tt> (or <tt>a</tt>) is the song / album being matched.
+    usage = "<b><tt>" + _("@(python: expression)") + "</tt></b>\n\n" + _(
```
declension, line 21: Let's split this up into varaibles 

e.g. 
```python
query_syntax = _("@(python: expression)")
```
etc

quodlibet/ext/query/conditional.py
```diff
@@ -15,7 +15,7 @@ class ConditionalQuery(QueryPlugin):
     PLUGIN_NAME = _("Conditional Query")
     PLUGIN_DESC = _("Chooses the query to match based on a condition query.")
     key = 'if'
-    usage = "<b><tt>@(if: condition-query, then-query, else-query)</tt></b>"
+    usage = "<b><tt>" + _("@(if: condition-query, then-query, else-query)") + "</tt></b>"
```
declension, line 18: Hmm seeing the changes below, might be worth a helper? (I should have done this originally...)
```python
def markup_for_syntax(text: str) -> str
    return f"<b><tt>{util.escape(text)}</tt></b>"
```

quodlibet/ext/query/missing.py
Mellthas: Oh, yes, totally missed that.

quodlibet/ext/query/pythonexpression.py
Mellthas: Have done that for all the query plugins now.

quodlibet/ext/query/conditional.py
Mellthas: I have added the helper.

Mellthas:
Added one more commit to make the usage text wrap.

declension:
> What should I do about the test `test_terms_letter_case` complaining about `@(python: expression)`? Can the test be skipped for that line? 

Not yet but i've just been updating some of that code so maybe we can add this feature then rebase this

Mellthas:
The tests now pass.

declension:
Thanks!

PR #3783, opened 2021-08-21
Fix string that should not be translatable

Mellthas:
I don't see how translating this would be useful and it could cause the pattern to not work, if the tag names are translated.


declension:
Thanks

PR #3784, opened 2021-08-21
Remove some superfluous tooltip text

Mellthas:
Remove some superfluous tooltip text in the MQTT publisher plugin, which also removes some unnecessary work (and possible confusion) for translators.


declension:
Thanks

Issue #3203, opened 2019-07-03
Gtk.Label with wrap=True and a lot of text uses much more vertical space then necessary


CreamyCookie:
Not sure, if it's okay to ask here, but since this is related to work on a plugin, I hope it is. 

I have been trying to get a multi line label to only use the space it actually needs, but whatever I've tried, it either disappears completely (e.g. with `vexpand = False`) or it takes so much space that the plugin window goes beyond the screen height. (e.g. putting it in another box as suggested [here](https://stackoverflow.com/questions/52553428/gtklabel-expands-vertically-in-a-horizontal-box))

**test case** (put into user plugin folder, e.g. `~/.quodlibet/plugins/`):

```python
from traceback import format_exc
import os.path

from gi.repository import Gtk

from quodlibet import _, app, config, get_user_dir, print_d
from quodlibet.qltk import Icons, ErrorMessage
from quodlibet.plugins import PluginConfig, ConfProp
from quodlibet.plugins.events import EventPlugin
from quodlibet.plugins.songshelpers import is_writable
from quodlibet.util.songwrapper import SongWrapper, check_wrapper_changed


class BatchRate(EventPlugin):
    PLUGIN_ID = "BatchRate"
    PLUGIN_NAME = _("Batch Rate")
    PLUGIN_DESC = _("Uses the content from a special file to rate the last "
                    "played song(s).")
    PLUGIN_ICON = Icons.DOCUMENT_SAVE

    def PluginPreferences(self, parent):
        e = Gtk.Entry()
        e.set_text("test")

        hb = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, spacing=6)
        hb.set_border_width(6)
        hb.pack_start(Gtk.Label(label=_("File:")), False, False, 0)
        hb.pack_start(e, True, True, 0)

        INSTRUCTIONS = "some very long text " * 30
        instructions_lbl = Gtk.Label(label=INSTRUCTIONS, wrap=True)

        main = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=0)
        main.set_border_width(0)
        main.pack_start(hb, False, False, 0)
        main.pack_start(instructions_lbl, False, False, 0)

        return main
```

It seems like the `Gtk.Label`'s size calculation gets buggy once the label uses more than a couple lines, but maybe there's another reason for what's happening.

Mellthas:
[...]

PR #3785, opened 2021-08-21
Fix excess minimum height of plugin window with long description

Mellthas:
See #3203. Compare the minimum size with the *Edit Playcount* plugin selected before and after the change. This reduces unexpected resizes of the plugin window when changing the selected plugin. It does not help with some of the plugins that have a lot of options, they probably need to be handled on an individual basis.

before
after


declension:
Looks much better, thanks

PR #3786, opened 2021-08-21
Various tiny string improvements

Mellthas:
I think the Unicode right arrow → looks a lot nicer compared to ->.


declension:
Thanks

PR #3787, opened 2021-08-21
Fix translated menu labels for playorder plugins and add accelerators

Mellthas:
By specifying the display name explicitly, auto-capitalization of the plugin name is prevented; it produced wrongly capitalized menu entries for languages with different capitalization rules compared to English.

While at it, added accelerators to all entries.


Mellthas:
From looking at the code, having both `display_name` and `accelerated_name` does not really seem to be necessary, `display_name` seems to be only used for sorting the entries and conversion to string. For the sorting, the underscore could be stripped dynamically and for string conversion using the `name` property might suffice (if that string is never displayed to the user under normal circumstances).

declension:
Thanks

PR #3788, opened 2021-08-21
ReplayGain: Disambiguate column heading

Mellthas:
The column is a combined column for track number and title, and as such the heading may need a different translation compared to cases where it only stands for the track number.

E.g. for German *Track* is translated to *Titel-Nr.*, which only refers to the number.


declension:
Thanks

PR #3798, opened 2021-09-09
Fix paths in the .po(t) files

lazka:
They included the build time absolute paths. Change back to
relative to the source root

This is fallout from 3ea15c04daa1fa2


declension:
Sorry about that, didn't realise.

PR #3806, opened 2021-09-12
Make link text translatable in plugin descriptions

Mellthas:
As a followup to #3792, I think it is better to set up the descriptions containing links like this, so that the relevant parts can be properly translated (_ReplayGain_ might need to be translated for languages using a non-latin script). I think it also makes sense to allow the ReplayGain link to be changed to a refer to a localized article.


declension:
Yeah, I thought about that too. Kinda wanted to avoid HTML creeping into translatable text, but if you think this is better for translators I'm happy

Mellthas:
Yes, I think having the markup in the translatable strings can't really be avoided here, as splitting up sentences is bad for translatability in general.

PR #3813, opened 2021-09-23
Tweak `test_terms_letter_case` to account for query-syntax examples

Mellthas:
Small tweak to the `test_terms_letter_case` test method to allow query-syntax examples like `@(python: expression)` to pass. Resolves the test fail in #3782.


declension:
Thanks

Issue #2951, opened 2018-08-29
AudioScrobbler cache when offline


racuna:
Sometimes I have to work with my laptop offline or behind a proxy, and AudioScrobbler can't submit nor connect to post.audioscrobbler.com

Then, when I'm working online again, the scrobbling starts with the current song, without uploading the songs I listened while I was offline.

Yep, sad. I know.

lazka:
In theory (from having a quick look at the code at least) this should already happen so this might be a bug.

racuna:
Thanks for checking it.

bassstorm:
@lazka as a late follow up, I tested that use case today. [Scrobble cache file](https://github.com/quodlibet/quodlibet/blob/8b58a81b077e00b32cc4ad30520ccc8ca70f4406/quodlibet/ext/events/qlscrobbler.py#L92) is created and loaded upon next QL start.
However it would be nice to have QL offloading cache more frequently (not at app quit only, but during the work). My situation is that corp proxy rejects scrobble requests, so in case of a power outage or something I could lose whole day of scrobbles. Frequency could be managed via advanced configuration (I'd prefer to update cache file after each song). Please let me know if you prefer to have a separate new issue for that.

bassstorm:
Closing as per above. Feature request is filed as #3649.

PR #3857, opened 2021-12-11
Scrobble plugin: autosave cache more (#2951 / #3649)

### LINKED ISSUE #3649: Save QL state to disk more frequently
The proposed change
-------------------
As a user I'd like QL to save its state to disk more frequently. Example areas of improvement would be scrobble cache and play queue -- both are currently saved to disk on app exit only.
Could be a core functionality or a plugin.

Source of inspiration: [Autosave & Autobackup](https://www.foobar2000.org/components/view/foo_jesus) plugin for foobar2000.

Benefits of this change
-----------------------
QL can survive any unexpected bad event (e.g. system crash, power outage) and resume from the most actual state snapshot.

Any issues this might cause
---------------------------
No foreseeable issues as those state files are usually relatively small and autosave frequency could be configurable.

declension:
* Add a timer loop for auto saving the queue to disk
 * Name a few things better
 * Add a constant for the interval between saves.
 * Add some (basic) unit tests around the scrobbler plugin.
 * Include one that tests auto-saving (a bit)
 * A bit more logging

Fixes #3649

### DIFF
--- a/quodlibet/ext/events/qlscrobbler.py
+++ b/quodlibet/ext/events/qlscrobbler.py
@@ -87,7 +87,7 @@ class QLSubmitQueue:
     CLIENT = "qlb"
     CLIENT_VERSION = const.VERSION
     PROTOCOL_VERSION = "1.2"
-    DUMP = os.path.join(quodlibet.get_user_dir(), "scrobbler_cache_v2")
+    SCROBBLER_CACHE_FILE = os.path.join(quodlibet.get_user_dir(), "scrobbler_cache_v2")
 
     # These objects are shared across instances, to allow other plugins to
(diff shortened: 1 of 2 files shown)


bassstorm:
Thanks for efforts!
Thinking out loud... I was expecting to see some event-based approach, but this impl seem to be safe as nothing happens if queue is empty, if I'm not missing anything :)

declension:
> Thanks for efforts! Thinking out loud... I was expecting to see some event-based approach, but this impl seem to be safe as nothing happens if queue is empty, if I'm not missing anything :)

Yes I realised the simple way is probably the best here for that reason.

PR #3879, opened 2022-01-06
Queue GUI improvements

jej:
Check-list
----------

See issue #3880

### DIFF
--- a/quodlibet/qltk/queue.py
+++ b/quodlibet/qltk/queue.py
@@ -124,6 +124,27 @@ class QueueExpander(Gtk.Expander):
 
         self.set_label_fill(True)
 
+        clear_item = SmallImageButton(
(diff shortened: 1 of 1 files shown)


declension:
Looks nice. I'm wondering if we want a confirm dialog if the clear button is now much easier to slip on...

jej:
I don't use the queue so often, but I guess that an empty click by mistake is quite improbable here, and if it happens it's of little consequence. Not like delete a playlist or tracks for example.
A smart way could be a cancel button to replace the trash button, until new tracks are added to the queue. This is now a standard practice in some software. Delete without warning but cancel action can follow if needed.
I am not sure to master enough QL to code that, but can play with it!

declension:
Rolling with this for now. Thanks for the PR!

Issue #3872, opened 2022-01-01
" [Unknown]" is added after the filename to music without metadata in i3status-rust music block.


ghost:
Steps to reproduce
------------------

Play a music file without metadata.

Expected Output
---------------

Only the filename should be shown in the music block just like when using strawberry music player.

Actual Output
-------------

All filenames has " Unknown" next to it.

Test System
-----------

### Which version of Quod Libet?
4.4

### Which operating system 
Arch Linux

### If it's audio-related, what back-end?
gstreamer 1.18.5-1

Additional Information
----------------------
" [Unknown]" was also showing in main window title, the display next to the control buttons, and in the tray icon's tooltip display. I was able to fix it by replacing everything with ~basename.
![Screenshot_2022-01-01_18-41-36](https://user-images.githubusercontent.com/59161287/147848837-143e6901-5b82-4344-8c56-be0b92cd6d65.png)

jej:
Hi,
The fix is easy but maybe has some side effects. To test...
https://github.com/jej/quodlibet/tree/fix-issue-3872
J.

PR #3881, opened 2022-01-07
Quick fix of issue #3872

jej:
Check-list
----------

Quick fix for #3872

I noticed this too and I don't know what was the idea behind. It was introduced in 2005: https://github.com/quodlibet/quodlibet/commit/be602696781f947e3d31398ae2742ab00f96a1c6

I find it strange to add a piece of information that says something is missing... Maybe the file name alone is enough.

Hope it helps,
J.

### DIFF
--- a/docs/guide/faq.rst
+++ b/docs/guide/faq.rst
@@ -13,17 +13,19 @@ Do you have a global filter in use? Check the *Browsers* tab in *Preferences*.
 Where does Quod Libet store all its metadata?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The short answer was originally: in ``~/.quodlibet``.
+The short answer was originally: in ``~/.quodlibet/``
 For newer versions of QL it's more complex:
 
+ * On all platforms, if the ``QUODLIBET_USERDIR`` environment variable is set, this will be used
  * On Windows, it will be in your user's ``AppData`` folder under ``Quod Libet``
    (except portable builds)
- * On OS X, it will be in ``~/.quodlibet``.
- * On Linux / Unix systems,
+ * On OS X, it will be in ``~/.quodlibet/``
+ * On Linux / Unix systems:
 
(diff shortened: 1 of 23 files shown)


declension:
I have no strong feelings. Some people might rely (I have at times) on the _Unknown_ to help seeing untagged songs...

@lazka any opinion?

jej:
It's not a big problem for me, everything I listen is tagged. But I am still sensible to illogical gui details... To display "Unknown" (or "Inconnu(e)" in French) is not a really useful information, about what is unknown. Maybe the gui should display an information related to the reason the information is displayed, and not a side effect information like a log/debug information.
So maybe "Untagged track" is better, if this point has to be improved.

declension:
>Maybe the gui should display an information related to the reason the information is displayed, and not a side effect information like a log/debug information.
> So maybe "Untagged track" is better, if this point has to be improved.

Yes that does sound like an improvement actually :+1:

jej:
Yep right, according to the code that's a title concern only.

I am still wondering if it's important to say that a title is missing? :) I have some ripped audio conferences from youteub, and it's not taggued. If I have the mp3 title it's ok.

Maybe an icon for this case could do the job, but don't know what could be this icon? It would be better than a text, because a text may be interpreted as something saved in the track (an idtag containing _Unknown_).

capture_ecran_20220111-131418
More ideas here: https://thenounproject.com/search/icons/?iconspage=2&q=unknown

For now if the code needs a change, something like you say would be ok for me. _(missing title)_ ou _(unknown title)_, without case to don't make it as pretty as a formal title.

jej:
Here an example with unicode char:

capture_ecran_20220111-133702

jej:
> Although I guess it's more _(Missing Title)_ or something (there could be plenty of _other_ tags)

It could be also an advanced option to display or not missing titles...

jej:
Completly lost with git rebase/conflict/etc... WTF! I close and open again?

jej:
Rework in progress...

PR #3882, opened 2022-01-07
New config option for search bar historic count

jej:
Check-list
----------

Add a new option to set in config file how many entries are saved in the search bar.
See also issue #3884


jej:
Have to close, completly lost with git! :|

PR #3905, opened 2022-02-02
Add an option to set how missing titles are displayed

### LINKED ISSUE #3872: " [Unknown]" is added after the filename to music without metadata in i3status-rust music block.

### LINKED ISSUE #3906: An option to set how missing titles are displayed
The proposed change
-------------------
The current way to manage missing tag titles is to display "_filename_ [Unknown]" :
capture_ecran_20220202-131954

I propose an option to allow the user to customize this behaviour, to get something like this:
capture_ecran_20220202-131958

With a new option and Advanced Preferences plugin:
capture_ecran_20220202-131956

Here the default value displays the stem of filename only: `{stem} [untitled {ext}]`
But it could be `{stem}` only or `{basename} (missing title)` or `{basename} [Unknown]"` for the current behaviour or anything else...

See commit #3905 and branch https://github.com/jej/quodlibet/tree/option-missing-title

Benefits of this change
-----------------------
User customization and default template is more clear than `[Unknown]` (IMO)
Fixes #3872

Any issues this might cause
---------------------------
- should not impact performance drastically.
- a new string to translate.

jej:
Check-list
----------

See issue #3906 for details.
Fixes #3872
Fixes #3906

### DIFF
--- a/quodlibet/ext/events/advanced_preferences.py
+++ b/quodlibet/ext/events/advanced_preferences.py
@@ -220,11 +220,11 @@ class AdvancedPreferences(EventPlugin):
                  "its total duration"),
                 label_value_callback=lambda value: f"{int(value * 100)}%"),
             text_config(
-                "browsers", "missing_title_pattern",
-                "Missing title pattern string:",
-                ("Pattern for building title of tracks when title tag is missing. "
+                "browsers", "missing_title_template",
+                "Missing title template string:",
+                ("Template for building title of tracks when title tag is missing. "
                  "{basename} is expanded to the filename of the track, "
-                 "{stem} is the filename without extension, ."
+                 "{stem} is the filename without extension, "
                  "{ext} is the extension (so that {basename}=={stem}.{ext})"))
         ]
 
--- a/quodlibet/ext/events/advanced_preferences.py
+++ b/quodlibet/ext/events/advanced_preferences.py
@@ -218,7 +218,12 @@ class AdvancedPreferences(EventPlugin):
                 "Minimum length proportion to consider a track as played:",
                 ("Consider a track played after listening to this proportion of "
                  "its total duration"),
-                label_value_callback=lambda value: f"{int(value * 100)}%")
+                label_value_callback=lambda value: f"{int(value * 100)}%"),
+            text_config(
+                "browsers", "missing_title_pattern",
+                "Missing title pattern string:",
+                ("Pattern for building title of tracks when title tag is missing. "
+                 "{basename} is replaced by the file name of the track."))
         ]
 
         for (row, (label, widget, button)) in enumerate(rows):
--- a/quodlibet/ext/events/advanced_preferences.py
+++ b/quodlibet/ext/events/advanced_preferences.py
@@ -223,9 +223,9 @@ class AdvancedPreferences(EventPlugin):
                 "browsers", "missing_title_template",
                 "Missing title template string:",
                 ("Template for building title of tracks when title tag is missing. "
-                 "{basename} is expanded to the filename of the track, "
-                 "{stem} is the filename without extension, "
-                 "{ext} is the extension (so that {basename}=={stem}.{ext})"))
+                 "Tags are allowed, like <~basename> <~dirname> <~format> <~length> "
+                 "<~#bitrate>, etc. See tags documentation for details.")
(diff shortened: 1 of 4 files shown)


quodlibet/ext/events/advanced_preferences.py
```diff
@@ -218,7 +218,14 @@ def changed(entry, name, section="settings"):
                 "Minimum length proportion to consider a track as played:",
                 ("Consider a track played after listening to this proportion of "
                  "its total duration"),
-                label_value_callback=lambda value: f"{int(value * 100)}%")
+                label_value_callback=lambda value: f"{int(value * 100)}%"),
+            text_config(
+                "browsers", "missing_title_template",
+                "Missing title template string:",
+                ("Template for building title of tracks when title tag is missing. "
+                 "{basename} is expanded to the filename of the track, "
```
declension, line 226: It might be nicer / "more QL" to use a QL `Pattern` for this, like many other things do (Album Browser, CoverGrid, Playlist), file renames, lots of plugins: scrobbler, Website Search, Gajim, MediaServer, etc) .

Then users can specify `<~basename> <~dirname> <~format> <~length> <~#bitrate> <~filesize>` and more, including any tags it _might_ have and any default logic they want. :up:

jej: Good, I will check that and be back. Thanks.

jej: Hi,
Some changes you can check.
- using Pattern
- 2 new tags: <~basename:stem> and <~basename:humanized>

plz check if it's implemented in the right way... I don't have 360° eyes :)

I will update the test/doc later if it's ok.

Read you, J.

quodlibet/formats/_audio.py
```diff
@@ -519,11 +525,19 @@ def __call__(self, key, default: Any = u"", connector=" - ", joiner=', '):
...
-                return title
+                # build a title with missing_title_template option
+                unknown_track_template = _(config.gettext(
+                    "browsers", "missing_title_template"))
+
+                from quodlibet.pattern import Pattern
+                try:
+                    pattern = Pattern(unknown_track_template)
+                except ValueError:
+                    title = decode_value("~basename", self("~basename"))
```
declension, line 536: :+1:

tests/test_formats__audio.py
```diff
@@ -162,8 +163,13 @@ def test_call(self):
 
         assert self.quux("~basename")
         assert self.quux("~dirname") == os.path.dirname(self.quux("~filename"))
-        assert self.quux("title") == \
-            "%s [Unknown]" % fsn2text(self.quux("~basename"))
+
+        basename = fsn2text(self.quux("~basename"))
+        stem = stem_of_file_name(basename)
+        extension = extension_of_file_name(basename)[1:]
+        assert self.quux("title") == config.gettext(
```
declension, line 170: I think we should _set_ these config items in the test and explicitly assert against expected values (otherwise it's just testing the implementation again slightly)

quodlibet/ext/events/advanced_preferences.py
declension: Yes this looks much better, thanks.

I **don't** think we should add any new _tags_  / keys in this PR though (and thus think `<~basename>` is probably fine for now, as it also mimics existing functionality). All keys need documenting and sometimes there are considerations in other places around the code (e.g. proxies, aggregations), that are far off the  scope of #3872.

Also this string / tooltip should be updated to talk about normal patterns (there are a few others around if you want to use the same style, for translation ease etc)

jej:
> I don't think we should add any new tags / keys in this PR

A wise decision! I had my doubts but it was very tempting to add these new tags...

What about these changes?

PR #3907, opened 2022-02-03
Dequeue after song ended

jej:
Check-list
----------

I am trying to change the behaviour of the playing queue.

Currently, a track is dequeued as soon it starts. I am still a bit confused about this way to do. If I want to rate the song, most of the time I think the current song is a the top of the list but I am wrong. Because the current song disappeared just when it started. The song at the top of the list is the next one, not the current.

I would like to propose the opposite behaviour. The song is dequeue when it finishes. When it's playing, the song remains at the top of the list. So it's easy to rate and see what track of the queue is played.

It's working this way with this patch, even if I didn't really understand all the tweaks about the song workflow in QL... So much calls, tests, events emitted on different classes, etc. I was lost. I just spotted where the track is dequeued in `songmodel.py`.

I had to remove the `self.q.sourced` condition (I didn't understand what is this sourced property) so that the last track of the queue is well removed. If I keep the condition, the last track loops forever and is never removed after it ends.

So, if someone can help, confirm it's ok or not...and wants to work on this feature too, you are welcome!

Read you.

### DIFF
--- a/quodlibet/qltk/songmodel.py
+++ b/quodlibet/qltk/songmodel.py
@@ -25,13 +25,13 @@ class PlaylistMux:
     def __init__(self, player, q, pl):
         self.q = q
         self.pl = pl
-        self._id = player.connect('song-started', self.__song_started)
+        self._id = player.connect('song-ended', self.__song_ended)
         self._player = player
 
     def destroy(self):
         self._player.disconnect(self._id)
 
-    def __song_started(self, player, song):
+    def __song_ended(self, player, song, stopped):
         if song is not None and self.q.sourced:
             iter = self.q.find(song)
             keep_song = config.getboolean("memory", "queue_keep_songs", False)
--- a/quodlibet/qltk/songmodel.py
+++ b/quodlibet/qltk/songmodel.py
@@ -33,12 +33,13 @@ class PlaylistMux:
 
     def __song_ended(self, player, song, stopped):
         if song is not None and self.q.sourced:
-            iter = self.q.find(song)
             keep_song = config.getboolean("memory", "queue_keep_songs", False)
-            if iter and not keep_song:
-                self.q.remove(iter)
-                # we don't call _check_sourced here since we want the queue
-                # to stay sourced even if no current song is left
+            if not keep_song:
+                iter = self.q.find(song)
+                if iter:
+                    self.q.remove(iter)
+                    # we don't call _check_sourced here since we want the queue
+                    # to stay sourced even if no current song is left
 
     @property
     def current(self):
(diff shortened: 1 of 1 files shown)


bassstorm:
Agree, I remember my first days being a QL user, that was something I was puzzled with for sure! I think described behavior should really make users feel right at home.
P.S. Sorry, don't have sufficient python experience as well as repo right to review.

jej:
Yes, I am the early adopter you mention!

afontenot:
The approach taken here won't work.

Because you removed `self.q.sourced`, the function no longer checks to see whether the queue is the currently active play source. Therefore, if you have a song in the queue, and start playing the song from the songlist instead, the song will be removed from the queue after it's done playing even though it wasn't played *from* the queue.

Also, the `song-ended` signal probably doesn't do what you think it does. This signal is sent when the song ends for *any* reason, including because you clicked play on another song. So imagine you have a queue, you double click on a song to play it, and halfway through you decide to play a song from the songlist instead. The song from the queue gets removed, because it "ended". Not what you would expect. You probably want to be able to come back to the song in the queue unless the song actually ended.

There's a bigger issue though (not your fault): because the song object is stateless, the code has to use `self.q.find` to locate the object in the queue. That *currently* works by accident because the code is removing the song from the queue while it's still playing: there's a fast path in `TrackCurrentModel.find` to return the current song if it's identical to the one requested. But if you switch this to removing the song after it's stopped playing, then `find` has to search through the whole queue for the song. This means that if you have the same song in the queue twice (lots of good reasons for this to happen), only the first instance of the song will be removed, not the one that actually played.

I do agree with you about what the intuitive behavior is. This part of the code should probably be reworked in general, but something like the following approach seems fine to me:

1. Remove `_check_sourced` from the code and instead have a `self.__current_source` that points to one of `self.q` or `self.pl` and is updated whenever the user requests playing from a different source. (Part of the problem is caused by this method choosing to set the songlist source as active simply because the last song in the queue ended.)
2. Change `__song_ended` to have some knowledge of *why* the song ended. Only remove the song if the user clicked next or the song ended normally.
3. Playlist models have a `current_iter` property that gets an actual reference to the item in the tree, so you can remove it precisely. It should also have a `prev_iter` property. (It has a `last_current` song item, but this is a stateless piece of metadata about the song, not a reference to the item in the tree.)

jej:
Thanks for the investigation and explanations.

So, I will not work on this if the scope of changes is so wide...

declension:
@jej shall we close this PR for now then?

jej:
Yep no problem. If queue is refactored one day...
Thx

Issue #3327, opened 2020-03-21
People column in songlist lists artist before albumartist


elcste:
Steps to reproduce
------------------
Albumartist should be listed before artist. The [docs agree](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html):

> The internal ~people tag combines the following tags to one: albumartist, artist, author, composer, ~performers, originalartist, lyricist, arranger, conductor in this exact order.

This seems to be true some places, like the People line on the Album List pane. But the People column in the songlist lists artist before albumartist.

Expected Output
---------------
The People column in the songlist should list albumartist before artist.

Actual Output
-------------
The People column in the songlist lists artist before albumartist.

Test System
-----------

I tried removing config files to make sure my settings didn't affect this.

### Which version of Quod Libet?
4.4 (development, 7025a5833) [from unstable PPA]

### Which operating system 
Ubuntu 19.10

Additional Information
----------------------
I'm not a programmer so I don't know if this is relevant, but I found this source entry that lists artist before albumartist:
https://github.com/quodlibet/quodlibet/blob/master/quodlibet/quodlibet/formats/_audio.py#L45

declension:
I suspect this was to deal with "Various Artists" type of problem (compilations)

Update: yes (by me lol) - https://github.com/quodlibet/quodlibet/commit/501561dab47cbb71184d6d636eec6182799ccd74
See discussion here in #1034

elcste:
Forgot to close this when I updated the documentation to match the real behavior…

PR #3938, opened 2022-03-11
Corrected description of ~people internal tag.

elcste:
Check-list
----------

In https://github.com/quodlibet/quodlibet/issues/3327, I pointed out the [docs](https://quodlibet.readthedocs.io/en/latest/guide/tags/internal_tags.html) say the ~people internal tag combines albumartist, artist, (etc), in that order. But the [code](https://github.com/quodlibet/quodlibet/blob/master/quodlibet/formats/_audio.py#L50) has the opposite order for those two. It was [pointed out](https://github.com/quodlibet/quodlibet/issues/3327#issuecomment-817193438) this was intentional, so I'm fixing the docs.


declension:
Thanks

PR #3947, opened 2022-03-22
Allow auto-playing added songs in Queue Only mode

### LINKED ISSUE #3381: Should be possible to start playing a song with a double click in Queue Only mode
The proposed change
-------------------
I'm using the Queue Only mode plugin which gets Quod Libet closer to my preferred behavior, but it's missing one thing which would make it similar to how many other music players behave. Suppose my queue is empty. If I double click a song, it will be added to my queue. That's good ... but I also want it to start playing that song out of the queue immediately. Similarly, if I've just played a song and I'm now at the end of the queue, if I double click a song it gets added into the queue, but I also want it to start playing right away.

Take it a step further: why not start playing anything I double click *unconditionally*? That is, add the item I double click to the end of the queue, and immediately skip the rest of the queue and start playing that item? I think that's actually a reasonable default; there's already an option to add an item to the queue (ctrl-enter, or right click -> add to queue), so it makes sense to think that a double-click signals "intent to play". In any case, I think a checkbox to enable this option in the window for the Queue Only plugin would be very useful.

(Actually, one further step is possible, though maybe I should make a different feature request for this. Maybe double clicking on a song should add the entire album that song is in to the queue, skip to the specific song in the album that was double clicked, and start playing it. This is the sort of behavior that I intuitively expect from a queue-oriented media player.)

Benefits of this change
-----------------------
In queue only mode, it currently requires multiple user actions to make a song start playing. This extra complexity is unnecessary when the user already signaled their intent to play the song. I suspect this would benefit anyone looking for an experience in line with many traditional, popular media players.

Any issues this might cause
---------------------------
If this was enabled by default, it might throw off existing users who have come to expect other behaviors. If this is a significant concern, I don't see any issue with putting it behind an option, since it's really immaterial to me so long as I can enable it.

afontenot:
When Queue Only mode is enabled, the user has multiple
neutral methods to add a song to the queue: both
Ctrl+Enter and the right click menu. Double clicking a
song usually indicates intent to play it immediately.

Queue Only mode changes the double click action to add
the song to the queue instead, but some users might like
Quod Libet to respect their intent and jump to the newly
added queue item and begin playing it. This adds the
ability to do that to the Queue Only plugin, and puts it
behind a preference.

Fixes #3381.

Implementation notes:

The code for jumping in the queue is specifically designed
to jump to the first added item in the queue. This is to
make implementing a double click action for albums easier
in the future. (QL will need to jump to the first song in
the added album.)

A change had to be made to the player code, because it was
emitting an explicit go_to when the main window restores
the last playing song when starting up and this was getting
caught by the plugin. Intuitively, restoring a previously
playing song is an implicit action by the player.

Check-list
----------

Note isn't really any documentation or tests for this plugin unfortunately, so I have not added anything pertaining to this change.

### DIFF
--- a/quodlibet/ext/playorder/queue.py
+++ b/quodlibet/ext/playorder/queue.py
@@ -8,9 +8,14 @@
 
 from quodlibet import _
 from quodlibet import app
+from quodlibet import config
+from quodlibet import qltk
 from quodlibet.plugins.playorder import ShufflePlugin
+from quodlibet.qltk.ccb import ConfigCheckButton
 from quodlibet.qltk import Icons
 from quodlibet.qltk.playorder import OrderInOrder
+
+from gi.repository import Gtk
 
 
 class QueueOrder(ShufflePlugin, OrderInOrder):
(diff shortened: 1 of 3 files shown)


declension:
Thanks!

PR #3975, opened 2022-04-10
Fix display pattern not updating in browsers

slosd:
Check-list
----------

The compiled display pattern never was updated and as a result browsers could not use it to update item labels. 
The new pattern string was saved to an instance attribute (`self.__pattern_text`) but compiled from a class attribute (`cls.__pattern_text`).


declension:
Thanks!

PR #3977, opened 2022-04-18
ci: add informational Codecov status checks

thomasrockhu-codecov:
Check-list
----------

Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,9 @@ comment: false
 coverage:
   status:
     project:
-      informational: true
+      default:
+        informational: true
     patch:
-      informational: true
+      default:
+        informational: true
     changes: false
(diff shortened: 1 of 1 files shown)


lazka:
I don't care for them tbh. Maybe someone else wants them.

declension:
> I don't care for them tbh. Maybe someone else wants them.

If it helps see where coverage needs adding I'm happy to give it a go. 

We ~can~ _should_ revert if it's not useful though

lazka:
@thomasrockhu-codecov is this just about the extra PR checks? I initially thought this is about the PR comments codecov used to make. If yes, than it's quite nice, thanks.

PR #4016, opened 2022-06-14
Fix some Bash warnings in Win installer

declension:
Thanks to shellcheck really...

### DIFF
--- a/dev-utils/win_installer/_base.sh
+++ b/dev-utils/win_installer/_base.sh
@@ -112,7 +112,7 @@ flake8==3.9.2
 "
 
     build_pip install --no-binary ":all:" \
-        --force-reinstall "$(echo "$PIP_REQUIREMENTS" | tr "\\n" " ")"
+        --force-reinstall $(echo "$PIP_REQUIREMENTS" | tr "\\n" " ")
 
     build_pacman --noconfirm -Rdds \
         mingw-w64-"${ARCH}"-shared-mime-info \
--- a/dev-utils/win_installer/_base.sh
+++ b/dev-utils/win_installer/_base.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 # Copyright 2016 Christoph Reiter
+#           2022 Nick Boultbee
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


dev-utils/win_installer/_base.sh
```diff
@@ -203,7 +206,7 @@ function cleanup_after {
     rm -Rf "${MINGW_ROOT}"/libexec
     rm -Rf "${MINGW_ROOT}"/share/gtk-doc
     rm -Rf "${MINGW_ROOT}"/include
-    rm -Rf "${MINGW_ROOT}"/var
+    rm -Rf "${MINGW_ROOT:?}"/var
```
lazka, line 209: Do you know why it does the error checking only in this line and not the 50 others?

dev-utils/win_installer/_base.sh
```diff
@@ -111,7 +112,7 @@ flake8==3.9.2
 "
 
     build_pip install --no-binary ":all:" \
-        --force-reinstall $(echo "$PIP_REQUIREMENTS" | tr ["\\n"] [" "])
+        --force-reinstall "$(echo "$PIP_REQUIREMENTS" | tr "\\n" " ")"
```
lazka, line 115: that's not the same, the lines need to be passed as separate arguments

dev-utils/win_installer/_base.sh
declension: Yes it "knows" about certain top-level directories that would be a disaster to remove and `/var` is one of them. I guess we could add this everywhere. 
Though might be easier to `test` for it earlier and exit if empty?

dev-utils/win_installer/_base.sh
declension: Oh do you mean the outer quotes?

dev-utils/win_installer/_base.sh
lazka: I see, not sure..

dev-utils/win_installer/_base.sh
lazka: Yes

declension: Fixed, thanks!

lazka:
lgtm

PR #4027, opened 2022-06-26
builtin cover: fix handling of invalid glob ranges with Python 3.10.5+

lazka:
Previously Python would raise if an invalid range was given
to glob, but with 3.10.5 they fixed it to not match anything.
https://github.com/python/cpython/issues/89973

Our tests depended on the previous logic and treating the glob pattern
as a literal file name in that case.

One could argue that this is wrong since a range that doesn't contain anything
should also not match anything, so wrap glob() to make it not match for all
Python versions in that case and adjust the tests accordingly.

This should fix the Windows CI, which is currently the only job using 3.10.5

### DIFF
--- a/quodlibet/util/cover/built_in.py
+++ b/quodlibet/util/cover/built_in.py
@@ -100,6 +100,15 @@ class FilesystemCover(CoverSourcePlugin):
         base = self.song('~dirname')
         images = []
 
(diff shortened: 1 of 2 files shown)


declension:
LGTM, shall we merge?

lazka:
sure, go ahead ;)

PR #4029, opened 2022-06-26
poetry: update dependencies

lazka:
and since we bump the sphinx minimum version, also bump it for readthedocs


pyproject.toml
```diff
@@ -34,13 +34,13 @@ plugins = ["musicbrainzngs", "pyinotify", "dbus-python", "paho-mqtt", "soco", "p
...
 pytest = '^7.1.1'
 polib = '^1.1.0'
-sphinx = '^4.4.0'
-sphinx_rtd_theme = '*'
+sphinx = '^5.0.2'
+sphinx_rtd_theme = '^1.0.0'
 sphinx-autobuild = "^2021.3.14"
 xvfbwrapper = { version = "^0.2", platform = "linux" }
-flake8 = "^3.9.2"
+flake8 = "^4.0.1"
```
declension, line 41: If we change this _here_... it now appears in a bunch of places too, perhaps it should be consistent

pyproject.toml
```diff
@@ -24,7 +24,7 @@ pygobject = "^3.34.0"
 # Optional (mainly plugins)
 musicbrainzngs = { version = "0.*", optional = true }
 dbus-python = { version = "*", optional = true }
-soco = { version = "^0.22.2", optional = true }
+soco = { version = "^0.28.0", optional = true }
```
declension, line 27: Do we _need_ this higher version? I definitely haven't used any of its features

Happy for it to _incidentally_ be higher (i.e. the lock file specifies it, but the interface (i.e. here) IMO should keep only what it requires

lazka: No idea, I'm just updating everything. Should I revert this?

pyproject.toml
lazka: done

pyproject.toml
lazka: reverted

PR #4083, opened 2022-08-06
the sync_to_device plugin shouldn't delete files named 'cover.jpg'

### LINKED ISSUE #4082: The sync_to_device plugin shouldn't delete files named 'cover.jpg'
I am using an [Automatic Cover Art Downloader](https://github.com/desbma/sacad) which adds 'cover.jpg' files in each album folder. The sync_to_device plugin deletes them again in the target folder. It shall not.

A nicer option would be to sync the cover files from the library to the target folder, too. But that's more complex to implement.

Tamriel:
Fixes  #4082

### DIFF
--- a/quodlibet/ext/events/synchronize_to_device.py
+++ b/quodlibet/ext/events/synchronize_to_device.py
@@ -1082,7 +1082,7 @@ class SyncToDevice(EventPlugin, PluginConfigMixin):
             for dirname in dirs:
                 dir_path = os.path.realpath(os.path.join(root, dirname))
                 files = os.listdir(dir_path)
-                last_file_is_cover = files[0] == 'cover.jpg'
+                last_file_is_cover = files and files[0] == 'cover.jpg'
                 if not files or last_file_is_cover:
                     entry = Entry(None)
                     entry.filename = dir_path
--- a/quodlibet/ext/events/synchronize_to_device.py
+++ b/quodlibet/ext/events/synchronize_to_device.py
@@ -254,7 +254,8 @@ class SyncToDevice(EventPlugin, PluginConfigMixin):
               'edited before starting the synchronization.'),
             Icons.DIALOG_WARNING, visible=False)
         self.status_deletions = self._label_with_icon(
-            _('Existing files in the destination path will be deleted!'),
+            _("Existing files in the destination path will be deleted (except "
+              "files named 'cover.jpg')!"),
             Icons.DIALOG_WARNING, visible=False)
 
         # Section for previewing exported files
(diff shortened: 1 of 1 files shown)


quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -741,7 +742,8 @@ def _run_preview(self):
         for root, __, files in os.walk(self.expanded_destination):
             for name in files:
                 file_path = os.path.join(root, name)
-                if file_path not in export_paths:
+                if file_path not in export_paths and \
+                        not file_path.__contains__("cover.jpg"):
```
declension, line 746:
```suggestion
                        "cover.jpg" not in file_path:
```

quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -1079,14 +1081,18 @@ def _remove_empty_dirs(self):
         for root, dirs, __ in os.walk(self.expanded_destination, topdown=False):
             for dirname in dirs:
                 dir_path = os.path.realpath(os.path.join(root, dirname))
-                if not os.listdir(dir_path):
+                files = os.listdir(dir_path)
```
declension, line 1084: Though it was there before, `listdir` shouldn't be necessary if `os.walk` is used; the third tuple items is filenames.

This will also be a lot faster since it avoids scanning twice

quodlibet/ext/events/synchronize_to_device.py
```diff
@@ -1079,14 +1081,18 @@ def _remove_empty_dirs(self):
...
+                last_file_is_cover = files and files[0] == 'cover.jpg'
```
declension, line 1085: That's the _first_ file, no? What if there are lots and `'cover.jpg'` just happens to be the first?

quodlibet/ext/events/synchronize_to_device.py
Tamriel: Oh yes, the elegance of Python! I do not use it much anymore, thanks for your feedback!

quodlibet/ext/events/synchronize_to_device.py
Tamriel: If there are other files than the music and `cover.jpg`, deleting the folder will fail. And I think that is okay, since the user shall notice that's not the way how this plugin sync process is designed. Then an error message explaining that would be nice I think.

declension: Ah, yeah that makes sense. Perhaps you could add a comment to explain this in the code...

declension:
Thanks

PR #4096, opened 2022-08-14
Softer cover art outline

declension:
Looks much nicer to me, especially for white(ish) cover art.

### Before
Screenshot from 2022-08-14 20-06-24 Screenshot from 2022-08-14 20-06-57 Screenshot from 2022-08-14 20-08-14 Screenshot from 2022-08-14 20-08-48

### After
Screenshot from 2022-08-14 20-13-39 Screenshot from 2022-08-14 20-13-50 Screenshot from 2022-08-14 20-15-14 Screenshot from 2022-08-14 20-15-40

### DIFF
--- a/quodlibet/qltk/image.py
+++ b/quodlibet/qltk/image.py
@@ -108,6 +108,7 @@ def add_border_widget(pixbuf, widget):
 
     context = widget.get_style_context()
     color = context.get_color(context.get_state())
+    color.alpha *= 0.1
     scale_factor = widget.get_scale_factor()
     border_radius = get_border_radius() * scale_factor
 


bassstorm:
Agree, new style looks a bit nicer

slosd:
In my case this unfortunately almost removes the rounded border:

### Before
cover_border_a

### After
cover_border_b

declension:
@slosd I see... any suggestions? Could make it configurable, or have a compromise alpha (i.e. darker)

slosd:
@declension I get a good result when I make the color lighter and keep alpha at 1. Seems like it could be a general solution without the need for configuration.

```patch
diff --git a/quodlibet/qltk/image.py b/quodlibet/qltk/image.py
index f66e28387..7cbbb4dd7 100644
--- a/quodlibet/qltk/image.py
+++ b/quodlibet/qltk/image.py
@@ -108,11 +108,12 @@ def add_border_widget(pixbuf, widget):
 
     context = widget.get_style_context()
     color = context.get_color(context.get_state())
-    color.alpha *= 0.1
+    rgba = [c + 0.7 * (1 - c) for c in color]
+    lighter_color = Gdk.RGBA(red=rgba[0], green=rgba[1], blue=rgba[2])
     scale_factor = widget.get_scale_factor()
     border_radius = get_border_radius() * scale_factor
 
-    return add_border(pixbuf, color, width=scale_factor, radius=border_radius)
+    return add_border(pixbuf, lighter_color, width=scale_factor, radius=border_radius)
 
 
 def scale(pixbuf, boundary, scale_up=True, force_copy=False):
```
The factor 0.7 can be adjusted.

declension:
Cool - but have you tried that in dark mode - won't that make it worse?

slosd:
Good point. One possibility is to change the formula a bit: `[c + 0.7 * (0.5 - c) for c in color]`. 

I created an issue, because I'm not sure if this implementation is actually a solution. As I was playing around with it the round corners would still be relatively small; smaller than expected for the border-radius used.

Issue #2840, opened 2018-05-09
Cover grid scroll performance


slosd:
I experience severe performance issues in the cover grid browser when I'm scrolling through the albums.

This only happens when I'm **not** using gnome-shell and it happens with or without a desktop session, with several different window managers (tried Mate's marco, awesomewm, xmonad).

Applying [this changeset](https://github.com/quodlibet/quodlibet/pull/2481/files#diff-725a1363e50790bee2c8cd9a7aff8ded) proposed in a pull request fixes the scroll performance issue.

> openSUSE Leap 42.3
> QuodLibet 4.0.2
> 
> Audio device: GStreamer (pulsesink)
> Python: 3.5.4
> Mutagen: 1.40.0
> GTK+: 3.20.10 (X11)
> PyGObject: 3.27.1
> GStreamer: 1.8.3.0
> 
> with proprietary nvidia driver v390.48

lazka:
Problem is that with the patch the view is unusable with GS here, so we need a different solution.

lazka:
I can't reproduce with xfce (with and without compositing) and openbox.

KoO-303:
I also have this with gnome and xfce. Very sluggish especially after first opening Quodlibet but does get faster as time goes by.Never quiet right though .With version 3 too.
Linux Debian Manjaro Antergos

slosd:
I've been trying to figure out what impact the callbacks of the cell renderers used for the items in the cover grid have: 
https://github.com/quodlibet/quodlibet/blob/36ef24bb56bd0b3b301a912b2bc696398313bff6/quodlibet/quodlibet/browsers/covergrid/main.py#L249 https://github.com/quodlibet/quodlibet/blob/36ef24bb56bd0b3b301a912b2bc696398313bff6/quodlibet/quodlibet/browsers/covergrid/main.py#L277

These are called **very** frequently and I think, potentially, this is an issue. Here are my results **with 1416 albums** in the cover grid. Unfortunately the data is just from one run per version and chart, so no averages. Each chart shows the number of calls per album with the changes I mentioned in my initial post applied ("patch") as well as without ("nopatch"). The charts only show the data for the first 100 albums because there is no change for albums beyond it. 

First, as a baseline, these are the calls (y-axis) per album (x-axis) after starting quodlibet and letting the cover grid browser initialize:
[...]

Issue #2181, opened 2017-01-05
Cover Grid: Speed up switching to the browser


CreamyCookie:
I don't know how achievable that is, but it would be great if the cover grid album would open faster. Switching from the album list to the cover grid takes about 5s here, while switching back to the album list takes less than 2s. This happens even when repeated within the same session.

declension:
Yep, it's too slow. 

This got mentioned in the original PR. See #2071

declension:
Fixed in #4101

PR #4101, opened 2022-08-16
Covergrid using FlowBox

slosd:
Check-list
----------

This is an effort to address the issues described in #2840 by using a [FlowBox](https://lazka.github.io/pgi-docs/#Gtk-3.0/classes/FlowBox.html) widget which unlike the `IconView` widget is better optimized for long lists.

My personal experience with the current CoverGrid is still that it is unusable because it freezes quodlibet for minutes when scrolling, resizing the window or toggling the menu bar. With these changes the CoverGrid is very responsive, although not entirely smooth until all covers have been loaded.

Some notes:
* Number of "columns" is determined dynamically but limited to a maximum of 10. Maybe the max. should be higher.
* Covers are loaded and _labels_ are generated as they scroll into view. This decreases the time it takes for the browser to be ready for interaction. 
* When selection of an item is lost (e.g. selected item gets filtered or gets moved when tags change) no other item is selected and the song list is cleared. This is different from the "album" browser in which the selection moves to the next item.
* I did check that instances are GCed, but this is an aspect of PyGObject I couldn't find much documentation on. So I'm not sure it's handled correctly.
* One model implements the interface `Gio.ListModel` which is expected by `FlowBox`. This is a wrapper around `TreeModelFilter` and `TreeModelSort` which are doing the actual work. When quodlibet migrates to GTK 4 this can be simplified with [FilterListModel](https://docs.gtk.org/gtk4/class.FilterListModel.html) and [SortListModel](https://docs.gtk.org/gtk4/class.SortListModel.html).
* Tests are adapted from the ones for the "album" browser.

### DIFF
--- a/quodlibet/browsers/covergrid/models.py
+++ b/quodlibet/browsers/covergrid/models.py
@@ -68,7 +68,7 @@ class AlbumListCountItem(AlbumListItem):
     def n_albums(self):
         return self.__n_albums
 
-    @n_albums.setter
+    @n_albums.setter # type: ignore
     def n_albums(self, value):
         self.__n_albums = value
         self.format_label()
@@ -160,7 +160,7 @@ class AlbumListFilterModel(GObject.Object, Gio.ListModel):
     def include_item_all(self):
         return self.__include_item_all
 
-    @include_item_all.setter
+    @include_item_all.setter # type: ignore
     def include_item_all(self, value):
         if self.__include_item_all == value:
             return
@@ -172,7 +172,7 @@ class AlbumListFilterModel(GObject.Object, Gio.ListModel):
     def filter(self):
         return self.__filter
 
-    @filter.setter
+    @filter.setter # type: ignore
     def filter(self, value):
         b = background_filter()
         if b is None and value is None:
--- a/quodlibet/browsers/covergrid/models.py
+++ b/quodlibet/browsers/covergrid/models.py
@@ -1,0 +1,254 @@
+# Copyright 2022 Thomas Leberbauer
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+
+from gi.repository import GObject, Gio
+
+from quodlibet import _, ngettext, app
+from quodlibet.qltk.models import ObjectModelSort, ObjectStore, ObjectModelFilter
+from quodlibet.util.library import background_filter
+
+
+class AlbumListItem(GObject.Object):
+
+    def __init__(self, album=None):
+        super().__init__()
+        self._album = album
+        self._cover = None
+        self._label = None
+
+        self.connect('notify::album', self._album_changed)
+
+    def load_cover(self, size, cancelable=None):
+        def callback(cover):
+            self._cover = cover
+            self.notify('cover')
+
+        manager = app.cover_manager
+        # Skips this during testing
+        if manager:
+            manager.get_pixbuf_many_async(
+                self._album.songs, size, size, cancelable, callback)
+
+    def format_label(self, pattern):
+        self._label = pattern % self._album
+        self.notify('label')
+
+    def _album_changed(self, model, prop):
+        self._label = None
+
+    @GObject.Property
+    def album(self):
+        return self._album
+
+    @GObject.Property
+    def cover(self):
+        return self._cover
+
+    @GObject.Property
+    def label(self):
+        return self._label
+
+
+class AlbumListCountItem(AlbumListItem):
+
+    def load_cover(self, *args, **kwargs):
+        self.notify('cover')
+
+    def format_label(self, pattern=None):
+        n = self.__n_albums
+        self._label = ('<b>%s</b>\n' +
+            ngettext('%d album', '%d albums', n)) % (_('All Albums'), n)
+        self.notify('label')
+
+    @GObject.Property
+    def n_albums(self):
+        return self.__n_albums
+
+    @n_albums.setter
+    def n_albums(self, value):
+        self.__n_albums = value
+        self.format_label()
+
+
+class AlbumListModel(ObjectStore):
+
+    def __init__(self, library):
+        super().__init__()
+        self.__library = library
+
+        albums = library.albums
+        self.__sigs = [
+            albums.connect('added', self._add_albums),
+            albums.connect('removed', self._remove_albums),
+            albums.connect('changed', self._change_albums)
+        ]
+
+        self.append(row=[AlbumListCountItem()])
+        self.append_many(AlbumListItem(a) for a in albums.values())
+
+    def destroy(self):
+        albums = self.__library.albums
+        for sig in self.__sigs:
+            albums.disconnect(sig)
+        self.__library = None
+        self.clear()
+
+    def _add_albums(self, library, added):
+        self.append_many(AlbumListItem(a) for a in added)
+
+    def _remove_albums(self, library, removed):
+        removed_albums = removed.copy()
+        iters_remove = []
+        for iter_, item in self.iterrows():
+            album = item.album
+            if album is not None and album in removed_albums:
+                removed_albums.remove(album)
+                iters_remove.append(iter_)
+                if not removed_albums:
+                    break
+        for iter_ in iters_remove:
+            self.remove(iter_)
+
+    def _change_albums(self, library, changed):
+        changed_albums = changed.copy()
+        for iter_, item in self.iterrows():
+            album = item.album
+            if album is not None and album in changed_albums:
+                changed_albums.remove(album)
+                self.iter_changed(iter_)
+                if not changed_albums:
+                    break
+
+
+class AlbumListFilterModel(GObject.Object, Gio.ListModel):
+
+    __filter = None
+
+    def __init__(self, child_model=None, include_item_all=True, **kwargs):
+        super().__init__(**kwargs)
+
+        self.__include_item_all = include_item_all
+
+        self._model = model = ObjectModelFilter(child_model=child_model)
+        self.__item_all = self._get_item(0)
+        self._update_n_albums()
+
+        # Tell the tree model that all nodes are visible, otherwise it does not
+        # emit the "row-changed" signal.
+        for row in model:
+            model.ref_node(row.iter)
+
+        model.set_visible_func(self._apply_filter)
+        self.__sigs = [
+            model.connect('row-changed', self._row_changed),
+            model.connect('row-inserted', self._row_inserted),
+            model.connect('row-deleted', self._row_deleted),
+            model.connect('rows-reordered', self._rows_reordered)
+        ]
+
+    def destroy(self):
+        model = self._model
+        for sig in self.__sigs:
+            model.disconnect(sig)
+        self._model = None
+
+    @GObject.Property
+    def include_item_all(self):
+        return self.__include_item_all
+
+    @include_item_all.setter
+    def include_item_all(self, value):
+        if self.__include_item_all == value:
+            return
+        self.__include_item_all = value
+        removed, added = (0, 1) if value else (1, 0)
+        self.items_changed(0, removed, added)
+
+    @GObject.Property
+    def filter(self):
+        return self.__filter
+
+    @filter.setter
+    def filter(self, value):
+        b = background_filter()
+        if b is None and value is None:
+            f = None
+        elif b is None:
+            f = lambda item: value(item.album)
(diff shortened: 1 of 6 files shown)


slosd:
I believe the failures in _ubuntu-poetry_ are false positives. I'm unsure what's wrong in the _flatpak_ tests.

quodlibet/browsers/covergrid/models.py
```diff
@@ -0,0 +1,254 @@
...
+
+class AlbumListCountItem(AlbumListItem):
+
+    def load_cover(self, *args, **kwargs):
+        self.notify('cover')
+
+    def format_label(self, pattern=None):
+        n = self.__n_albums
+        self._label = ('<b>%s</b>\n' +
+            ngettext('%d album', '%d albums', n)) % (_('All Albums'), n)
```
declension, line 64: We should escape that really, in case translations end up with markup

quodlibet/browsers/covergrid/models.py
```diff
@@ -0,0 +1,254 @@
...
+                self._album.songs, size, size, cancelable, callback)
+
+    def format_label(self, pattern):
+        self._label = pattern % self._album
+        self.notify('label')
+
+    def _album_changed(self, model, prop):
+        self._label = None
+
+    @GObject.Property
```
declension, line 43: That's cool, didn't know you could do that... :+1:

quodlibet/browsers/covergrid/models.py
```diff
@@ -0,0 +1,254 @@
...
+        if index >= 0:
+            self.items_changed(index, 1, 0)
+            self._update_n_albums()
+
+    def _rows_reordered(self, model, path, iter, new_order):
+        n = len(self)
+        self.items_changed(0, n, n)
+
+
+class AlbumListSortModel(ObjectModelSort):
```
declension, line 253: Why do we need this? Perhaps you can add a docstring here...

quodlibet/browsers/covergrid/models.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class AlbumListCountItem(AlbumListItem):
+
+    def load_cover(self, *args, **kwargs):
+        self.notify('cover')
+
+    def format_label(self, pattern=None):
+        n = self.__n_albums
+        self._label = ('<b>%s</b>\n' +
```
declension, line 63: This is quite confusing to read for me; why not just substitute the _All Albums_ immediately?

Also why are we ignoring `pattern`?

declension:
This is great, thanks @slosd !

Did you get any performance comparisons? Having pulled and quickly played I can confirm unscientifically that browser startup time and reflow are both _much_ better with `FlowBox` (as expected) :rocket: 

Can't see any immediate bugs or gotchas either so far.

A few things that would make it easier - we're trying to add (some) type hints to all new code - could you add some here, especially where they help clear up parameter or return types. And a couple of docstrings on the key new classes would help too :+1:

declension:
> I believe the failures in _ubuntu-poetry_ are false positives. I'm unsure what's wrong in the _flatpak_ tests.

Yes a few tests are flaky on some systems. I've re-run the flatpak and it's fine now :heavy_check_mark: 

The Mypy (Poetry) ones issues I replicate locally:
```
E     quodlibet/browsers/covergrid/models.py:75: error: Name "n_albums" already defined on line 71
E     quodlibet/browsers/covergrid/models.py:167: error: Name "include_item_all" already defined on line 163
E     quodlibet/browsers/covergrid/models.py:179: error: Name "filter" already defined on line 175
```
I don't think Mypy believes that `Gobject.property` is like a normal Python `@property` (I assume it is?) so complains about redefinitions.
Adding `# type: ignore` on those three lines fixes these so I suggest we just do that for now, as the decorator seems nice enough to want.

quodlibet/browsers/covergrid/models.py
slosd: We don't _need_ it. It's an indirection to make `models.py` the single place to change implementation of the models.

quodlibet/browsers/covergrid/models.py
slosd: Are you referring to this? https://github.com/quodlibet/quodlibet/blob/fec3e52ee97014b7f692314b581ae44fa7817485/quodlibet/util/__init__.py#L203-L205

quodlibet/browsers/covergrid/models.py
slosd: I don't quite understand. Like this?
```python
self._label = (('<b>%s</b>\n' % _('All Albums')) +
    (ngettext('%d album', '%d albums', n) % n))
```
I guess I just went for doing it in one substitution instead of two.

Your second point: `AlbumListCountItem` inherits from `AlbumListItem` and overrides `format_label`. So `pattern` is included to match the method signature from the base class, but the "All albums" entry doesn't actually use it.

declension: Yes, like that! To me it's easier to read (easier still if a variable is extracted, or maybe if f-strings work with gettext, but not sure)

>Your second point: AlbumListCountItem inherits from AlbumListItem and overrides format_label. So pattern is included to match the method signature from the base class, but the "All albums" entry doesn't actually use it.

Ah right, yes that one doesn't need the pattern at all, thanks

quodlibet/browsers/covergrid/models.py
declension: That's the one!

slosd:
Thanks for the feedback @declension. I will look at performance and type hints.

slosd:
Performance comparison with a library of 1209 albums:

Method A: I run `time python ./quodlibet.py` and click the button to close the window as soon as it appears. 

Results for https://github.com/quodlibet/quodlibet/commit/8910c14cdcb59dd6079f4db62bd8d3ceaa5de1ec:

1. python ./quodlibet.py  15,18s user 0,12s system 100% cpu 15,203 total
2. python ./quodlibet.py  14,96s user 0,14s system 100% cpu 14,993 total
3. python ./quodlibet.py  14,96s user 0,15s system 100% cpu 15,045 total

Results for https://github.com/slosd/quodlibet/commit/901c1ed48199ce321f3e0726d29572302cfaa700:

1. python ./quodlibet.py  4,35s user 0,12s system 103% cpu 4,340 total
2. python ./quodlibet.py  4,37s user 0,13s system 103% cpu 4,357 total
3. python ./quodlibet.py  4,49s user 0,17s system 97% cpu 4,765 total

Method B: I run `time python ./quodlibet.py`, focus the covergrid, hold Page-Down until I see the last page with fully loaded album covers, then click the button to close the window. 

Results for https://github.com/quodlibet/quodlibet/commit/8910c14cdcb59dd6079f4db62bd8d3ceaa5de1ec:

1. python ./quodlibet.py  42,44s user 0,31s system 100% cpu 42,562 total
2. python ./quodlibet.py  51,63s user 0,30s system 100% cpu 51,868 total
3. python ./quodlibet.py  49,05s user 0,38s system 99% cpu 49,520 total

Results for https://github.com/slosd/quodlibet/commit/901c1ed48199ce321f3e0726d29572302cfaa700:

1. python ./quodlibet.py  14,69s user 0,63s system 110% cpu 13,819 total
4. python ./quodlibet.py  15,02s user 0,56s system 134% cpu 11,562 total
5. python ./quodlibet.py  14,71s user 0,65s system 132% cpu 11,561 total

slosd:
I think I addressed all issues. One bug turned up in the meantime which is fixed with https://github.com/quodlibet/quodlibet/pull/4101/commits/1fc1582325ec06dd73d2efb4ee03fe72070ad734, so we are reminded that this probably still has some bugs which we'll eventually discover ;)

slosd:
Just noticed #2181 which this PR can probably fix too.

declension:
Thanks @slosd !

everdred:
I'm not sure how to quantify my library (definitely thousand of albums, virtually all with 600px or above cover art) or what other information might help diagnose this (please let me know how I can help), but the predictable scrolling lag I reported last year is definitely still present using the new version from git. Here's how I [described it last year](https://github.com/quodlibet/quodlibet/issues/2840#issuecomment-883804774):

> This is weird but it seems like the smoothness of the scrolling I get in the cover view can vary wildly... however, always in a completely predictable way. For some reason scrolling stutters more at the top of my library than the bottom, and while scrolling from top to bottom I can actually notice the "FPS" of the scrolling increase the further down I scroll. The same thing happens when scrolling from the bottom to the top (just reversed; this time performance decreases as I scroll), so it doesn't seem related to how long I've been scrolling continuously.

> I thought about the possibility of cover loading being the issue, so I took care to "preload" covers by PgDown'ing through the entire library before assessing the performance. With all the covers loaded I'm still seeing a progressively better/worse performance effect depending on location.

With all the covers loaded, scrolling near the top of my library gets me ~2 FPS, while scrolling near the bottom is as smooth as the UI in the rest of the application.

everdred:
Without digging too deep, there are a few more issues I've come across.

- The "wide mode" preference has the opposite effect of the label (wide mode is enabled when unchecked).
- The "show album text" preference can get out of sync with the checkbox. For instance, if the preference is checked, unchecking it makes the labels _disappear_ until you switch to another view and then back. At that point, labels reappear even though "show album text" can be seen to still be unchecked.
- I'm newly seeing an extreme amount of padding around album covers. (I thought this might be related to my custom "Album display" text, but even when replaced with "test" and the application restarted, the issue reproduces. The amount of padding seems to scale with the with the "Cover magnification" setting, but in the screenshot below it has the effect of only fitting two covers per row where four used to fit before.

2022-08-18_20-25-29_spectacle

slosd:
* Wide mode splits the browser vertically. In your screenshot that seems to match the checkbox.
* I can reproduce the issue with the album text, thanks for pointing that out.
* There is definitely not supposed to be that much padding. What Gtk theme do you use?

I have not tested the grid with thousands of albums. If you activate "Show All Albums items" you can see the number of albums in there. That information could be useful to reproduce your performance problem.

slosd:
@everdred  About the padding: There's actually a setting which influences this. Can you please provide the output of `grep item_padding $HOME/.quodlibet/config` or `grep item_padding $HOME/.config/quodlibet/config`

everdred:
> Wide mode splits the browser vertically. In your screenshot that seems to match the checkbox.

Oops, of course! I was thinking of "wide" as more space for covers.

> If you activate "Show All Albums items" you can see the number of albums in there.

I thought I was going to have to dig around the Python console documentation to find out how to do this, but thank you, this is much easier. Turns out it's around 5,000!

> Can you please provide the output of one of these (whichever matches the location of your config)

For whatever reason, it looks like my config file doesn't have any of those settings.

> What Gtk theme do you use?

It's [Graphite](https://github.com/vinceliuice/Graphite-gtk-theme), but changing to other themes doesn't seem to have an effect.

slosd:
> For whatever reason, it looks like my config file doesn't have any of those settings.
> It's [Graphite](https://github.com/vinceliuice/Graphite-gtk-theme), but changing to other themes doesn't seem to have an effect.

Thanks for the info. Neither is a surprise, but it let's us eliminate some possibilities. Do you by any chance remember if you had the issue with padding when you made your [comment](https://github.com/quodlibet/quodlibet/issues/2840#issuecomment-883804774)? You probably would have mentioned that too, right?

everdred:
> Do you by any chance remember if you had the issue with padding when you made your https://github.com/quodlibet/quodlibet/issues/2840#issuecomment-883804774?

I checked the short screencast I made at the time demonstrating the performance, and the padding issue wasn't present back then. It looked identical to the old implementation and I had four covers per row visible.

everdred:
@slosd Confirming that the spacing issue is now fixed for me with the update!

everdred:
I found an odd little issue to add... it seems like doing "reload album cover" on an album from the context menu results in the cover being shrunken. I noticed it after I placed an updated cover in an album filesystem folder and did the reload, but the issue reproduces even on albums without a changed cover.

2022-08-20_15-11-31_spectacle

The effect is just temporary... switching to another view and then changing back to cover grid fixes it, though.

slosd:
@everdred Thanks for all your testing and reports. I never tested with display scaling factor > 1, which let some bugs slip through. Should be fixed as soon as the linked PR gets merged.

declension:
@everdred @slosd thanks for the testing and fixes - now that this is merged can we start new (separated) issues for further bugs or tweaks, avoid this being a zombie issue :pray:

everdred:
@slosd And I appreciate you improving my favorite way to use my favorite music player! I'll give the new update a try and open separate issues as needed @declension.

PR #4118, opened 2022-08-21
Clip images for rounded border

### LINKED ISSUE #4117: Rounded border on album art is barely noticeable with any cover that's not mostly white
Steps to reproduce
------------------
Add an album with a dark cover to the library. Find it in the album or covergrid browser.

Expected Output
---------------
Rounded border. In my case it uses a border-radius of 5px.
cover_border_a

Actual Output
-------------
Rounded border is barely noticeable.
cover_border_b

Test System
-----------

git https://github.com/quodlibet/quodlibet/commit/e9ae92e14ab883d11e648e28b52a3d0e59fe954f

Additional Information
----------------------

Here's a zoomed in view with the old border style...
border_old_zoom

...and with the new border style (https://github.com/quodlibet/quodlibet/pull/4096).
border_zoom

slosd:
The function `add_border` now first creates a wide path along the
edge of the final image to draw the border, then it creates a narrower
path in which to draw the original image. This way the border-radius is
actually applied to the original image, which was not the case before.

Check-list
----------

Fixes #4117.

The runtime -- summed over 10000 runs -- is even consistently 0.3 seconds lower, so this shouldn't impact performance.

### DIFF
--- a/quodlibet/qltk/image.py
+++ b/quodlibet/qltk/image.py
@@ -75,28 +75,32 @@ def add_border(pixbuf, color, width=1, radius=0):
     Can not fail.
     """
 
-    w, h = pixbuf.get_width(), pixbuf.get_height()
-    w += width * 2
-    h += width * 2
+    pi = math.pi
+
+    def new_border_path(ctx, w, h, r, margin=0):
+        d = r + margin
+        ctx.new_path()
+        ctx.arc(w - d, d, r, -pi / 2, 0)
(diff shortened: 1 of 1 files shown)


declension:
Thanks!

slosd:
@declension One thing that just came to my mind: This change could lead to undesired behavior with images that have transparent areas. Those areas would be filled with the border color. Though, as far as I can see `add_border_widget` is only used for album art, so that is very unlikely to come up.

PR #4119, opened 2022-08-25
Remove redundant tooltips in playlist browser

Mellthas:
With the buttons being labeled now in the playlist browser, the tooltips are redundant.


declension:
Thanks

PR #4140, opened 2022-09-18
Override focus management in covergrid

### LINKED ISSUE #4121: Issues with arrow-key navigation in new covergrid browser
Steps to reproduce
------------------
1. Open *Browse – Cover Grid*
2. Move focus to the cover grid
3. Press and hold the up or down arrow key (also right in case of wide mode) 

Expected Output
---------------
I would expect the focus to eventually land and stay on the first resp. last row of the grid, similarly to how the title list works. This allows holding an arrow key to quickly navigate the grid without the need to be precise near the start or end.

Actual Output
-------------
Focus moves to an adjacent widget after having reached the start resp. end of the grid. This is especially jarring when moving up from the first row, as focus will move to the search box and then go on to cycle through existing saved searches and open the saved-searches window, often multiple instances on top of each other (can report a separate issue for that if needed).

Test System
-----------
Current master (815b7d2039bca5145f7734c29dc9908979cc789f) on Ubuntu 20.04.

slosd:
Up, down, left and right keys move focus only within the covergrid. The tab key moves focus before or after the covergrid.

Fixes #4121

Check-list
----------

### DIFF
--- a/quodlibet/browsers/covergrid/main.py
+++ b/quodlibet/browsers/covergrid/main.py
@@ -102,6 +102,59 @@ class PreferencesButton(AlbumPreferencesButton):
         self.pack_start(button, True, True, 0)
 
 
+class CoverGridContainer(ScrolledWindow):
+    def __init__(self, fb):
+        super().__init__(
+            hscrollbar_policy=Gtk.PolicyType.NEVER,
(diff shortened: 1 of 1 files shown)


declension:
Have tried locally and confirm this fixes the issue.

declension:
Thanks!

PR #4141, opened 2022-09-20
Track List: Add shortcut Ctrl-Return to the searchbar

nemethf:
Ctrl-Return in the searchbar enqueues the complete songlist

Check-list
----------

This is a follow-up to #2745 and #2701, because after four years I'm finally starting to use the "Track List" browser.  I've now looked the rest of the browsers and I don't see how they could be extended with this feature.  This pull request copies code from the implementation of the panned browser, so it results in code duplication.  However, a substantial duplication is already present, so this PR doesn't change the status quo.  

Once again, I am not able to find a place where the change should be documented.  `docs/guide/shortcuts.rst` is the obvious target, but I just don't know where to write what.

If necessary, please, tell me how to refine the PR, and I'll do my best to update it.  Thanks again.

### DIFF
--- a/quodlibet/browsers/paned/main.py
+++ b/quodlibet/browsers/paned/main.py
@@ -146,7 +146,8 @@ class PanedBrowser(Browser, util.InstanceTracker):
         if (is_accel(event, "<Primary>Return") or
                 is_accel(event, "<Primary>KP_Enter")):
             songs = app.window.songlist.get_songs()
-            app.window.playlist.enqueue(songs)
+            limit = config.getint("browsers", "searchbar_enqueue_limit")
+            app.window.enqueue(songs, limit)
             return True
         return False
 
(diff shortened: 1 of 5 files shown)


declension:
Thanks!
My initial concern is that the track list songlist often contains _thousands_ of songs, so an accidental ctrl-enter could make for a very long queue.

Elsewhere (e.g. songlist plugins) we have protection mechanisms above a threshold, that might be an idea here - e.g. a dialog that asks _Are you *sure* you want to enqueue 5,234 files?_ or something

nemethf:
>  My initial concern is that the track list songlist often contains _thousands_ of songs, so an accidental ctrl-enter could make for a very long queue.

I've pushed a new commit that ask for confirmation if the number of songs exceed a configurable limit.

declension:
Thanks!

PR #4142, opened 2022-09-22
Fix regression of --print-query

nemethf:
"quodlibet --print-query=asdf" results in a timeout if the music collection is large and the main quodlibet instance is started with the --debug flag because currently Query is unnecessarily instantiated for every song.  This commit fixes this, which results in a performance increase even without --debug.

I think commit e40270f70 introduced the regression.

Check-list
----------

This is described above, but now I've actually looked at commit e40270f70 and it seems it consists of similar changes, so it might introduced other regressions.  Probably not, but it's hard for me to tell.

### DIFF
--- a/quodlibet/library/song.py
+++ b/quodlibet/library/song.py
@@ -81,7 +81,8 @@ class SongLibrary(Library[K, V], PicklingMixin):
 
         songs = self.values()
         if text != "":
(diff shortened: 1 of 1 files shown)


declension:
Thanks and well spotted. It'd be nice to have a test here to prove it's not happening but maybe let's merge this first

Issue #3031, opened 2018-11-04
Ability to queue options on the command iine


Phidica:
I expected to be able to run `quodlibet --set-volume=50 --seek=00:00` on the command line with a running instance open and have the volume set *and* the song restarted.

However, it seems only the first of any options on the command line sent to a running instance is executed. Further options have no effect, *unless* the first option is `--run`.

Looking at the source code in `quodlibet/quodlibet/cli.py`, I note that each option on the command line is added to `cmds_todo` with the `queue()` function. Then, each option in `cmds_todo` is put through `control()` to send it to the remote instance. However, if the first option sent this way succeeds in execution, then `exit_()` is called and no further options are run.

Since the language in the source code involves the term "queue", it seems unusual to me that the current implementation does not allow queuing of options. Is this intended? If so, must it stay this way or can it be changed to allow option queuing?

frestr:
Yeah, this seems like a bug. As you imply, `exit_()` should be called after each command has run or when one of them fails

When chaining multiple dependent commands, it's possible that some of them might not run properly because of timing issues (e.g. a play after a slow query), but I wouldn't worry too much about that.

PR #4143, opened 2022-09-25
Add CLI args: --set-pattern and --set-pattern-help

### LINKED ISSUE #1979: Make --print-queue output match the required input of --enqueue & control file
I may have mentioned before that I have a script to manage quodlibet's queue with Dropbox for me. Well, I've run into major trouble with that script. I've been trying to modify it to support queues that are longer than bash's maximum amount of arguments. Your first thought should be: why should the max args even be an issue when I can pipe it into something? The answer is, I can't, because --print-queue gives me crap that does NOT match the input syntax of commands like enqueue, like file:// and HTML character codes, and to fix that, I have to do this:
`python2 -c "import sys, urllib as ul; print ul.unquote_plus(sys.argv[1])" "$(quodlibet --print-queue | sed 's|file://||g')"`
Which, now that I've modified it to support long queues, has turned into:
`quodlibet --print-queue | xargs -P 6 -I{} -d '\n' python2 -c "import sys, urllib as ul; print ul.unquote_plus(sys.argv[1])" "{}" | sed 's|file://||g'`
Long story short, it refused to work without the -I{} option, which slows things way down and makes this way too unwieldy, which the -P 6 option does not do a good enough job of fixing.

If --print-queue would just have proper output--namely by feeding it through that python code I'm using IN the program itself before outputting--I could just do `quodlibet --print-queue` with no problem.

### LINKED ISSUE #3080: command enhancements: print-query, print-playlist
Currently the `--print-playlist` and `--print-query` commands only return the songs' paths & filenames.  The `--print-playing` command however allows a tag pattern argument and, if present, returns the matching string. Could the `--print-playlist` and `--print-query` commands be enhanced to return strings matching an optional tag pattern?

eg 
`quodlibet --print-query=performer:guitar='John McLaughlin'  '<~album~title>'`

also, a blank query with a tag pattern should also print all the existing tag values:
`quodlibet --print-query= '<genre>'`

nemethf:
Because of #3031, it's now a two-step process to get, for example, the filenames of the songs in the queue:

  $ quodlibet --set-pattern '<~filename>'
  $ quodlibet --print-queue

Closes #1979
Closes #3080

Check-list
----------

It allows the user to configure the output format of the --print-* cli commands.

It has a negligible performance impact because, for example,  it uses `Pattern('<~filename>').format(song)` instead of `song("~filename")`.  For a queue of 500 songs, the difference is lost in the noise:
```
$ # The PR version is running
$ time ./quodlibet.py --print-queue > /tmp/q

real    0m0.298s
user    0m0.249s
sys     0m0.035s
$ # Switch to the master, restart QL
$ time ./quodlibet.py --print-queue > /tmp/q2

real    0m0.320s
user    0m0.284s
sys     0m0.024s
$ wc -l /tmp/q
500 /tmp/q
$ diff /tmp/q /tmp/q2
$ # Empty the queue
$ time ./quodlibet.py --print-queue > /tmp/q3

real    0m0.307s
user    0m0.275s
sys     0m0.029s
$ cat /tmp/q3

$ 
```

In the discussion of #1979, the desire for "format=json" has been mentioned.  This is a potential TODO, but it can later be achieved by adding support for a new tag like:  `<~json>`.

I haven't updated the documentation (the *.rst files),  I wait for some feedback first.

### DIFF
--- a/quodlibet/cli.py
+++ b/quodlibet/cli.py
@@ -86,6 +86,8 @@ def process_arguments(argv):
         _("a music library and player"),
         _("[option]"))
 
+    options.add("set-pattern", help="Set template for --print-* commands")
+    options.add("set-pattern-help", help="Show help for --set-pattern")
     options.add("print-playing", help=_("Print the playing song and exit"))
     options.add("start-playing", help=_("Begin playing immediately"))
     options.add("start-hidden", help=_("Don't show any windows on start"))
@@ -239,6 +241,13 @@ def process_arguments(argv):
             queue(command, filename)
         elif command == "enqueue-files":
             queue(command, arg)
+        elif command == "set-pattern":
+            try:
+                queue(command, args[0])
+            except IndexError:
+                queue(command)
+        elif command == "set-pattern-help":
+            queue(command)
         elif command == "play-file":
             if uri_is_valid(arg) and arg.startswith("quodlibet://"):
                 # TODO: allow handling of URIs without --play-file
--- a/quodlibet/commands.py
+++ b/quodlibet/commands.py
@@ -26,6 +26,7 @@ from quodlibet.order.repeat import RepeatListForever, RepeatSongForever, OneSong
 from quodlibet.order.reorder import OrderWeighted, OrderShuffle
 
 from quodlibet.config import RATINGS
+from quodlibet.pattern import Pattern
 
 
 class CommandError(Exception):
@@ -123,6 +124,19 @@ def arg2text(arg):
         return fsn2text(arg, strict=True)
     except ValueError as e:
         raise CommandError(e)
+
+
+default_patterns = {
+    'playing': Pattern('<artist~album~tracknumber~title>'),
+    'playlist': Pattern('<~uri>'),
+    'query': Pattern('<~filename>'),
+    'queue': Pattern('<~uri>'),
+}
+print_pattern = None
+
+
+def get_pattern(kind):
+    return print_pattern or default_patterns[kind]
 
 
 registry = CommandRegistry()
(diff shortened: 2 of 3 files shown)


nemethf:
Hm. One test failed.  But it seems unrelated, doesn't it?

quodlibet/cli.py
```diff
@@ -239,6 +241,13 @@ def queue(*args):
             queue(command, filename)
         elif command == "enqueue-files":
             queue(command, arg)
+        elif command == "set-pattern":
+            try:
+                queue(command, args[0])
+            except IndexError:
+                queue(command)
+        elif command == "set-pattern-help":
```
afontenot, line 249: This isn't the approach taken by any of the other CLI commands. I would try to remove this and fit a short, useful help message in the `help` parameter for `set-pattern` itself. The more expansive explanation for how to use the option should go in the documentation.

quodlibet/cli.py
```diff
@@ -239,6 +241,13 @@ def queue(*args):
             queue(command, filename)
         elif command == "enqueue-files":
             queue(command, arg)
+        elif command == "set-pattern":
+            try:
+                queue(command, args[0])
```
afontenot, line 246: I think this should just be `arg`. This will fail if `--set-pattern` isn't the first option provided.

afontenot (commented):
Note: I don't have merge permissions for this project, just providing some feedback because I like the idea.

 * The blank `--set-pattern` to unset the pattern seems too obscure. I would follow the approach taken by `--filter` / `--unfilter` and also do an `--unset-pattern`.
 * I don't like using a `global` to store state like this. You have access to the `app` context, so maybe there would be somewhere in there appropriate for storing this data. I'm not really sure, maybe another contributor can suggest something.

I kinda think it makes more sense to be able to store a different pattern for each print output and make it part of the program's settings (and so potentially editable in the UI or via a plugin) but maybe this is making things too complicated for a feature not many people use.

declension:
Thanks for the PR. Generally easier / more inclusive to discuss ideas in the issue tracker, but that aside there seem to be some interesting ideas here 👍

What I'm less keen on is introducing statefulness into the CLI (setting and resetting things). There are very few POSIX-type tools that do this and probably with good reason. Is there a way we could just have these as arguments / flags, with sensible defaults?


e.g. `--print-queue --with-pattern='<artist~title>' or similar

nemethf:
@afontenot, thank you for your feedback.  It seems there are more substantial issues with this PR at the moment. 

>  Generally easier / more inclusive to discuss ideas in the issue tracker

I keep that in mind for the next time.
 
> What I'm less keen on is introducing statefulness into the CLI (setting and resetting things). 

@declension, I totally understand your concern, but the current CLI parsing seems quite limited.  It is not possible for a command to have two arguments, which is required for your proposal.  For example in `--print-query FamousArtist --with-pattern="<title>"`.   Can you outline how the CLI argument parsing and the remote command machinery should be enhanced to handle this case?

Alternatively, if #3031 is fixed, then one could potentially write `QL --set-pattern <title> --print-query FamousArtist` and the effect of the set-pattern command could be limited to rest of the command line.  (But this approach would still be problematic for concurrent commands.)

nemethf:
I close this in favor of #4166.

PR #4145, opened 2022-09-26
Add disjunction pattern: use first non-empty tag

afontenot:
This is a proposed new piece of tag pattern syntax for QuodLibet. Rather than having to nest conditional expressions, you can simply put all the tags you want to consider next to each other, separated by a new "or" token: ||.

For instance:

Old syntax:

    <albumartist|<albumartist>|<artist>>

New syntax:

    <<albumartist>||<artist>>

The first non-empty sub-pattern will be used.

Much more concise and readable, especially with long chains.

### DIFF
--- a/quodlibet/pattern/_pattern.py
+++ b/quodlibet/pattern/_pattern.py
@@ -24,7 +24,7 @@ from quodlibet.util.path import strip_win32_incompat_from_path, limit_path
 from quodlibet.formats._audio import decode_value, FILESYSTEM_TAGS
 
 # Token types.
-(OPEN, CLOSE, TEXT, COND, EOF) = range(5)
+(OPEN, CLOSE, TEXT, COND, EOF, DISJ) = range(6)
 
 
 class error(ValueError):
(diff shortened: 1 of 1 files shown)


afontenot:
This is a DRAFT pull request to get feedback on a proposed new piece of syntax for tag patterns. My goal is to make someone say "WOW I WANT THIS"! 😊

This PR would enable simplifying many complex tag patterns. In my multi-pane setup, the "artist" pane changed from this:

    <albumartists|<albumartists>|<albumartist|<albumartist>|<genre=\|(classical, instrumental)|<composer|<composer>|<artist>>|<artist>>>>

To this:

    <<albumartists>||<albumartist>||<genre=\|(classical, instrumental)|<composer>>||<artist>>

Which is vastly easier to read.

Limitations at the moment (i.e. why this is just a draft):

 * Tests and documentation will be updated.
 * The `PatternCompiler` code is rather ugly IMO. QL's pattern code is complicated (and undocumented) and I didn't see any obvious easier way to do this.
 * This PR adds a new piece of syntax and I think reviewers might have opinions about whether the proposed approach is optimal.

declension:
Probably a bit back-to-front but could you add an issue (or even discussion) and link this to that, so we can encourage wider feedback than just devs? 🙏

afontenot:
@declension do you think you'd merge a version of this with documentation updated and tests added? Or are there major changes you'd want to see?

declension:
@afontenot yes I've come round to it more although I'm probably not an intended user.

Just a few naming discussions perhaps so yet let's see where it goes :+1:

afontenot:
Obsoleting this in favor of https://github.com/quodlibet/quodlibet/pull/4372.

PR #4149, opened 2022-09-27
Fix crash in console when autocompleting

afontenot:
Some objects (like custom classes) don't have a defined `__repr__`. When an object like this is the default argument of a function, we currently receive an uncaught `TypeError` exception.

Instead, use `repr()` to get the string representation of the object. When `__repr__` isn't defined, it falls back to returning a reasonable representation of most objects.


declension:
Thanks

PR #4151, opened 2022-09-27
Fix crash when using covergrid + default art size

afontenot:
The default config value for browsers.cover_size is -1. The new covergrid code works fine when the config value is absent, but with this default it will crash after requesting a negative size from `GdkPixbuf.Pixbuf.new_from_file_at_size`.

### DIFF
--- a/quodlibet/browsers/covergrid/main.py
+++ b/quodlibet/browsers/covergrid/main.py
@@ -157,7 +157,10 @@ class CoverGridContainer(ScrolledWindow):
 
 def _get_cover_size():
     mag = config.getfloat("browsers", "covergrid_magnification", 3.)
-    return mag * config.getint("browsers", "cover_size", 48)
+    size = config.getint("browsers", "cover_size")
+    if size <= 0:
+        size = 48
+    return mag * size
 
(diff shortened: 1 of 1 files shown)


afontenot:
> Is this fixing a documented issue / or have you seen this crash yourself?

I found this myself. You can replicate:

    import quodlibet
    quodlibet.pattern<tab>

The auto-completion causes a crash because `quodlibet.pattern.Pattern()` has a default argument `Kind=PatternFormatter`, where `PatternFormatter` is a `class` that doesn't have a `__repr__()` function.

afontenot:
Yes, I hit this myself - immediate crash when switching to covergrid. Probably not many people have hit it because it doesn't affect any released version, it's the new covergrid that's at fault. My `config` file contains the following:

`cover_size = -1`

The existing code for this `config.getint("browsers", "cover_size", 48)` returns -1, not 48, because the value in the config file is an integer. If you look at every other place `cover_size` is used in the code, it's sanity-checked in exactly the way I'm doing it in this PR.

declension:
Thanks @afontenot

PR #4153, opened 2022-09-27
Improve efficiency of tag pattern cache

afontenot:
Currently the code clears the cache and starts over whenever it reaches its capacity. This change uses an OrderedDict in place of the existing dict based cache. It behaves the same, but guarantees insertion ordering on older platforms, so we can use the popitem method to remove old cache entries as necessary.

We also use the move_to_end method to cheaply keep recently accessed items at the front of the cache, where they are not at risk of expiry.

Benchmarking with a set of 200 simple tags and (conservatively) assuming random access shows a performance improvement of 30%.

In the worst case scenario for the original approach, the cache always expires an entry right before it is needed. In this case this commit is about 90 *times* faster.

Check-list
----------

### DIFF
--- a/quodlibet/pattern/_pattern.py
+++ b/quodlibet/pattern/_pattern.py
@@ -13,6 +13,7 @@
 
 import os
 import re
+from collections import OrderedDict
 from re import Scanner  # type: ignore
 from urllib.parse import quote_plus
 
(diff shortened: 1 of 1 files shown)


declension:
LGTM, thanks

PR #4156, opened 2022-09-28
Disable ordering when active order disappears

### LINKED ISSUE #4155: Disabling a play order plugin doesn't take effect until after restart
Steps to reproduce
------------------
1. Enable queue only plugin (as an example) and activate it. Make sure it's working as expected.
2. Disable the plugin in the settings.
3. Double click on a song.

Expected Output
---------------
The song I double click on starts playing from the songlist (and is not added to the queue).

Actual Output
-------------
The song is added to the queue exactly as if the queue-only plugin was still enabled. It is fixed after restarting.

It is also fixed if you disable the play order mode, or change it e.g. to "random". If you click on the play order menu, you will see that despite the play order being active, none of the options are selected. The queue-only play order seems to be still active in the background.

afontenot:
When you disable a plugin-based play order or repeat mode, the UI is refreshed so that it will stop showing the mode in the menu. However, this change is only visual. The mode remains active until the app is restarted or the user actively changes the mode.

I considered several ways of fixing this, but I ended up opting to just disable the play mode entirely until the user takes action. This is the only thing that makes sense, because a user who has been using "queue only" mode probably doesn't want to be switched to "shuffle" mode just because they decide to disable the plugin.

Fixes #4155.

Check-list
----------

### DIFF
--- a/quodlibet/qltk/playorder.py
+++ b/quodlibet/qltk/playorder.py
@@ -174,6 +174,8 @@ class ToggledPlayOrderMenu(Gtk.Box):
         self.__orders = values
         if self.__current not in self.orders:
             self.__current = None
+            # if current order has become invalid, deactivate
+            self.enabled = False
         self.__rebuild_menu()
 
     @property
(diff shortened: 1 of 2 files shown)


declension:
Is there any way we can add a unit test here?

afontenot:
> Thanks. Is there any way we can add a unit test here?

Done.

declension:
Thanks

PR #3275, opened 2019-10-31
Support XDG Base Dirs: Electric Boogaloo

somini:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix
 * [ ] Unit tests have been added where possible
 * [ ] I've added / updated documentation for any user-facing features.
 * [x] Performance seems to be comparable or better than current `master`

What this change is adding / fixing
-----------------------------------

First up, I'm not too familiar with Quodlibet architecture, but I'm fluent in Python and this seemed like a good first issue.

First pass on #2202 , the XDG Base Dirs: Electric Boogaloo. I followed @op3's battle plan.

This supports the legacy `~/.quodlibet` folder better than the current code(see point 1), and the new configuration is taken from the same location for post-#138, so this doesn't lose data.

At best, you will need to migrate some data from `~/.config` to `~/.local/share`:

- `~/.config/quodlibet/lists` to `~/.local/share/quodlibet/lists`
- `~/.config/quodlibet/playlists` to `~/.local/share/quodlibet/playlists`
- `~/.config/quodlibet/feeds` to `~/.local/share/quodlibet/feeds`
- `~/.config/quodlibet/stations` to `~/.local/share/quodlibet/stations`
- `~/.config/quodlibet/stations_all` to `~/.cache/quodlibet/stations_all`, but if this is not done the Online Radio prompt should kick in and redownload it

After moving these files, clear everything from `~/.config/quodlibet` except `config` and `accels` (this is a better location for this, methinks).
We can also automatically migrate this, or consider having a fallback (to the config folder), but with loud nagging of the user.

This should be mentioned on the release notes.

I haven't look around on the tests not the docs yet.

---

1) By this I mean the current code only considers `~/.quodlibet` if it can't find the XDG directories. It should be the other way around.

PR #4158, opened 2022-09-30
Use XDG directories for storing user files

### LINKED ISSUE #2202: Support XDG Base Directory Specification
Applicable to Linux, maybe Mac OS. All versions of Quod Libet affected.

The [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/latest/) (XDGBDS) proposes where a program stores its files. That is, directories for configuration (`$XDG_CONFIG_HOME`, which defaults to `$HOME/.config`), data (`$XDG_DATA_HOME`, defaults to `$HOME/.local/share`) and cache (`$XDG_CACHE_HOME`, defaults to `$HOME/.local/cache`) are specified (among others). This scheme is already supported (with various degrees of compliance) by a wide range of software and the majority of desktop environments. It is [strongly encouraged](https://wiki.debian.org/XDGBaseDirectorySpecification) to follow the spec. For reference, [here](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support) is a concise summary of the spec and an incomplete list of supporting applications.

This scheme has several advantages:

* `$HOME` is not polluted with dotdirs
* It is simple for the user to handle different kinds of files differently: Roam the configuration (maybe version-controlled), backup the data, put the cache on a fast ramdisk, …
* Users can immediately distinguish config files from data and cache

It should be fairly simple to implement this scheme in a backwards-compatible way. The current `get_user_dir()` function could be replaced by `get_{config,data,cache}_dir()`. These functions would all return the current default path if on Windows, Windows-Portable, if `$QUODLIBET_USERDIR` is set, or if `$HOME/.quodlibet` already exists. Otherwise, the appropriate `$XDG_*`-subdirectory would be returned (e.g. `get_config_dir()` would return `$XDG_CONFIG_HOME/quodlibet` if `$XDG_CONFIG_HOME` is set, otherwise `$HOME/.config/quodlibet`). A comparable standard seems to exist for Mac OS, that utilizes the `~/Library{/Preferences,,/Cache}`-directories. It could be implemented as well.

Finally, the different files would have to be saved in the correct directories, for example:

`config` → `$XDG_CONFIG_HOME`
`lists/`, `playlists/`, `songs`, `queue` → `$XDG_DATA_HOME`
`accels`, `current`, `stations_all`? → `$XDG_CACHE_HOME`

The named pipe (currently ~/.quodlibet/control) would probably be best suited in `$XDG_RUNTIME_DIR`.

afontenot:
This is a rebased version of work originally done by @somini. It works for me (caveat to be discussed in comments).

I'm submitting this as a draft PR for discussion and suggestions, before I go too much farther with fixing the remaining issues. It is intended to replace #3275.

TODO:

 * triple-check that the file migration code works correctly
 * album_pattern should be in config dir, not data dir
 * playlist_pattern should be in config dir, not data dir
 * after testing, data migration should only be an "info" level debug item
 * appinfo plugin: check appropriate CAN_ENABLE value
 * regenerate any needed translation items
 * fix documentation mentions of config location (and internal doc strings / comments)
 * where do saved queries belong? what about "renamepatterns", "searchsites"?
 * flip the switch to `migrate_legacy_data`

Fixes #2202.

Check-list
----------

### DIFF
--- a/quodlibet/__init__.py
+++ b/quodlibet/__init__.py
@@ -29,7 +29,8 @@ from ._main import get_base_dir, is_release, app, \
 
 
 _, C_, N_, ngettext, npgettext, is_release, init, init_cli, print_e, \
-    get_base_dir, print_w, print_d, get_config_dir, app, set_application_info, \
+    get_base_dir, print_w, print_d, app, set_application_info, \
     init_plugins, enable_periodic_save, run, finish_first_session, \
-    get_image_dir, is_first_session, get_build_description, \
-    get_build_version, get_config_dir, get_data_dir, get_cache_dir, get_runtime_dir
+    get_image_dir, is_first_session, \
+    get_build_description, get_build_version, \
+    get_config_dir, get_data_dir, get_cache_dir, get_runtime_dir
--- a/quodlibet/__init__.py
+++ b/quodlibet/__init__.py
@@ -21,14 +21,15 @@ install_redirect_import_hook()
 from .util.i18n import _, C_, N_, ngettext, npgettext
 from .util.dprint import print_d, print_e, print_w
 from ._init import init_cli, init
-from ._main import get_base_dir, is_release, get_user_dir, app, \
+from ._main import get_base_dir, is_release, app, \
     set_application_info, init_plugins, enable_periodic_save, run, \
     finish_first_session, get_image_dir, is_first_session, \
-    get_build_description, get_build_version, get_cache_dir
+    get_build_description, get_build_version, \
+    get_config_dir, get_data_dir, get_cache_dir, get_runtime_dir
 
 
 _, C_, N_, ngettext, npgettext, is_release, init, init_cli, print_e, \
-    get_base_dir, print_w, print_d, get_user_dir, app, set_application_info, \
+    get_base_dir, print_w, print_d, app, set_application_info, \
     init_plugins, enable_periodic_save, run, finish_first_session, \
     get_image_dir, is_first_session, get_build_description, \
-    get_build_version, get_cache_dir
+    get_build_version, get_config_dir, get_data_dir, get_cache_dir, get_runtime_dir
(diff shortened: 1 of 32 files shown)


afontenot:
OK, so the caveat. I haven't tested the migration code very well. If I understand the approach taken correctly, I'm a bit skeptical of it. It appears to run the migration code every time the program runs, and then bails out for individual source / destination directories if they don't / do exist (respectively). I think it'd probably be better to always bail out if `.config/quodlibet` exists, under the assumption that the user knows what they're doing...

Also, I could have messed something up while working on it, but I don't think the migration code moved my songs database from the config directory to the data directory.

In any case, beware if you decide to try this. I'd make a backup of you configuration. It's working fine for me now that everything has been migrated. I'll be dogfooding this while I work on the PR.

declension:
Broadly I'm keen on this (and thank you!), but also pretty nervous (especially about migration). I won't have that much time to test either, so I'm not sure what we can do here to close the gap. Any ideas @afontenot ?

(Apologies for the new conflicts, they should mostly be easy)

afontenot:
I might have time to look through the migration code over the course of the next month or so. Couple of thoughts:

 * I don't hate the idea of just getting rid of migration entirely. Plenty of programs accept config files in both `$XDG_CONFIG_HOME` and directly under `$HOME`. Most users already have a ton of these files under `$HOME` and if they can't be bothered to clean them up (I certainly can't!) then one more program with a legacy config isn't a big deal.
 * I think ideal behavior, if migration is enabled, is to check on each startup whether `$XDG_CONFIG_HOME` exists. If it does, use it. If it doesn't, check the legacy config location, and perform migration. Does that sound reasonable to you? It's a change from the current behavior of this PR.

With respect to merge conflicts, do you do merge commits in this repository, or do I need to rebase the changes?

declension:
It would be nice if this PR could be updated and tested, though that might be an uphill battle. I'm still keen to move to XDG personally...

PR #4162, opened 2022-10-08
Stop query date regex from matching > 4 digit year

### LINKED ISSUE #4161: Can't compare a numeric tag to any number greater than 4 digits in a search
Steps to reproduce
------------------
- Go to the search bar in any browser that has one
- Enter a numeric search where the RHS of a comparison operator has 5 digits or more, e.g. `#(samplerate>44100)`

Expected Output
---------------
This should work like any other numeric search, for the example given of `#(samplerate>44100)`, this should show any song with a sample rate above 44100Hz.

Actual Output
-------------
The search bar is highlighted red, indicating it thinks this query is invalid for one reason or another.

Test System
-----------

4.5.0

### Which operating system 
Xubuntu 22.04

It's not audio related 

Additional Information
----------------------
This would only affect samplerate realistically, since units for size/time fields exist and no music has been released in the year 10000 yet.

Had to poke around in private modules to find out what the parse error actually is (if there is something on quodlibet.query.Query that does it, I did not notice and I am a silly goose):
```
>>> quodlibet.query._parser.QueryParser('#(meow=12345)').StartQuery()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/quodlibet/query/_parser.py", line 96, in StartQuery
    s = self.Query(outer=True)
  File "/usr/lib/python3/dist-packages/quodlibet/query/_parser.py", line 114, in Query
    return self.Intersection(self.Numcmp)
  File "/usr/lib/python3/dist-packages/quodlibet/query/_parser.py", line 138, in Intersection
    self.expect(')')
  File "/usr/lib/python3/dist-packages/quodlibet/query/_parser.py", line 73, in expect
    raise ParseError("'{0}' expected at index {1}, but not found"
quodlibet.query._match.ParseError: ')' expected at index 11, but not found
```

afontenot:
Some queries are ambiguous between dates and numbers, e.g. "2002" could be a year or a number. In order to simplify quering parsing, Quod Libet stores these ambiguous subqueries as the "NumexprNumberOrDate" type.

When parsing a query, Quod Libet will use this type if the string matches a regex for parsing dates. However, before this commit, the regex will match the first four digits of any four or more digit number. This results in a parsing error for large numbers because of the remaining digits after the four that get matched and parsed.

Instead, use a negative lookahead to limit the year part of the date to four digits. Quod Libet will fall back to using an unambiguous number representation for large numbers.

As an alternative, I tried to allow >4 digit years for future-proofing, but a 4 digit year assumption is baked in too deeply, e.g. in Python's `strptime` function.

Fixes #4161.

Check-list
----------

### DIFF
--- a/quodlibet/query/_parser.py
+++ b/quodlibet/query/_parser.py
@@ -27,7 +27,7 @@ SINGLE_STRING = re.compile(r"([^'\\]|\\.)*")
 DOUBLE_STRING = re.compile(r'([^"\\]|\\.)*')
 MODIFIERS = re.compile(r'[cisld]*')
 TEXT = re.compile(r'[^,)]+')
-DATE = re.compile(r'\d{4}(-\d{1,2}(-\d{1,2})?)?')
+DATE = re.compile(r'\d{4}(?!\d)(-\d{1,2}(-\d{1,2})?)?')
 
 
 class QueryParser:
--- a/tests/test_query.py
+++ b/tests/test_query.py
@@ -194,7 +194,7 @@ class TQuery(TestCase):
              "version": u"cake mix",
              "~filename": fsnative(u"/dir1/foobar.ogg"),
              "~#length": 224, "~#skipcount": 13, "~#playcount": 24,
-             "date": u"2007-05-24"})
+             "date": u"2007-05-24", "~#samplerate": 44100})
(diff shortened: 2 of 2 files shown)


declension:
Thanks

PR #4165, opened 2022-10-09
Don't insert blank tag when doing column sort magic

### LINKED ISSUE #4164: Error when Disc column is hidden and a tag pattern is used instead
Steps to reproduce
------------------
- Add a new column: `<~#discs|<~#disc> of <~#discs>|<~#disc>>`
- Move this column to the leftmost position
- Hide the Disc column
- Hide the track number column as well?

Expected Output
---------------
No error message. If it turns out this tag pattern is invalid, I would expect it to tell me that

Actual Output
-------------
The column does display something like "1 of 1" (or just 1 for files with the disc total tag not set), but an error message appears every time a new song starts playing (related to #3957):
```
NameError: name 'v0' is not defined
------

  File "/usr/lib/python3/dist-packages/quodlibet/qltk/quodlibetwindow.py", line 1338, in __browser_cb
    self.songlist.set_songs(songs, sorted)

  File "/usr/lib/python3/dist-packages/quodlibet/qltk/songlist.py", line 830, in set_songs
    default = self.find_default_sort_column()

  File "/usr/lib/python3/dist-packages/quodlibet/qltk/songlist.py", line 523, in find_default_sort_column
    if not get_sort_tag(c.header_name):

  File "/usr/lib/python3/dist-packages/quodlibet/qltk/songlist.py", line 170, in get_sort_tag
    tag = Pattern(tag).format

  File "/usr/lib/python3/dist-packages/quodlibet/pattern/_pattern.py", line 371, in Pattern
    cache[(Kind, string)] = Kind(func, list_func, tags)

  File "/usr/lib/python3/dist-packages/quodlibet/pattern/_pattern.py", line 193, in __init__
    self.format(self.Dummy())  # Validate string

  File "/usr/lib/python3/dist-packages/quodlibet/pattern/_pattern.py", line 245, in format
    value = u"".join(self.__func(self.SongProxy(song, self._format)))

  File "<string>", line 6, in f

NameError: name 'v0' is not defined
```

Test System
-----------

4.5.0

### Which operating system 
Xubuntu 22.04

Prooobably not audio related? But gstreamer 1.20.3

Additional Information
----------------------
This one's weird and hopefully makes sense. I was just trying to have a "Disc X of Y" column, guess I could have used `~#disc~~#discs~` but I was being a fussy little bastard about the separator, which doesn't seem to be changeable from what I can tell. Tried to find other conditions where this happens other than this one specific thing, but no luck.

afontenot:
If a user has a column with a tag pattern, and tries to sort by this column, Quod Libet will try to make sorting work intuitively by mangling the pattern and removing or replacing tags.

For example, `~#track` tags in patterns will be removed, on the assumption that the user will prefer the default ordering, which already looks at track numbers.

This commit fixes an issue where this is done incorrectly, and a tag like `<~#track>` is replaced with an empty tag `<>`, producing an exception when sorting by the column.

It also adds several missing tags so that magic sorting will also work for them.

Fixes #4164.

Check-list
----------

### DIFF
--- a/quodlibet/qltk/songlist.py
+++ b/quodlibet/qltk/songlist.py
@@ -163,7 +163,7 @@ def get_sort_tag(tag):
 
     if "<" in tag:
         for key, value in replace_order.items():
-            if value != "":
+            if value:
                 value = "<%s>" % value
             tag = tag.replace("<%s>" % key, value)
         for key, value in TAG_TO_SORT.items():
--- a/quodlibet/qltk/songlist.py
+++ b/quodlibet/qltk/songlist.py
@@ -151,6 +151,8 @@ def get_sort_tag(tag):
     replace_order = {
         "~#track": "",
         "~#disc": "",
+        "~#tracks": "",
+        "~#discs": "",
         "~length": "~#length"
     }
 
@@ -161,7 +163,9 @@ def get_sort_tag(tag):
 
     if "<" in tag:
         for key, value in replace_order.items():
-            tag = tag.replace("<%s>" % key, "<%s>" % value)
+            if value != "":
+                value = "<%s>" % value
+            tag = tag.replace("<%s>" % key, value)
         for key, value in TAG_TO_SORT.items():
             tag = tag.replace("<%s>" % key,
                               "<{1}|<{1}>|<{0}>>".format(key, value))
--- a/quodlibet/qltk/songlist.py
+++ b/quodlibet/qltk/songlist.py
@@ -165,7 +165,7 @@ def get_sort_tag(tag):
         for key, value in replace_order.items():
             if value:
                 value = f"<{value}>"
-            tag = tag.replace("<%s>" % key, value)
+            tag = tag.replace(f"<{key}>", value)
         for key, value in TAG_TO_SORT.items():
             tag = tag.replace("<%s>" % key,
                               "<{1}|<{1}>|<{0}>>".format(key, value))
--- a/quodlibet/qltk/songlist.py
+++ b/quodlibet/qltk/songlist.py
@@ -164,7 +164,7 @@ def get_sort_tag(tag):
     if "<" in tag:
         for key, value in replace_order.items():
             if value:
-                value = "<%s>" % value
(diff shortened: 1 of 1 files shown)


afontenot:
This could potentially get unit tested, but the code I'm replacing doesn't seem to have been tested at all. I started by ripping it out entirely to see what it did and didn't get a single test failure.

quodlibet/qltk/songlist.py
```diff
@@ -161,7 +163,9 @@ def get_sort_tag(tag):
 
     if "<" in tag:
         for key, value in replace_order.items():
-            tag = tag.replace("<%s>" % key, "<%s>" % value)
+            if value != "":
```
declension, line 166:
```suggestion
            if value:
```

quodlibet/qltk/songlist.py
```diff
@@ -161,7 +163,9 @@ def get_sort_tag(tag):
...
+                value = "<%s>" % value
```
declension, line 167:
```suggestion
                value = f"<{value}>"
```
Trying to convert to f-strings slowly everywhere

quodlibet/qltk/songlist.py
```diff
@@ -161,7 +163,9 @@ def get_sort_tag(tag):
...
+            tag = tag.replace("<%s>" % key, value)
```
declension, line 168:
```suggestion
            tag = tag.replace(f">{key}>", value)
```

afontenot:
Thanks for the suggestions, I had it in my head that f-strings were blocked by needing Python 3.5 support on some old platform, bt I guess not.

declension:
> Thanks for the suggestions, I had it in my head that f-strings were blocked by needing Python 3.5 support on some old platform, bt I guess not.

NP, and anyway we're on [Python 3.8+ now](https://github.com/quodlibet/quodlibet/blob/418b2e43800a7200e0068e68d1ed86c2dfb445e0/quodlibet/const.py#L57)

declension:
Thanks

PR #4166, opened 2022-10-11
Add CLI arg: --with-pattern

### LINKED ISSUE #1979: Make --print-queue output match the required input of --enqueue & control file

### LINKED ISSUE #3080: command enhancements: print-query, print-playlist

nemethf:
Closes #1979
Closes #3080

Check-list
----------

This is the second iteration of #4143.  It addresses the reviews of the original pull requests.   Performance-wise it should be the same.  I've also extended the documentation.   Caveat: I wrote the shell completion parts in a cargo-cult style and haven't tested them.  Also, I don't really understand the senf library, but followed existing usage.

### DIFF
--- a/quodlibet/commands.py
+++ b/quodlibet/commands.py
@@ -7,6 +7,7 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import json
 import os
 
 from senf import uri2fsn, fsnative, fsn2text, text2fsn
@@ -24,6 +25,7 @@ from quodlibet.util.library import scan_library
 
 from quodlibet.order.repeat import RepeatListForever, RepeatSongForever, OneSong
 from quodlibet.order.reorder import OrderWeighted, OrderShuffle
+from quodlibet.pattern import Pattern
 
 from quodlibet.config import RATINGS
 
@@ -123,6 +125,12 @@ def arg2text(arg):
         return fsn2text(arg, strict=True)
     except ValueError as e:
         raise CommandError(e)
+
+
+def make_pattern(fstring, default):
+    if fstring is None:
+        return Pattern(default)
+    return Pattern(arg2text(fstring))
 
 
 registry = CommandRegistry()
@@ -501,22 +509,24 @@ def _queue(app, value):
     window.qexpander.set_property('visible', value)
 
 
-@registry.register("dump-playlist")
-def _dump_playlist(app):
+@registry.register("dump-playlist", optional=1)
+def _dump_playlist(app, fstring=None):
+    pattern = make_pattern(fstring, '<~uri>')
     window = app.window
-    uris = []
+    items = []
     for song in window.playlist.pl.get():
-        uris.append(song("~uri"))
-    return text2fsn(u"\n".join(uris) + u"\n")
+        items.append(pattern.format(song))
+    return text2fsn(u"\n".join(items) + u"\n")
 
 
-@registry.register("dump-queue")
-def _dump_queue(app):
+@registry.register("dump-queue", optional=1)
+def _dump_queue(app, fstring=None):
+    pattern = make_pattern(fstring, '<~uri>')
     window = app.window
-    uris = []
+    items = []
     for song in window.playlist.q.get():
-        uris.append(song("~uri"))
-    return text2fsn(u"\n".join(uris) + u"\n")
+        items.append(pattern.format(song))
+    return text2fsn(u"\n".join(items) + u"\n")
 
 
 @registry.register("refresh")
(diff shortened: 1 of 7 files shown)


nemethf:
Once again, I have no idea what msys2 test failure means.  Hopefully, it's the tests and not the pull request.

afontenot:
This looks good to me, with one exception. I don't really love the `extract_args_from_string` approach. Most of the command running code is already built to support multiple arguments. I think it would be nicer to try to get Quod Libet to pass arbitrary data. AFAIK the only reason this hasn't been done is that until now we haven't needed to pass multiple arguments for anything.

Probably the best way to do this would be to serialize the command and arguments as JSON, e.g.

    {"command": "some_command", "args": ["value 1", "value 2"], "kwargs": {"named_arg1": "value 3"}

On the other hand, I'm not against merging this as is and going back to improve it later.

> Once again, I have no idea what msys2 test failure means. Hopefully, it's the tests and not the pull request.

MSYS is Windows. It's not clear to me what the cause of the failure is.

nemethf:
> This looks good to me, with one exception. I don't really love the `extract_args_from_string` approach.  [...] 
> Probably the best way to do this would be to serialize the command and arguments as JSON, [...]

I force-pushed a new version that implements json encoding for the arguments of the print-query command and I opened a new issue (#4170) to discuss how the general argument handling could be changed.

afontenot:
I think this looks quite nice! Solves the problem while leaving room for future progress, e.g. by switching all the communication to serialized JSON or going with one of your ideas discussed in #4170.

One question - do you need the backwards compatibility stuff? Is it necessary that an old version of Quod Libet be able to send arguments to a newer version? I'm not saying it isn't, I really don't know what the requirements are for this...

nemethf:
> do you need the backwards compatibility stuff? 

I don't.  I just wasn't sure whether it is necessary to provide it.

tests/test_commands.py
```diff
@@ -109,3 +109,24 @@ def test_enqueue_files(self):
         self.__send("enqueue-files "
                     "one,two\\, please,slash\\\\.mp3,four")
         self.assertEquals(app.window.playlist.q.get(), songs)
+
+    def test_with_pattern(self):
```
declension, line 113: I think it would be good to test the case of some valid JSON that's also a query

`"foo"`
(or even `1.0` etc)

nemethf: You're right.  This actually revealed a bug in the PR.

quodlibet/commands.py
```diff
@@ -525,14 +535,22 @@ def _refresh(app):
...
     """Queries library, dumping filenames of matches to stdout
     See Issue 716
     """
 
-    query = arg2text(query)
+    try:
+        args = json.loads(arg2text(json_encoded_args))
+        query = args['query']
+        fstring = args['pattern']
+    except (json.decoder.JSONDecodeError, IndexError):
```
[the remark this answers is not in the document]
nemethf: Yeah, you're right

nemethf:
@declension, thank your for the review.  I pushed a new version addressing your concerns.

tests/test_commands.py
```diff
@@ -109,3 +109,46 @@ def test_enqueue_files(self):
...
+        songs = [AudioFile({"~filename": fn, "title": fn.upper()})
+                 for fn in ["one", "two, please", "slash\\.mp3", "4.0-four"]]
+        app.library.add(songs)
+
+        self.assertFalse(app.window.playlist.q.get())
+        self.__send("enqueue-files "
+                    "one,two\\, please,slash\\\\.mp3,4.0-four")
+        # Old syntax
+        assert self.__send("print-query two ") == "two, please\n"
+        # Old syntax that is also a valid json
```
declension, line 123: This is all much better :+1:

Could you split the assertions up into tests? Motivation: it's way easier to debug when a test called `test_old_syntax_that_is_valid_json()` starts failing, than "line 123 of _test_everything()`" etc

The comments pretty much become the test name, and anything shared is a fixture (ideally) or here perhaps just in `setUp()`

nemethf: I refactored these tests into a new class as requested.  (I think the resulting code is a bit complex.)  Thanks.

declension: Great, thanks!

declension:
Thanks @nemethf

PR #4172, opened 2022-10-18
Make --rating-up work when song has no previous rating

nemethf:
Now assume songs with no ratings have the default rating.  (Also fix other relative rating changes.)

Check-list
----------

`QL --rating-up` fails to change the rating of a song if the song does not currently have a rating.  Unfortunately, I once again forgot to create a bug report first, but the fix is very straightforward.  Thanks.

### DIFF
--- a/quodlibet/commands.py
+++ b/quodlibet/commands.py
@@ -323,7 +323,7 @@ def _rating(app, value):
             change = (1 / RATINGS.number)
         if value[0] == '-':
             change = -change
-        rating = song["~#rating"] + change
+        rating = song("~#rating") + change
     else:
         try:
             rating = float(value)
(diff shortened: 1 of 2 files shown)


quodlibet/commands.py
```diff
@@ -323,7 +323,7 @@ def _rating(app, value):
             change = (1 / RATINGS.number)
         if value[0] == '-':
             change = -change
-        rating = song["~#rating"] + change
+        rating = song.get("~#rating", RATINGS.default) + change
```
declension, line 326: I think `song("~#rating")` should work here and not duplicate that logic

nemethf: Thanks.  I force pushed a new version

declension:
Thanks

PR #4175, opened 2022-10-20
Modernise entire preferences UI :art:

declension:
* Convert to Gtk.Switch everywhere. It's the one true way now.
 * Rework the player prefs too
 * Make prefs window a bit more resize-friendly
 * Simplify some UI elements
 * Use more Gtk.Scale (for replaygain)
 * Integer RG gain steps only
 * Use FlowBox for column headings
 * Add a few icons for consistency
 * Fix a few tooltips etc
 * Add a behaviour frame for Player
 * Nicer margins

### Song List Prefs (wide) 
image

### Playback Prefs (normal)
image

### DIFF
--- a/quodlibet/player/gstbe/prefs.py
+++ b/quodlibet/player/gstbe/prefs.py
@@ -42,7 +42,8 @@ class GstPlayerPreferences(Gtk.VBox):
         apply_button = Button(_("_Apply"), Icons.VIEW_REFRESH)
 
         def format_buffer(scale, value):
-            return _("%.1f seconds") % value
+            # Translators: s = seconds
+            return _("%.1f s") % value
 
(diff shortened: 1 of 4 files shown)


jej:
I tried to put the replay gain sliders in a table to align them. I don't know if it's the right way to do that... It's aligned but not perfect anyway...
https://github.com/jej/quodlibet/commit/cf946e818feb289fa39ba19b3bc5d03d2bff7618

afontenot:
Before:
ql_before

After:
ql_after

I think I like the original tab style better, but the rest of it looks nice.

`Gtk.Switch` is a bit too big in my Gtk theme, but that's okay.

jej:
Great!

PR #4195, opened 2022-10-30
Add a new internal tag: ~json

### LINKED ISSUE #4191: json output format for --print-* CLI commands
The proposed change
-------------------

In #1979, @declension briefly contemplated whether it would make sense to have a json output format for --print-* CLI commands.  Since QL now has the --with-pattern option, implementing this is quite easy:

```diff --git a/quodlibet/formats/_audio.py b/quodlibet/formats/_audio.py
index 26895b8f7..d294f5272 100644
--- a/quodlibet/formats/_audio.py
+++ b/quodlibet/formats/_audio.py
@@ -517,6 +517,8 @@ class AudioFile(dict, ImageContainer, HasKey):
                             return float(val)
                         except ValueError:
                             return default
+            elif key == "json":
+                return json.dumps(self)
             else:
                 return dict.get(self, "~" + key, default)
 
```
So, one can, for example, execute `quodlibet --print-queue --with-pattern=<~json>`

Benefits of this change
-----------------------

If the output is intended to be post processed by a script, then it is better to use the json format than to use and ad-hoc separator like in `quodlibet --print-queue --with-patern=<artist>/<title>` because a field might contain the separator.  "AC/DC" in this case.

Any issues this might cause
---------------------------

Nothing that I can think of.  However, it might make sense to complicate the syntax a bit.  If the user is only interested in a small number of fields, it might be wasteful to encode all of them.  So maybe a `<json:artist,title>` syntax is the way to go.  What do you think?

nemethf:
Closes #4191

Check-list
----------

It adds an internal tag named ~json.  #4191 provides background information.

### DIFF
--- a/docs/guide/tags/internal_tags.rst
+++ b/docs/guide/tags/internal_tags.rst
@@ -26,6 +26,7 @@ String Tags
  * ``~originalyear``: The original year, derived from the ``originaldate`` tag
  * ``~playlists``: Comma-separated playlist names in which the song is included
  * ``~filesize``: Human formatted size (e.g. *4.5 MB*)
+ * ``~json``: A json-formatted string of all the non-internal tags and some internal tags
 
 
 The ``~people`` Internal Tag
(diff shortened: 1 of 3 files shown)


declension:
Thanks

Issue #4194, opened 2022-10-30
Move default branch to `main`


declension:
The proposed change
-------------------
Change default branch from `master` to `main`

Benefits of this change
-----------------------
This is the modern non-tainted naming, it's the Git default, [the Github default](https://github.com/github/renaming) ([and Gitlab](https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/))

Any issues this might cause
---------------------------
* CI might break until sorted out?
* Build tools themselves
* Downstream packaging things maybe?
* In-flight PRs will need fixing up

declension:
Looking at https://github.com/github/renaming#renaming-existing-branches, this looks surprisingly well automated via Github. @lazka shall we just try it out and fix what breaks?

lazka:
Sure

declension:
Updated in Github.. Now let's see (hence leaving this ticket open)

declension:
Thanks for fixing Weblate stuff @lazka

lazka:
Some more things in #4201

declension:
Done

PR #4201, opened 2022-11-01
More master -> main things

lazka:
See #4194


declension:
I guess this is good to go?

PR #4210, opened 2022-11-23
Fix description of JEP-118 plugin not being translated

Mellthas:
Plugin description appears untranslated even though translation exists.


declension:
Thanks

Issue #3969, opened 2022-04-05
[Feature Request] Add a Stop button 


krumpfwylg:
Would it be possible to add a Stop button that would stop playback & release Alsa device ? 

Quodlibet already does that when "Stop after this song" is used, it's a bit sad there's not a button to do this.

By "release Alsa device", I mean kill the connection to alsa driver such as :
Quodlibet paused/on playback/starting after a track is pause gives
```
$ fuser -vu /dev/snd/*
                     USER        PID ACCESS COMMAND
/dev/snd/pcmC0D0p:   krumpf    23048 F...m (krumpf)quodlibet
/dev/snd/timer:      krumpf    23048 f.... (krumpf)quodlibet
```

Quodlibet after using Stop function
```
$ fuser -vu /dev/snd/*
[.. no results..]
```

Thanks for reading

toupeira:
FWIW there's `quodlibet --stop`, I have that mapped to a shortcut and use it very regularly when I need to release my audio device.

But you're right, this seems to be completely missing in the UI! I submitted https://github.com/quodlibet/quodlibet/pull/4213 to at least add it to the "Control" menu, which turned out to be a very simple change.

Not sure if a button in the main UI is really needed too, and anyway I'm too unfamiliar with the Quod Libet code to implement that :sweat_smile:

PR #4213, opened 2022-11-29
Add Stop command to Control menu

toupeira:
Check-list
----------

   * Might address https://github.com/quodlibet/quodlibet/issues/3969, not sure if a button in the main UI is really needed.
 * [-] Unit tests have been added where possible
   * `tests/test_qltk_quodlibetwindow.py` is currently very barebones, so I figured I'll skip this :wink: 
   * Tested manually, seems to work fine.
 * [-] I've added / updated documentation for any user-facing features.
   * Did a quick search but couldn't find any description of the Control menu.

This adds a menu item to fully stop playback, which is currently only available through the CLI, and some other integrations like MPRIS.

**Screenshot:**

image

### DIFF
--- a/quodlibet/qltk/quodlibetwindow.py
+++ b/quodlibet/qltk/quodlibetwindow.py
@@ -424,6 +424,7 @@ MENU = """
       <menuitem action='Previous' always-show-image='true'/>
       <menuitem action='PlayPause' always-show-image='true'/>
       <menuitem action='Next' always-show-image='true'/>
+      <menuitem action='Stop' always-show-image='true'/>
       <menuitem action='StopAfter' always-show-image='true'/>
     </menu>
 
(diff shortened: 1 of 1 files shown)


quodlibet/qltk/quodlibetwindow.py
```diff
@@ -957,6 +958,11 @@ def add_top_level_items(ag):
         act.connect('activate', self.__next_song)
         ag.add_action_with_accel(act, "<Primary>period")
 
+        act = Action(name="Stop", label=_("Stop"),
+                     icon_name=Icons.MEDIA_PLAYBACK_STOP)
+        act.connect('activate', self.__stop)
+        ag.add_action(act)
```
toupeira, line 964: **question:** Should we add a keyboard shortcut?

Some suggestions: Ctrl+Shift+Space, Ctrl+Escape

declension: Maybe yes. Not Ctrl+Escape I'd say, that's too weird and might do things in some WMs

declension:
Generally I think this is worth a try. If people don't want it they won't have to use it

declension:
Thanks

PR #4219, opened 2022-12-15
Fix regression when calling "Edit saved searches" window

jej:
Check-list
----------

Fix a regression bug introduced in #4188

To reproduce:
- open QL
- in searchbar: "Edit saved searches..."

Returns:
```
TypeError: GObject.set_property() takes exactly 2 arguments (3 given)
------

  File "/usr/local/lib/python3.9/dist-packages/quodlibet/qltk/cbes.py", line 143, in __cdf
    cell.set_property("markup",

TypeError: GObject.set_property() takes exactly 2 arguments (3 given)
```


declension:
Thanks!

PR #4220, opened 2022-12-16
Fix xgettext warning about 'Invalid tag' plural

jej:
Check-list
----------

During build I get this warning:
```
/home/user/src/quodlibet/gdist/gettextutil.py:164: GettextWarning: xgettext: WARNING : msgid 'Invalid tag' is used without plural and with plural.
                          quodlibet/qltk/edittags.py:798: Here is the occurrence without plural.
                          quodlibet/qltk/tagsfrompath.py:211: Here is the occurrence with plural.
                          Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
xgettext: WARNING : msgid 'Invalid tag %s
                          
                          The files currently selected do not support editing this tag.' is used without plural and with plural.
                          quodlibet/qltk/edittags.py:799: Here is the occurrence without plural.
                          quodlibet/qltk/tagsfrompath.py:212: Here is the occurrence with plural.
                          Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
xgettext: WARNING : msgid 'Invalid tag' is used without plural and with plural.
                          quodlibet/util/__init__.py:511: Here is the occurrence without plural.
                          quodlibet/qltk/edittags.py:798: Here is the occurrence with plural.
                          Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.
xgettext: WARNING : msgid 'Invalid tag' is used without plural and with plural.
                          quodlibet/util/tags.py:244: Here is the occurrence without plural.
                          quodlibet/qltk/edittags.py:798: Here is the occurrence with plural.
                          Workaround: If the msgid is a sentence, change the wording of the sentence; otherwise, use contexts for disambiguation.

  warnings.warn(stderr, GettextWarning)
```

This patch fixes it with a pluralized string forced to 1 element (maybe not on the cleanest way, check it...)

### DIFF
--- a/quodlibet/qltk/edittags.py
+++ b/quodlibet/qltk/edittags.py
@@ -796,8 +796,8 @@ class EditTags(Gtk.VBox):
             assert isinstance(value, str)
             if not self._group_info.can_change(tag):
                 title = ngettext("Invalid tag", "Invalid tags", 1)
-                msg = ngettext("Invalid tag %s\n\nThe files currently"
-                        " selected do not support editing this tag.",
+                msg = ngettext("Invalid tag %s\n\nThe files currently "
+                        "selected do not support editing this tag.",
                         "Invalid tags %s\n\nThe files currently "
                         "selected do not support editing these tags.", 1
                         ) % util.bold(tag)
@@ -979,8 +979,8 @@ class EditTags(Gtk.VBox):
         elif not self._group_info.can_change(new_tag):
             # Can't add the new tag.
             title = ngettext("Invalid tag", "Invalid tags", 1)
-            msg = ngettext("Invalid tag %s\n\nThe files currently"
-                    " selected do not support editing this tag.",
+            msg = ngettext("Invalid tag %s\n\nThe files currently "
+                    "selected do not support editing this tag.",
                     "Invalid tags %s\n\nThe files currently "
                     "selected do not support editing these tags.", 1
                     ) % util.bold(new_tag)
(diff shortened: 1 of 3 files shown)


quodlibet/qltk/edittags.py
```diff
@@ -795,9 +795,11 @@ def __add_tag(self, activator, model, library):
             value = massagers.validate(tag, value)
             assert isinstance(value, str)
             if not self._group_info.can_change(tag):
-                title = _("Invalid tag")
-                msg = _("Invalid tag %s\n\nThe files currently"
-                        " selected do not support editing this tag."
+                title = ngettext("Invalid tag", "Invalid tags", 1)
+                msg = ngettext("Invalid tag %s\n\nThe files currently "
```
nemethf, line 799: I think `_(...)` is for internationalization.  It shouldn't be removed.

nemethf: I take it back. It's OK. Sorry

declension:
Thanks. I think there must be a cleaner way - to me it's the warning that seems wrong here not the code...

jej:
I agree that it's a dirty fix :)

declension:
OK seems like an improvement still, thanks

jej:
I am surprised by the merge... Really dirty :)
I didn't find much information about the `otherwise, use contexts for disambiguation`. Maybe it's the right way to manage that.

PR #4225, opened 2022-12-20
Update Norwegian bokmål (nb) translation

koratfood:
Check-list
----------

Complete translation to Norwegian bokmål (nb) as of current POT.

### DIFF
--- a/po/nb.po
+++ b/po/nb.po
@@ -1,17 +1,19 @@
 # Norwegian bokmål translation of Quod Libet
-# Copyright (C) 2004-2017 Joe Wreschnig and other Quod Libet authors
+# Copyright (C) 2004-2022 Joe Wreschnig and other Quod Libet authors
 # This file is distributed under the same license as the Quod Libet package.
 #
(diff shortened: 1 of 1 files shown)


declension:
FYI we now use [weblate](https://hosted.weblate.org/projects/quodlibet/quodlibet/) for handling all translations

koratfood:
From [docs --> Translation Guide --> How-to (Legacy Manual Process)](https://quodlibet.readthedocs.io/en/latest/translation/howto_manual.html):

> The new preferred method to translate is via [Weblate](https://hosted.weblate.org/engage/quodlibet). The process described here still works and is supported in case you for some reason prefer it.

Are you saying the latter sentence in that quote no longer applies?

declension:
No I _think_ it's still all fine, just noting that nobody has used it now for some time,  so can't guaranteed that.

Probably @lazka knows better

lazka:
It should still work. There is a higher chance of conflicts if updates happen in both places at the same time, but that doesn't seem to be the case here.

declension:
Cool I guess we can merge this as-is then?

Issue #4232, opened 2022-12-26
Album List sidebar header lost a newline


elcste:
A newline was lost with [recent changes to quodlibet/browsers/albums/main.py](https://github.com/quodlibet/quodlibet/commit/2f516a2cf88b67397da8fbed38cabbc35e60548a#diff-f1827e01aed34ab8614828e1c4fa86cd252731388f5b39baccff0ffa85894533)

In the current git version, without the newline, the sidebar displays wrongly as

All Albums1 album

In version 4.5, with the newline, it correctly uses two lines:

All Albums
1 album

declension:
Sorry - my fault. Closed by #4233

PR #4233, opened 2022-12-26
Re-add newline in Album List sidebar header

elcste:
Check-list
----------

 * bug: https://github.com/quodlibet/quodlibet/issues/4232
 * [n/a] Unit tests have been added where possible
 * [n/a ] I've added / updated documentation for any user-facing features.
 * [ x] Performance seems to be comparable or better than current `main`

The newline was lost with the changes in the last commit. Without the newline the sidebar displays as

All Albums1 album

instead of on two lines

All Albums
1 album

***

I know only enough python to be dangerous. This fixes it for me in my local version on Ubuntu 22.10. But there may be a better way to do this!

### DIFF
--- a/quodlibet/browsers/albums/main.py
+++ b/quodlibet/browsers/albums/main.py
@@ -543,7 +543,7 @@ class AlbumList(Browser, util.InstanceTracker, VisibleUpdate,
             album = model.get_album(iter_)
 
             if album is None:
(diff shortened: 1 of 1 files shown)


declension:
Oh, my bad, I missed that.

declension:
Thanks!

PR #4236, opened 2022-12-27
playlist export: prefix relative paths with '.' if the path starts with '#'

### LINKED ISSUE #4235: Playlist export plugin: M3U playlists with relative paths starting with "#" are invalid
Lines starting with "#" are treated specially by most (all?) M3U parsers as control lines or comments. This means that if I export a playlist with relative paths with a directory or file name starting with "#" such that that path component is at the start of the relative path, most M3U implementations will ignore those files because they think those lines are control data or comments.

For example, consider this hastily rendered directory structure:

```
+ parent_dir
|   + #mysubdir
|   |   somefile.mp3
|   playlist.m3u 
```

So if I export a playlist containing "somefile.mp3" using the playlist export plugin and save it as an M3U file at "playlist.m3u", the playlist will contain "#mysubdir/somefile.mp3" which most M3U parsers (Gnome Videos and the LineageOS music player at least) will ignore.

fkrull:
Check-list
----------

 * [ ] (n/a) I've added / updated documentation for any user-facing features.

Fixes #4235, by prefixing the path with "./" if it detects that a relative path starts with "#". Always adding the prefix would probably work as well, but I opted for the least intrusive change.

### DIFF
--- a/quodlibet/ext/songsmenu/playlist.py
+++ b/quodlibet/ext/songsmenu/playlist.py
@@ -1,5 +1,6 @@
 # Copyright 2009 Christoph Reiter
 #      2014-2020 Nick Boultbee
+#      2022 Felix Krull
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -23,28 +24,32 @@ from quodlibet.qltk import Icons
 from quodlibet.plugins.songsmenu import SongsMenuPlugin
 
 
+FORMAT_M3U = "m3u"
+FORMAT_PLS = "pls"
+
+
 lastfolder = get_home_dir()
 
 
 class PlaylistExport(PlaylistPlugin, SongsMenuPlugin):
-    PLUGIN_ID = 'Playlist Export'
-    PLUGIN_NAME = _('Export as M3U / PLS Playlist File')
-    PLUGIN_DESC = _('Exports songs to an M3U or PLS playlist.')
+    PLUGIN_ID = "Playlist Export"
+    PLUGIN_NAME = _("Export as M3U / PLS Playlist File")
+    PLUGIN_DESC = _("Exports songs to an M3U or PLS playlist.")
     PLUGIN_ICON = Icons.DOCUMENT_SAVE_AS
     REQUIRES_ACTION = True
 
     lastfolder = None
 
     def plugin_single_playlist(self, playlist):
-        return self.__save_playlist(playlist.songs, playlist.name)
+        return self.__save_playlist_ui(playlist.songs, playlist.name)
 
     def plugin_songs(self, songs):
-        self.__save_playlist(songs)
+        self.__save_playlist_ui(songs)
 
-    def __save_playlist(self, songs, name=None):
-        dialog = Gtk.FileChooserDialog(self.PLUGIN_NAME,
-            None,
-            Gtk.FileChooserAction.SAVE)
+    def __save_playlist_ui(self, songs, name=None):
+        dialog = Gtk.FileChooserDialog(
+            self.PLUGIN_NAME, None, Gtk.FileChooserAction.SAVE
+        )
         dialog.set_show_hidden(False)
         dialog.set_create_folders(True)
         dialog.add_button(_("_Cancel"), Gtk.ResponseType.CANCEL)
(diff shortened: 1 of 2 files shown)


quodlibet/ext/songsmenu/playlist.py
```diff
@@ -118,6 +118,9 @@ def __get_files(self, songs, dir_path, relative=False):
                 path = song('~filename')
                 if relative:
                     path = relpath(path, dir_path)
+                    if path.startswith("#"):
+                        # avoid lines starting with '#' which don't work with M3U
+                        path = os.path.join(os.curdir, path)
```
declension, line 123: Is `os.curdir` always relative though?

declension: `os.path.join(".", path)` might work though

A test would be good here...

fkrull: It's [documented as](https://docs.python.org/3/library/os.html?highlight=os%20curdir#os.curdir) "The constant string used by the operating system to refer to the current directory. This is '.' for Windows and POSIX." That seems pretty clear-cut to me. Specifically, "constant string" suggests it can't be implemented any other way, e.g. using the process working directory.

> A test would be good here...

Yeah ok, I can do that.

declension: ok cool

fkrull: Ok, I added some tests.

tests/plugin/test_playlist.py
```diff
@@ -0,0 +1,195 @@
+# Copyright 2022 Felix Krull
```
declension, line 1: Great, thanks for adding tests!

tests/plugin/test_playlist.py
```diff
@@ -0,0 +1,195 @@
...
+                title="c",
+            )
+
+            plugin.save_playlist(
+                [song1, song2], playlist_file_path, self.mod.FORMAT_M3U, relative=True
+            )
+            with open(playlist_file_path, "r") as f:
+                result = f.read()
+
+        self.assertEqual(
```
declension, line 81: With new tests, we're generally switching over to simpler pytest asserts
e.g.
```assert result == expected_content, "Helpful failure reason"```

File: tests/plugin/test_playlist.py [hunk 4]
```diff
@@ -0,0 +1,195 @@
...
+            /a/b/c.mp3
+            #EXTINF:400,b - d
+            /a/b/d.mp3
+            """
+            ),
+        )
+
+    def test_m3u_playlist_relative(self):
+        plugin = self.mod.PlaylistExport()
+        with temp_filename() as playlist_file_path:
```
declension, line 62: FYI you can now use `as_path=True` here to get a `pathlib.path` which are IMO much nicer to work with...

tests/plugin/test_playlist.py
```diff
@@ -0,0 +1,195 @@
...
+            d = os.path.dirname(playlist_file_path)
```
declension, line 63: This can then be `playlist_file_path.parent`

tests/plugin/test_playlist.py
```diff
@@ -0,0 +1,195 @@
...
+                filename=os.path.join(pd, "c.mp3"),
+                length=1,
+                artist="a",
+                title="c",
+            )
+
+            plugin.save_playlist(
+                [song1, song2], playlist_file_path, self.mod.FORMAT_M3U, relative=True
+            )
+            with open(playlist_file_path, "r") as f:
```
declension, line 78: And this is just `playlist_file_path.read_text()` :up:

tests/plugin/test_playlist.py
```diff
@@ -0,0 +1,195 @@
...
+            ),
+        )
+
+    def test_m3u_playlist_relative(self):
+        plugin = self.mod.PlaylistExport()
+        with temp_filename() as playlist_file_path:
+            d = os.path.dirname(playlist_file_path)
+            pd = os.path.dirname(d)
+            song1 = audio_file(
+                filename=os.path.join(d, "a", "b.mp3"), length=23, artist="a", title="b"
```
declension, line 66: this becomes `str(d / "a" / "b.mp3")` in Pathlib world

declension (approved):
LGTM, see comments

tests/plugin/test_playlist.py
fkrull: :heavy_check_mark:

tests/plugin/test_playlist.py
fkrull: :heavy_check_mark:

tests/plugin/test_playlist.py
fkrull: :heavy_check_mark:

tests/plugin/test_playlist.py
fkrull: :heavy_check_mark:

declension:
Thanks!

PR #3581, opened 2021-03-10
Ignore auto_sort when shuffle is enabled

luk1337:
Check-list
----------

 * [ ] There is a linked issue discussing the motivations for this feature or bugfix
 * [x] Unit tests have been added where possible
 * [x] I've added / updated documentation for any user-facing features.
 * [x] Performance seems to be comparable or better than current `master`

What this change is adding / fixing
-----------------------------------

After #2509 has been merged the ability to go back to the previous track when shuffle mode
was enabled stopped working. This change makes it so that auto_sort is ignored when needed.

PR #4255, opened 2023-01-14
Fix shuffle history

### LINKED ISSUE #4108: Shuffle mode loses played-history when song list is reordered
Steps to reproduce
------------------

1. Enable random shuffle
2. Start a song (_song A_)
3. Press "next" button
4. Sort song list by a different column so the order of the songs change
5. Press "previous" button (a second time if playback goes to the beginning of the current song)

Expected Output
---------------

Playback goes to _song A_.

Actual Output
-------------

Playback stops. UI shows "Not playing".

Test System
-----------

Quod Libet 4.5.0

Additional Information
----------------------

Because of this issue auto-sorting was disabled while shuffle mode is active in https://github.com/quodlibet/quodlibet/pull/3581. The auto-sorting feature seems to have made this bug more obvious, but it's not the root cause. This also happens when the list is manually re-sorted.

Also related: #3592

slosd:
Check-list
----------

### Reset play order in queue for new set of songs

Used a specialized `QueueModel` because the test `test_random_queue_666` requires behavior that's slightly different from `PlaylistModel`.

### Reorder songlist in place

Instead of replacing the whole list reorder the existing one. This
allows play order implementations to continue to use any saved
TreeIters.

Fixes #4108

### Revert "Ignore auto_sort when shuffle is enabled (#3581)"

This reverts commit 68ba6786072557d05f9da99c25d14d1d8cf1d192. This limitation shouldn't be necessary anymore.

### DIFF
--- a/quodlibet/order/__init__.py
+++ b/quodlibet/order/__init__.py
@@ -6,7 +6,9 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
-from typing import Optional
+from typing import Any, Dict, List, Optional
+
+from gi.repository import Gtk
 
 from quodlibet import _, print_d
 
@@ -103,39 +105,42 @@ class OrderRemembered(Order):
     """Shared class for all the shuffle modes that keep a memory
     of their previously played songs."""
 
+    _played: List[Gtk.TreeIter]
+
     def __init__(self):
         super().__init__()
         self._played = []
 
     def next(self, playlist, iter):
         if iter is not None:
-            self._played.append(playlist.get_path(iter).get_indices()[0])
+            self._played.append(iter)
 
     def previous(self, playlist, iter):
-        try:
-            path = self._played.pop()
-        except IndexError:
-            return None
-        else:
-            return playlist.get_iter(path)
+        if self._played:
+            return self._played.pop()
+        return None
 
     def set(self, playlist, iter):
         if iter is not None:
-            self._played.append(playlist.get_path(iter).get_indices()[0])
+            self._played.append(iter)
         return iter
 
     def reset(self, playlist):
         del(self._played[:])
 
-    def remaining(self, playlist):
+    def remaining(self, playlist) -> Dict[int, Any]:
         """Gets a map of all song indices to their song from the `playlist`
         that haven't yet been played"""
-        all_indices = set(range(len(playlist)))
-        played = set(self._played)
+
+        def get_index(iter):
+            return playlist.get_path(iter).get_indices()[0]
+
+        played = set(map(get_index, self._played))
         print_d("Played %d of %d song(s)" % (len(self._played), len(playlist)))
-        remaining = list(all_indices.difference(played))
-        all_songs = playlist.get()
-        return {i: all_songs[i] for i in remaining}
+        remaining = (
+            (get_index(iter), value) for iter, value in playlist.iterrows())
+        return {
+            index: song for (index, song) in remaining if index not in played}
 
 
 class OrderInOrder(Order):
--- a/quodlibet/qltk/queue.py
+++ b/quodlibet/qltk/queue.py
@@ -9,7 +9,7 @@
 
 import os
 import time
-from typing import Optional
+from typing import Any, Optional, Sequence
 
 from gi.repository import Gtk, Gdk, Pango, GLib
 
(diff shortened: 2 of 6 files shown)


quodlibet/order/__init__.py
```diff
@@ -109,33 +109,34 @@ def __init__(self):
 
     def next(self, playlist, iter):
         if iter is not None:
-            self._played.append(playlist.get_path(iter).get_indices()[0])
+            self._played.append(iter)
```
declension, line 112: Adding a type annotation in the constructor would let readers know it's now a different type of list element which might be nice

quodlibet/qltk/queue.py
```diff
@@ -345,7 +345,35 @@ def __queue_disable(self, disabled):
...
+    """
+
+    __reset = False
+
+    def set(self, songs: Sequence[Any]):
+        self.__reset = True
+        super().set(songs)
+
+    def next(self):
+        print_d("Using %s.next_explicit() to get next song" % self.order)
```
declension, line 363:
```suggestion
        print_d(f"Using {self.order} to get next song")
```
(we're moving slowly to f-strings everywhere)

declension:
Side note: I guess we do need a strong "songlike type" than Any. `AudioFile` is good if it's actually that, but sometimes it's a string so a few times I've added a union but, I dunno maybe it's not worth it.

declension:
(I think the Windows job is failing because of a different version of `mypy`, despite efforts to unify these...

anyway it just wants an annotation on `found = []`)

declension (commented):
LGTM

quodlibet/order/__init__.py
slosd: I've been hesitant to use GObject classes in type hints, because they are not usable for the type checker. This code at the moment infers the type `List[Unknown]`. If I write `self._played: List[Gtk.TreeIter] = []` it infers the same. In fact I can write `self._played: List[Gtk.TreeeeIter] = []` and get no error.
I'll add it of course if you prefer. I guess it functions just as well as a comment would.

declension: >  I guess it functions just as well as a comment would.

Yes - this is the win IMO (also MyPy is improving and the stubs hopefully will do too)

tests/test_qltk_songmodel.py
```diff
@@ -126,6 +128,20 @@ def test_shuffle(self):
             self.assertEqual(self.pl.current, None)
             self.pl.order.reset(self.pl)
 
+    def test_shuffle_previous_after_reorder(self):
+        rand = Random(7)
+        self.pl.order = OrderShuffle()
+        history = [self.pl.current for _ in range(10) if self.pl.next() or True]
```
declension, line 134: Why the `or True` - is the `if` clause functioning as a `zip()` here?

tests/test_qltk_songmodel.py
```diff
@@ -126,6 +128,20 @@ def test_shuffle(self):
...
+        order = list(range(10))
+        for i in reversed(range(10)):
+            values = list(self.pl.itervalues())
+            rand.shuffle(order)
+            self.pl.reorder(order)
+            self.assertNotEqual(values, list(self.pl.itervalues()))
+            self.failUnlessEqual(self.pl.current, history[i],
+                    "expected different item at index %d" % i)
```
declension, line 142:
```suggestion
                   f"expected different item at index {i}")
```

tests/test_qltk_songmodel.py
slosd: (I followed the test above for consistency)

This construct is just to make a one-liner out of a for-loop + `history.append(self.pl.current)` and since `self.pl.next()` returns None it needs `or True`.

quodlibet/qltk/songmodel.py
```diff
@@ -215,18 +217,13 @@ def find(self, song):
...
         """
 
         songs = set(songs)
-        found = []
-        append = found.append
-        for iter_, value in self.iterrows():
-            if value in songs:
-                append(iter_)
-        return found
+        return [iter_ for iter_, value in self.iterrows() if value in songs]
```
slosd, line 226: @declension btw, I just compared these two options with the timeit module, since I keep seeing this pattern in the quodlibet codebase. I have to say, my results are inconclusive. None is always better than the other and the difference is usually very small.

declension: Cool; personally I think comprehensions are neater but yeah not much in it

declension:
Thanks @slosd !

Issue #4060, opened 2022-07-26
Conver prefs checkboxes to Gtk.Switch 


declension:
The proposed change
-------------------
Use [Gtk.Switch](https://docs.gtk.org/gtk4/class.Switch.html) for prefs, instead of [Gtk.CheckButton](https://docs.gtk.org/gtk4/class.CheckButton.html)

Benefits of this change
-----------------------
* It looks nicer (YMMV)
* It's the [GTK way now](https://developer.gnome.org/hig/patterns/controls/switches.html)
* [Fits Android recommendations](https://material.io/components/selection-controls#usage) too FWIW

Any issues this might cause
---------------------------
* Confusing at first
* Bigger / scrolling windows (but this is maybe a win long-term?)

declension:
This was done in #4175

PR #4262, opened 2023-01-24
Merge advanced prefs

jej:
Check-list
----------

 * [X] I've added / updated documentation for any user-facing features.

I am always confused about the advanced preferences being in a separate plugin settings pane. It's a lot of clicks to access it. I expect advanced preferences are in the preferences, like most of software. Maybe at the early beginning of QL development it had a sens, but now I think it's better to have these preferences in the core.

This pull request merge the advanced prefs plugin in the prefs pane. It adds also a tooltip on the revert button.

screenshot_20230124-152703

screenshot_20230124-160325

I renamed `advanced_preferences.py` (plugin file) to `advanced_prefs.py`, to have a better filename consistency regarding the current `prefs.py` filename.

Hope you like this big change!

### DIFF
--- a/docs/guide/stats_rating.rst
+++ b/docs/guide/stats_rating.rst
@@ -9,7 +9,7 @@ Play Count
 The internal ``~#playcount`` tag is incremented when a song ends
 or is forced to end by the user, and the playback time has exceeded
 a certain proportion of the song's duration. This proportion is
-configurable in the config file or with the *Advanced Preferences* plugin.
+configurable in the config file or in the advanced preferences pane.
 
 In case of radio streams, which don't have a defined duration, the play
 count gets incremented whenever the stream is played.
--- a/quodlibet/qltk/advanced_prefs.py
+++ b/quodlibet/qltk/advanced_prefs.py
@@ -35,6 +35,7 @@ def _config(section, option, label, tooltip=None, getter=None):
     revert.add(Gtk.Image.new_from_icon_name(
         Icons.DOCUMENT_REVERT, Gtk.IconSize.BUTTON))
     revert.connect("clicked", on_reverted)
+    revert.set_tooltip_text(_("Revert to default"))
 
     lbl = Gtk.Label(label=label, use_underline=True)
     lbl.set_mnemonic_widget(entry)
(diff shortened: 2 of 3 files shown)


jej:
Tests fail because of filename renaming and translation `.po` files.

declension:
> Tests fail because of filename renaming and translation `.po` files.

That one is easy to fix, you just need to update [`POTFILES.in`](https://github.com/quodlibet/quodlibet/blob/main/po/POTFILES.in) probably

declension:
This PR should probably reference #2156 (in its description)

declension:
Thanks for this @jej . Probably a few tweaks needed still.

One thing is we've moved to `Gtk.Switch` recently for boolean configs, see #4060 . It would be good if this pane didn't then use the older checkboxes, but might not be the worst problem. 

More product-wise, not everybody was keen on have this there. I personally think it's a slight simplification / improvement which should feel more intuitive to users, but I can see the other side. @lazka would you be happy with this?

jej:
Thanks for your feedback.

I can do this "switch", need to geek a bit these days 🤓

Concerning the pro and cons to have these prefs here: I guess QL users are not too novice. When I try to convert friend users they don't use it so much and fall back to Rhythmbox and/or online music platforms... So I guess most QL will not be surprised (and maybe happy) to see all these new preferences here for free. Firefox has its `about:config` in core, not as extension, so it's the same logic here to me. Waiting for Lazka feeback of course.

Thanks,
J.

lazka:
> More product-wise, not everybody was keen on have this there. I personally think it's a slight simplification / improvement which should feel more intuitive to users, but I can see the other side. @lazka would you be happy with this?

My initial motivation for the advanced prefs was to get them out of the preferences, so I'm probably not the right person to ask ;)

jej:
> My initial motivation for the advanced prefs was to get them out of the preferences, so I'm probably not the right person to ask ;)

May I ask if you would change your mind now on this topic? 😀

Maybe an alternative is to display adv prefs pane only when the plugin is enabled. But I am suprised to see core features configuration away from the software itself.

lazka:
> May I ask if you would change your mind now on this topic? 😀

Probably a reason for each setting why it is better to show it to all users.

But anyway, I'm not very active in this project anymore, so I don't think things should block on my opinion.

declension:
@jej could you rebase / merge from `main` and I'll have a proper look now 4.6 is out :smiley:

declension:
Or, I can make another branch and redo it on current `main`, I don't mind

jej:
Hi,
I completely forgot this PR... It was too difficult to merge manually after the big _ruff_ cleaning so I opened a new PR here: #4408

PR #4264, opened 2023-02-05
feat: Unique filters plugin

LoveIsGrief:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix --> https://github.com/quodlibet/quodlibet/discussions/4222
 > There don't seem to be any tests for plugins?
 * [x] I've added / updated documentation for any user-facing features 
 > Is the plugin description enough?

This plugin allows filtering by a unique tag e.g `@(unique: artist)` which will result only one song showing up per artist. It can be used multiple times e.g if one has mix albums in the library (albums with multiple artists),
 it's possible to use `&(@(unique: artist), @(unique: album))` to have one song per artist and one song per album

### DIFF
--- a/quodlibet/ext/query/unique.py
+++ b/quodlibet/ext/query/unique.py
@@ -1,0 +1,55 @@
+# Copyright 2023 LoveIsGrief
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+from typing import Any, Optional
+
+from quodlibet import _, print_d, print_w
+from quodlibet.formats import AudioFile
+from quodlibet.plugins.query import QueryPlugin, QueryPluginError
+
+
+class UniqueQuery(QueryPlugin):
+    PLUGIN_ID = "unique_query"
+    PLUGIN_NAME = _("Unique Query")
+    PLUGIN_DESC = _("Filter search results by unique tags.")
+    key = 'unique'
+    query_syntax = _("@(unique: tag)")
+    query_description = "<tt>tag</tt> can be album, artist, title or any other tag. " \
+        "Use multiple <tt>@(unique: tag)</tt> to filter by multiple tags."
+
+    usage = f"{query_syntax}\n\n{query_description}"
+
+    def __init__(self):
+        self.unique_tag_values = set()
+        """The unique tag values that have been seen in the songs being filtered"""
+        self._reported = set()
+        """Unique errors to counter error log spam"""
+
+    def search(self, song: AudioFile, body: Optional[Any]) -> bool:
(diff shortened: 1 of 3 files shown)


LoveIsGrief:
This should be ready now :slightly_smiling_face:

quodlibet/ext/query/unique.py
```diff
@@ -0,0 +1,54 @@
...
+            return_value = field_value not in self.unique_tag_values
+            self.unique_tag_values.add(field_value)
+            print_d(f"unique filtering value '{field_value}': {return_value}")
+        except KeyError:
+            pass
+        except Exception as e:
+            err_str = str(e)
+            if err_str not in self._reported:
+                self._reported.add(err_str)
+                print_w("%s while filtering unique values for '%s': %s "
```
declension, line 43: Generally we're moving to f-strings for new code

quodlibet/ext/query/unique.py
```diff
@@ -0,0 +1,54 @@
...
+    PLUGIN_DESC = _("Filter search results by unique tags.")
+    key = 'unique'
+    query_syntax = _("@(unique: tag)")
+    query_description = "<tt>tag</tt> can be album, artist, title or any other tag. " \
+        "Use multiple <tt>@(unique: tag)</tt> to filter by multiple tags."
+
+    usage = f"{query_syntax}\n\n{query_description}"
+
+    def __init__(self):
+        print_d("Initialising unique query")
```
declension, line 24:
```suggestion
```
Probably not necessary any more

quodlibet/ext/query/unique.py
```diff
@@ -0,0 +1,54 @@
...
+        self.unique_tag_values = set()
+        """The unique tag values that have been seen in the songs being filtered"""
+        self._reported = set()
+        """Unique errors to counter error log spam"""
+
+    def search(self, data, body):
```
declension, line 30: Could you add some type annotations? Trying to do that for all / nearly all new code

declension (approved):
LGTM generally, thanks!

quodlibet/ext/query/unique.py
LoveIsGrief: I added it for the method signature. Dunno if variables and such also need them. If would be to have `mypy` check them and throw errors for where you want type hints to be put.

quodlibet/ext/query/unique.py
LoveIsGrief: Done

quodlibet/ext/query/unique.py
LoveIsGrief: Removed

LoveIsGrief:
Thanks for the review @declension . I hope it's alright now.

declension:
Thanks!

PR #4277, opened 2023-03-11
Update poetry deps + Fix Windows CI

lazka:
And try to fix Windows CI broken since #4260


declension:
Thanks

lazka:
The downside is that this now requires poetry 1.3+

I couldn't get an older version to generate a lock file (it hangs forever)

PR #4331, opened 2023-07-21
Simple plugin to show player in fullscreen

toy:
Check-list
----------

Simple plugin to switch player to fullscreen mode when enabled.


declension:
Thanks. I don't think this is something that makes sense as a plugin though (we have too many probably already) - it should be a config option at app level.

toy:
My impression was that global settings are intentionally small and everything possible should be covered in plugins. Also I don't see a tab in settings fitting for "open in fullscreen" checkmark.

PR #4332, opened 2023-07-21
create console context only with app when no song is selected

toy:
Check-list
----------

Currently song is created without any context when it is not created for a song, though app can be present in any case.


declension:
This doesn't seem like a problem AFAICT - songsmenu plugins should only ever get called _on_ [a non-empty iterable of] songs.

toy:
@declension Previously I was able to have either the console or sidebar console init without any variables, but can't reproduce it anymore, so closing

PR #4341, opened 2023-08-01
Repeat should pass set_explicit and set_implicit instead of set

toy:
Check-list
----------

If only set is defined, both set_explicit and set_implicit are calling set on wrapped object as defined by Order, so it is not possible to distinguish in the wrapped. So this change is needed if I want to distinguish set_explicit/set_implicit in some shuffle plugin and have also repeat enabled.

### DIFF
--- a/quodlibet/order/repeat.py
+++ b/quodlibet/order/repeat.py
@@ -21,8 +21,11 @@ class Repeat(Order):
     def next(self, playlist, iter):
         raise NotImplementedError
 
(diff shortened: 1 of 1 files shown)


declension:
Thanks. I'm not sure I see the use case here. Could you file an feature request / bug first?

toy:
@declension https://github.com/quodlibet/quodlibet/issues/4354

declension:
Thanks

PR #4358, opened 2023-08-23
Fix startup on Python 3.12

luk1337:
Check-list
----------

This makes QuodLibet work with Python 3.12.

### DIFF
--- a/quodlibet/_import.py
+++ b/quodlibet/_import.py
@@ -8,6 +8,7 @@
 
 import sys
 import importlib
+import importlib.util
 
 
 class RedirectImportHook:
@@ -30,6 +31,11 @@ class RedirectImportHook:
 
         self._name = name
         self._packages = packages
+
+    def find_spec(self, fullname, path, target=None):
+        loader = self.find_module(fullname, path)
+        if loader is not None:
+            return importlib.util.spec_from_loader(fullname, loader)
 
     def find_module(self, fullname, path=None):
         package = fullname.split(".")[0]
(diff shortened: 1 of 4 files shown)


luk1337:
> Plugins don't load.

`importlib.util._find_spec()` works but is hacky...

quodlibet/util/importhelper.py
```diff
@@ -92,16 +92,16 @@ def load_module(name, package, path):
     except KeyError:
         pass
 
-    loader = importlib.find_loader(fullname, [path])
-    if loader is None:
+    spec = importlib.util._find_spec(fullname, [path])
```
luk1337, line 95: Any idea how we can avoid calling private method here?

quodlibet/_import.py
```diff
@@ -31,6 +32,19 @@ def __init__(self, name, packages):
         self._name = name
         self._packages = packages
 
+    def find_spec(self, fullname, path, target=None):
```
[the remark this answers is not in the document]
luk1337: Now it's ok

quodlibet/util/importhelper.py
luk1337: This is also done.

luk1337:
This is ready for review.

lazka:
thanks

luk1337:
Rebased since fbc3260282645403dc2bc7c248bb584b27f0958f pulled some of my changes.

lazka:
This broke the loading of the titlecase plugin for some reason

luk1337:
> This broke the loading of the titlecase plugin for some reason

This works for me with Python 3.12.2 and quodlibet#main on Linux.

lazka:
It depends on the order the plugins are loaded, which is random.

Likely fixed in 7b247a5e0c39006c089efd38ca37e80b527baa23

Issue #4152, opened 2022-09-27
[Proposal] Tag pattern syntax to return the first non-empty tag in a list


afontenot:
The proposed change
-------------------
It's common to have complex, chained if-else tag patterns like this:

    <composer|<composer>|<albumartist|<albumartist>|<artist>>>

If you use the above in a string field that accepts tag patterns, the result will be the `composer` of a given song (if it exists) falling back to `albumartist` of a given song (if it exists), falling back to `artist`. What if, instead, you could just do this:

    <<composer||<albumartist>||<artist>>

This proposal introduces a new bit of syntax, an "or" token `||` that tells QuodLibet to use the first provided tag pattern that is non-empty (an empty string or missing tag).

I have already written the code to do this https://github.com/quodlibet/quodlibet/pull/4145. Please check it out and give feedback if you're familiar with the QL codebase.

Yes, you can chain tags as expected:

    <<tag1>||<tag2>||<tag3>>

You can use arbitrary sub-patterns within the "or" pattern

    <<genre=classical|<composer>>||<albumartist>||<artist>||Artist Missing>

If anyone has a good idea for a name for this feature to use in documentation, I'm happy to hear it. Internally I'm calling it the [disjunction](https://en.wikipedia.org/wiki/Logical_disjunction) operator, but most people won't be familiar with that concept. "Or" is a bit vague unless you're a programmer, in which case you might be familiar with Python's `x or y or z`, which behaves the same way.

Benefits of this change
-----------------------
This change would massively simplify complex tag chains, making them considerably easier to create, read, and maintain.

Any issues this might cause
---------------------------
This would not affect existing syntax. The "if-else" pattern continues to work as before.

Because the `|` character has to be escaped in tag patterns anyway, I don't believe it is possible for this change to break any tag pattern that is currently valid. The one possibility I can see is that a user might inadvertently use `||` when they want to use `|`, but we already have tons of cases where syntax errors break things with no help or error reporting to the user. (*whispers: I'm working on something that will fix this.* :rocket:)

bassstorm:
Looks nice, great work!

slosd:
This kind of pattern `<x|<x>|<y|<y>|z>>` shows up in a few places in my config, so this feature would be very useful to keep these definitions more readable.

afontenot:
I'll also point out that it enables some stuff that currently can't be done without repeating yourself. For example, 

[...]

PR #4372, opened 2023-09-12
Add disjunction / fallback tag expression

afontenot:
Check-list
----------

Add a new piece of tag pattern syntax, with the operator `||`. This operator returns the tag pattern on the left side, if it is not empty, and otherwise returns the tag pattern on the right side.

It is implemented so as to be chainable, e.g. one can do

    <<composer>||<albumartist>||<artist>>

And the first non-empty expression will be used.

In documentation, this syntax will be called the "fallback" tag expression, for greater clarity.

Fixes https://github.com/quodlibet/quodlibet/issues/4152.

This PR also documents the new feature and improves the documentation for tag patterns to make it more accessible to new users. Tests have been added.

### DIFF
--- a/docs/guide/tags/patterns.rst
+++ b/docs/guide/tags/patterns.rst
@@ -3,12 +3,28 @@
 Tag Patterns
 ============
 
-Tag patterns allow more complex string representation of tags, with a
-notion of if-null defaulting; whilst this sounds complex (and it can be!)
-it's very useful where libraries have wildly varying tags.
+Tag patterns can be used to change the information displayed in various places
+like the playing song area, columns in the song list and the album list in the
+album browser.
 
-Tag patterns can include strings (e.g. tag names) enclosed in angled
-brackets like this
+They can also be used to group songs in the paned and album collection browsers
+in more complex ways.
+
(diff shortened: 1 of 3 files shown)


docs/guide/tags/patterns.rst
```diff
@@ -3,79 +3,154 @@
...
+in more complex ways.
+
+The file renaming tool uses tag patterns to determine the file names and folder
+structure to be used for the moved files.
+
+Usage
+-----
+
+A tag pattern is a string (piece of text) in which certain symbols receive
+special meaning. Parts of the string that carry no special meaning will be
```
declension, line 20:
```suggestion
special meaning.
Parts of the string that carry no special meaning will be
```

(Generally I'm always trying now to edge towards [sembr](https://sembr.org/) for maintainability.)

declension (commented):
This is great, thanks!

If you can be bothered to reflow markup with sembr then great, otherwise I'll do this some time later.

Like the new naming too.

afontenot:
I haven't encountered sembr before, but I think I've formatted the document as requested.

declension:
Thanks!

Issue #4381, opened 2023-09-27
AttributeError: 'NoneType' object has no attribute 'get_indices'


declension:
https://quodlibet.sentry.io/issues/4507867841/?referrer=github_plugin

```
AttributeError: 'NoneType' object has no attribute 'get_indices'
(2 additional frame(s) were not displayed)
...
  File "quodlibet/qltk/songmodel.py", line 267, in next_ended
    self.current_iter = self.order.next_implicit(self, iter_)
  File "quodlibet/order/__init__.py", line 72, in next_implicit
    return self.next(playlist, iter)
  File "quodlibet/order/reorder.py", line 27, in next
    remaining = self.remaining(playlist)
  File "quodlibet/order/__init__.py", line 138, in remaining
    played = set(map(get_index, self._played))
  File "quodlibet/order/__init__.py", line 136, in get_index
    return playlist.get_path(iter).get_indices()[0]

AttributeError: 'NoneType' object has no attribute 'get_indices'
```

declension:
From #4255. See also #4373

Issue #4373, opened 2023-09-16
Shuffle mode causes segfaults


Konzertheld:
Steps to reproduce
------------------
Start QL, try to play some songs in peace

Expected Output
---------------
Songs are played until I decide not to or playlist is empty

Actual Output
-------------
After two songs, playback stops. It's always after two songs. In a lot of cases, QL also crashes. When restarted, it has one of these "an issue has occured, you can continue to use QL but it may be unstable" things. When I press restart, QL starts without issues but apparently the database has not been saved: The songs that played last have no updated "last played" timestamp. Also, the search bar does not contain the search string that it contained last (I always use QL in search library mode).

Test System
-----------

### Which version of Quod Libet?
4.6

### Which operating system 
SteamOS (Arch Linux)

### More
It might be relevant that the file system the songs are on is btrfs. QL was installed as Flatpak.

declension:
Thanks for filling. Did you submit the crash reports? (there are a few similar ones, so will link here if so). 
Without details of what the "crash" was it's hard to say, but there are a few segfaults that have been happening in GTK+ code for people. Unfortuantely there's often less we can do about these but the first thing is getting the logs

Konzertheld:
I am not sure about past crash reports but I submitted one now and added the link to this ticket to it.

I would like to point out though that QL does not always crash. In about half the cases it just stops playing after two songs.

lbrueckner:
I've the same issues since I've updated to 4.6 a couple of days ago.

I noticed that the playback stops and occasional crashes after the second song only then occur (for me) when 

a) I listen on shuffle and 
b) the track list is filtered in a way that the just finished song gets removed, i.e. with a filter like `#(lastplayed > 10 days)`. 

Then the crash always happens in `"quodlibet-4.6.0/quodlibet/order/__init__.py", line 136 in get_index`. I have already submitted some crash reports on this ;-)

Konzertheld:
I can confirm that QL does not stop playback or crash for me when I remove the `#(lastplayed > X)` filter. Shuffle is okay.

Oldbwl:
I too suffered the same issue  - crash report(s) submitted. I am playing a saved search to play only items which have not been played before, the filter is set to prefer the lower playcounts as well.

For now I have had to regress to v 1.4.0 from the Linux Mint Repos.

declension:
[...]

PR #4382, opened 2023-09-27
Revert "Fix shuffle history (#4255)"

declension:
This reverts commit 500d0d11f9d0f45aeda64455f90b645d89ffaa72.

See #4373 #4381


lazka:
(the guess why this crashes is that it stores away Gtk.TreeIter for later user, and that's not supported really by the API)

PR #4391, opened 2023-10-11
libsoup 3 also needs gir1.2-soup-3.0

lbt:
This is not provided by any non-dev packages on debian 12

I see it has been mentioned in the distant past:
https://debian-bugs-dist.debian.narkive.com/dkv2jhqf/bug-820443-quodlibet-quodlibet-won-t-start-valueerror-namespace-soup-not-available

Fixes:
ValueError: Namespace Soup not available for version 3.0


declension:
Thanks!

PR #4404, opened 2023-10-21
🔧 Repo code quality upgrade (with Ruff)

declension:
* Add in naming and other Ruff rules
 * All exclusions configured where possible and resonable, but minimised
 * Use auto-fixes and push through MyPy, test suite, and manual tests to verify as much as poss
 * A lot of diff is just quotes changing
 * From 13,713 errors (!!) down to zero now
 * And a whole bunch of manual fixing too
 * Hardly any IDE warnings now
 * Remind MyPy we're on 3.10 now

:crossed_fingers:


declension:
eek image

PR #4405, opened 2023-10-22
Fix shuffle history (2)

### LINKED ISSUE #4108: Shuffle mode loses played-history when song list is reordered

slosd:
Check-list
----------

This is https://github.com/quodlibet/quodlibet/pull/4255 again, with a fix for the usage of invalid iters (#4373 and #4381).

For reference I looked at the super class `TrackCurrentModel` which keeps an iter around in `__iter`. My implementation does not hold on to iters for longer than that class.

Fixes #4108

### DIFF
--- a/quodlibet/order/__init__.py
+++ b/quodlibet/order/__init__.py
@@ -132,8 +132,9 @@ class OrderRemembered(Order):
         """Gets a map of all song indices to their song from the `playlist`
         that haven't yet been played"""
 
-        def get_index(iter):
-            return playlist.get_path(iter).get_indices()[0]
+        def get_index(iter) -> int:
+            path = playlist.get_path(iter)
+            return path.get_indices()[0]
 
         played = set(map(get_index, self._played))
         print_d("Played %d of %d song(s)" % (len(self._played), len(playlist)))
--- a/quodlibet/order/__init__.py
+++ b/quodlibet/order/__init__.py
@@ -6,7 +6,9 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
-from typing import Optional
+from typing import Any, Dict, List, Optional
+
+from gi.repository import Gtk
 
 from quodlibet import _, print_d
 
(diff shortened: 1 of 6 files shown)


quodlibet/order/__init__.py
```diff
@@ -103,39 +105,42 @@ class OrderRemembered(Order):
...
 
-    def remaining(self, playlist):
+    def remaining(self, playlist) -> Dict[int, Any]:
         """Gets a map of all song indices to their song from the `playlist`
         that haven't yet been played"""
-        all_indices = set(range(len(playlist)))
-        played = set(self._played)
+
+        def get_index(iter):
+            return playlist.get_path(iter).get_indices()[0]
```
declension, line 136: I remember this line coming up a bit whilst debugging. Can we put some null-saftety (etc) here?

If somehow `playlists.get_path(iter)` returns `None`. Or `get_indices()` does, or is empty?...
If they're at least on separate lines, stack traces are a bit quicker to read

declension (changes requested):
Thanks. I think it looks better but haven't had time to understand fully.

Ideally I'd like someone on Arch to test this so we can be sure the crashes won't happen.

quodlibet/order/__init__.py
slosd: I'll split up the expression into multiple lines. Other than that, I'm worried that not throwing on `None` or an empty list will make it harder to discover incorrect code.

declension:
Thanks

PR #4407, opened 2023-10-22
Upgrade all Python syntax to 3.10

declension:
* Mostly auto-applied by Ruff
* Pycharm helped with harder f-strings
* ~Some~ a sizeable amount of manual intervention...


quodlibet/browsers/__init__.py
```diff
@@ -1,23 +1,23 @@
 # Copyright 2004-2005 Joe Wreschnig, Michael Urman, Iñigo Serna
 #           2012 Christoph Reiter
-#           2016-21 Nick Boultbee
+#           2016-23 Nick Boultbee
```
kurtmckee, line 3: This copyright update is not consistently applied throughout all of the changed files.

kurtmckee (commented):
There's a good chance that if this merges it will invalidate some number of existing PRs.

It might be good to merge this after addressing those other PRs, particularly since the majority of these changes are tool-driven, not manual labor (copyright updates excepted).

PR #4408, opened 2023-10-22
Merge advanced prefs plugin into prefs pane

jej:

This PR adds the changes done in #4262. It was too difficult to merge manually after the big _ruff_ cleaning...


jej:
`quodlibet/ext/events/advanced_preferences.py` renamed & moved to `quodlibet/qltk/advanced_prefs.py` so all _.po_ files references are failing...

declension:
A bit of prettification might be called for but I think we can roll forwards and merge - thanks!

PR #4411, opened 2023-10-24
Advanced prefs pane boolean switches

jej:
Convert checkboxes to switches in the advanced prefs pane + cosmetic.

(according to an old discussion: https://github.com/quodlibet/quodlibet/pull/4262#issuecomment-1416762705)

screenshot_20231024-090951


declension:
Thanks; actually I was just about to push the same(ish) thing. For simplicity I guess we can merge this one first and I'll push any leftover changes afterwards

PR #4413, opened 2023-10-24
🎨 Some leftover tidying of advanced prefs pane

declension:
See #4411, now rebased to bits that already achieved

* Inherit from VBox like other panes, makes life a bit simpler
* Tweak padding to match a bit more
* Use customised config Switch, reduces duplication
* Left align labels for consistency with other prefs panes
* Remove some redundant code

### Looks like
image

### DIFF
--- a/quodlibet/operon/commands.py
+++ b/quodlibet/operon/commands.py
@@ -219,9 +219,11 @@ class EditCommand(Command):
                 song.add(key, value)
 
     def _text_to_songs(self, text, songs):
-        text = re.sub(r"^#.*", "", text, 0, re.MULTILINE) # remove comments
-        text = re.sub(r"(\r?\n){2,}", "\n", text.strip()) # remove empty lines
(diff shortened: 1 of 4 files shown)


jej:
Cool.

But labels are left-aligned now, not sure it's more accessible...

declension:
> Cool.
> 
> But labels are left-aligned now, not sure it's more accessible...

Yeah. I think consistency is still the win here though.

Also, newer GTK does this everywhere (including Accessibility prefs!)

image

jej:
No more introduction message after your merge. Is it intentional?

Before:
screenshot_20231025-190200

After:
screenshot_20231025-190803

The message could be modified to be more explicit. IMO it makes the GUI in this pane a little more dressy...

Firefox has this kind of page for about:config

screenshot_20231025-190344

declension:
Yes - that was intentional - that sentence wasn't adding anything new. I would happily add a graphic if someone wants to come up with one (and that Firefox one is new after all)

jej:
Not so new...

FF screenshot

PR #4425, opened 2023-11-13
Fix incorrect property name

thisfred:
Not sure how this happened, looks like maybe an overeager global search and replace, but the `RGSong` and `RGAlbum` objects do not have an `.Error` property, instead it should be `.error`.

Check-list
----------


thisfred:
I think the issue was introduced in commit 3d57a2d39e so maybe @declension you could take a look if my fix makes sense.

thisfred:
(I did not add unit tests, but without this change, the replay plugin sends QL into a crash loop, and with it, everything seems to work as it did before)

declension:
Great spot, thanks! There are some more that got caught, I'm looking now

PR #4430, opened 2023-11-19
Pass the name of the playlist to create (#4392)

### LINKED ISSUE #4392: Imported playlist name should be the same as the filename
The proposed change
-------------------
When playlists are imported, the name of the playlist is the first track's title followed by "and n more", where n is the number of other tracks.

I'm sure there's a good reason for this but I've not seen it in any other music player: rhythmbox, Strawberry, and Tauon Music Box all use the playlist's filename. I find the current default name completely unhelpful! 

Usually, the name of the imported playlist is the name of the file, which an exporting app would set from the name of the playlist.

Benefits of this change
-----------------------
I just exported 25 playlists from Winamp on Windows to Quod Libet on Linux. I can't tell which is which! Using the filename will help me.

Any issues this might cause
---------------------------
I can't think of any potential issues.

lbt:
Check-list
----------

 * [ x] There is a linked issue discussing the motivations for this feature or bugfix

Fixes #4392

### DIFF
--- a/quodlibet/browsers/playlists/util.py
+++ b/quodlibet/browsers/playlists/util.py
@@ -137,7 +137,7 @@ def __create_playlist(name, source_dir, files, songs_lib, pl_lib):
         if win.step():
             break
     win.destroy()
-    return pl_lib.create_from_songs(songs)
+    return pl_lib.create_from_songs(songs, title=name)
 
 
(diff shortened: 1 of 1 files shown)


declension:
Thanks. Can you update your fork from latest `main` to fix those issues.

Ideally a test or two here would be good too

PR #4431, opened 2023-11-20
Assertion missing the f qualifier from an f-string.

lbt:
I noticed a failing test after a PR and then saw the missing f in the test code.


lbt:
update the PR as the output supposed to be in the assert wasn't being captured.

PR #4453, opened 2023-12-27
Fix the clickability when always allowing sorting

### LINKED ISSUE #4447: "Always allow sorting" preference isn't restored from config properly
### Steps to reproduce
* Launch Quod Libet
* Enable "always allow sorting" in preferences
* Restart QL
* Select Browse>Playlists
* Click on any column header

### Expected Output
Sorting

### Actual Output
Nothing

4.6.0

### Which operating system 
Ubuntu 23.10

declension:
Override wasn't affecting the _clickability_ of the column header

Unit testing this is proving pretty impossible though...

Fixes #4447


toto789:
The fix results in a new bug: the context menu on a song title no longer works.

Issue #4470, opened 2024-01-18
Request: Increase maximum track number support in Ex Falso from the current limit of 999


atc0005:
The proposed change
-------------------

Increase the maximum supported track number in Ex Falso from the 999 limit to 9999.

This would effectively mirror the change last made in d8c637012173d75ed574d7f668c6068b47c6797b (nearly 17 years ago) when the limit was raised by a factor of 10.

Benefits of this change
-----------------------

When processing batches of files from multiple discs the current limit of 999 can be exceeded (e.g., audiobooks). Raising the limit to 9999 would better support larger batches of tracks that are part of the same overall "album".

Any issues this might cause
---------------------------

None that I am aware of. I've directly modified the `/usr/lib/python3/dist-packages/quodlibet/qltk/tracknumbers.py` file on my Ubuntu 20.04 system and have experienced no issues when processing multiple batches of files. The UI resized perfectly to accommodate the larger maximum track limit.

PR #4471, opened 2024-01-18
TrackNumbers: Up limit to 9999

atc0005:
Check-list
----------

- [x] There is a linked issue discussing the motivations for this feature or bugfix
   - https://github.com/quodlibet/quodlibet/issues/4470
- [ ] Unit tests have been added where possible
- [ ] I've added / updated documentation for any user-facing features.
- [x] Performance seems to be comparable or better than current `main`

When processing batches of files from multiple discs the current limit of 999 can be exceeded (e.g., audiobooks). This PR raises the limit from 999 to 9999. 

This effectively mirrors the change last made in d8c637012173d75ed574d7f668c6068b47c6797b (nearly 17 years ago) when the limit was raised by a factor of 10 from 99 to 999.


declension:
Thanks. We'd need a unit test around this larger number but otherwise looks fine

PR #4491, opened 2024-02-19
Move to zlib 1.3.1 to fix OS X build failure

### LINKED ISSUE #4490: OS X build of main fails because zlib 1.3 cannot be downloaded.
Building QL branch main for OS X fails with a 404 error while trying to download zlib 1.3.  

The problem is that zlib 1.3.1 has been released, and the zlib team has deleted 1.3. from their website.

Patch in progress.

jdykstra:
Move to zlib 1.3.1.  The previous version is no longer available upstream.

Fixes #4490.


declension:
Thanks

PR #4493, opened 2024-02-23
Include configured title/album details in MPRIS plugin

toupeira:
Check-list
----------

 * [-] There is a linked issue discussing the motivations for this feature or bugfix
 * [-] I've added / updated documentation for any user-facing features.

If the main song list is configured to include the version in the title,
or the disc subtitle in the album, add these to the MPRIS metadata as
well, since MPRIS doesn't define any fields for them.

These are the existing settings I'm reusing (enabled by default):

image

Tested with https://extensions.gnome.org/extension/4470/media-controls/ on GNOME.

### DIFF
--- a/quodlibet/ext/events/mpris/mpris2.py
+++ b/quodlibet/ext/events/mpris/mpris2.py
@@ -14,6 +14,7 @@ from senf import fsn2uri
 
 from quodlibet import app
 from quodlibet.order.repeat import RepeatListForever, RepeatSongForever
+from quodlibet.qltk.songlist import get_columns
 from quodlibet.util.dbusutils import DBusIntrospectable, DBusProperty
 from quodlibet.util.dbusutils import dbus_unicode_validate as unival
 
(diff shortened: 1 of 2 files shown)


toupeira:
@declension could you please give this a review when you have time, and do you think it makes sense adding this by default? I wouldn't mind adding plugin preferences instead if you prefer that.

This was something that was bugging me since I have quite a few songs with `version` tags in my library, and now I just got a remix album where all titles are in fact the same :sweat_smile:  I don't use disc subtitles but figured I could add those as well while I'm at it.

In my case I have a GNOME extension configured to show only the title and artist in my top bar, truncated to 300px, and can click on it to see more details:

image

So I really only care about seeing the title and version quickly, and the artist is usually truncated anyway.

But of course there are many MPRIS clients and many possible ways to display this metadata :wink:

quodlibet/ext/events/mpris/mpris2.py
```diff
@@ -279,7 +280,13 @@ def ignore_overflow(dbus_type, value):
                 metadata["xesam:" + xesam] = list(map(unival, vals))
 
         # All single values
-        sing_val = {"album": "album", "title": "title", "asText": "~lyrics"}
+        columns = get_columns()
+        title_tag = "~title~version" if "~title~version" in columns \
+                    else "title"
+        album_tag = "~album~discsubtitle" if "~album~discsubtitle" in columns \
+                    else "album"
```
declension, line 287: It's a shame perhaps we don't save that setting, as this _feels_ hacky to need to do.


```suggestion
        title_tag = ("~title~version" if "~title~version" in columns 
                     else "title")
        album_tag = ("~album~discsubtitle" if "~album~discsubtitle" in columns
                     else "album")
```

declension:
Thanks

quodlibet/ext/events/mpris/mpris2.py
toupeira: @declension ah yes, took me a moment to figure this out :grinning: 

Thanks for merging!

PR #4499, opened 2024-03-11
PlaycountEqualizer extension improvements

### LINKED ISSUE #4500: PlaycountEqualizer improvement for big playlists
Hi I've realized that your plugin PlaycountEqualizer does not work well for huge playlists.

After several days I was able to notice track repeating in the 1000+ playlist several times.
Plus there were still many not played tracks. I guess this is because random distribution does not work well when you have like 500 never played tracks and 500 tracks played only once.

So my solution is to introduce additional weight magnitude that will increase probability for a less played tracks. So now weights will be like math.pow(old_weight, magnitude)

See linked PR

rumatoest:
Check-list
----------

Hi I've realized that your plugin PlaycountEqualizer does not work good for huge playlists. 

After several days I was able to notice track repeating in the 1000+ playlist several times.
Plus there were still many not played tracks. I guess this is because random distribution does not work well when you have like 500 never played tracks and 500 tracks played only once.

So my solution is to introduce additional weight magnitude that will increase probability for a less played tracks. So now weights will be like `math.pow(old_weight, magnitude)`

Closes #4500

_P.S. I've decided that maximum magnitude power should be 3 but you could increase it to 4 I guess._

### DIFF
--- a/quodlibet/ext/playorder/playcounteq.py
+++ b/quodlibet/ext/playorder/playcounteq.py
@@ -56,7 +56,7 @@ class PlaycountEqualizer(ShufflePlugin, OrderRemembered, PluginConfigMixin):
         weights_sum = sum(weights.values())
         choice = int(max(1, math.ceil(weights_sum * random.random())))
 
-        print_d("Weighted random: %s for weights sum:%s" % (choice, weights_sum))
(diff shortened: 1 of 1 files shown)


declension:
Thanks

Issue #4504, opened 2024-03-15
MacOS build aborts on XCode 15.3


jdykstra:
Steps to reproduce
------------------
As of XCode 15.3, MacOS builds abort with compilation errors within the internal glib part of `pkg-config`.  Apple changed certain pointer/integer conversion warnings into errors.  

[https://gitlab.freedesktop.org/pkg-config/pkg-config/-/issues/81](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/issues/81)

Expected Output
---------------
```
LIBRARY_PATH=/usr/lib ./build.sh
[an hour of build output]
*** success *** [85/85]
```

Actual Output
-------------
```
LIBRARY_PATH=/usr/lib ./build.sh
...
  CC       libglib_2_0_la-gconvert.lo
  CC       libglib_2_0_la-gdataset.lo
/Volumes/nobackup-cs/quodlibet/dev-utils/osx_bundle/_home/jhbuild_checkoutroot/pkg-config-0.29.2/glib/glib/gatomic.c:392:10: error: incompatible integer to pointer conversion passing 'gssize' (aka 'long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_add ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Volumes/nobackup-cs/quodlibet/dev-utils/osx_bundle/_home/jhbuild_checkoutroot/pkg-config-0.29.2/glib/glib/gatomic.h:170:46: note: expanded from macro 'g_atomic_pointer_add'
    (gssize) __sync_fetch_and_add ((atomic), (val));                         \
                                             ^~~~~
/Volumes/nobackup-cs/quodlibet/dev-utils/osx_bundle/_home/jhbuild_checkoutroot/pkg-config-0.29.2/glib/glib/gatomic.c:416:10: error: incompatible integer to pointer conversion passing 'gsize' (aka 'unsigned long') to parameter of type 'gpointer' (aka 'void *') [-Wint-conversion]
  return g_atomic_pointer_and ((volatile gpointer *) atomic, val);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #4505, opened 2024-03-15
Work around internal glib build failure in dependency pkg-config

jdykstra:
XCode 15.3 is treating some compile warnings as errors that it didn't before, breaking the internal glib build in pkg-config.  Turn the errors back into warnings while the pkg-config team addresses the underlying issue. (#4504)

Check-list
----------

 * [ X ] There is a linked issue discussing the motivations for this feature or bugfix
 * [ n/a ] Unit tests have been added where possible
 * [ n/a ] I've added / updated documentation for any user-facing features.
 * [ X ] Performance seems to be comparable or better than current `main`

XCode 15.3 is treating some compile warnings as errors that it didn't before, breaking the internal glib build in pkg-config.  Turn the errors back into warnings while the pkg-config team addresses the underlying issue. (#4504)

### DIFF
--- a/dev-utils/osx_bundle/modulesets/quodlibet.modules
+++ b/dev-utils/osx_bundle/modulesets/quodlibet.modules
@@ -128,7 +128,10 @@
             hash="sha256:31747ae633213f1eda3842686f83c2aa1412e0f5691d1c14dbbcc67fe7400cea"/>
   </cmake>
 
-  <autotools id="pkg-config" autogen-sh="configure"
+  <!-- Disable integer conversion warnings (which are fatal in XCode 15.3 and later)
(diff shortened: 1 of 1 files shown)


lazka:
thanks.

a long-term fix would probably be to switch to pkgconf.

jdykstra:
> a long-term fix would probably be to switch to pkgconf.
> 

The current MacOS build arrangement of building 88 packages from source is fragile.  If I get inspired to clean this up, my first thought would be to download built dependencies from some place like homebrew (without disturbing any existing homebrew installation).  Unfortunately, we’d still need some of the GNOME build infrastructure, and they use pkg-config.

  —  John

PR #4515, opened 2024-04-13
Support floats in MPD seekcur command

dmcc:
Some MPD clients use the `seekcur` command with non-integer arguments, which results in an error from Quod Libet. For example, attempting to seek in Home Assistant's [MPD integration](https://www.home-assistant.io/integrations/mpd/) results in the following error:

image

This PR has been confirmed to work locally and includes some new unit tests.

### DIFF
--- a/quodlibet/ext/events/mpdserver/main.py
+++ b/quodlibet/ext/events/mpdserver/main.py
@@ -815,7 +815,7 @@ def _cmd_seekcur(conn, service, args):
         relative = True
 
     try:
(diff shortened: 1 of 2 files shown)


lazka:
Thanks, that's also what MPD itself does: https://github.com/MusicPlayerDaemon/MPD/blob/823d6c9c0f68ba56ef995d83f3084d4a386f04b1/src/protocol/ArgParser.cxx#L156

Issue #4522, opened 2024-05-05
Lyrics panel: hide timestamps when embedded lyrics are in .lrc format


viltered:
The proposed change
-------------------
The lyrics panel shows the raw contents of the `~lyrics` tag. This includes timestamp annotations for synced lyrics, which is not helpful. I propose an option to display only the text itself without timestamps.

Benefits of this change
-----------------------
Users with embedded synced lyrics in their library get a cleaner view.

Any issues this might cause
---------------------------
None that I'm aware of.

Issue #1989, opened 2016-07-17
synchronized lyrics plugin: Support embedded lyrics


cweiske:
The "synchronized lyrics" plugin tries to read an `.lrc` file with the same name as the current audio file.

- MP3's ID3v2 defines the `SYLT` tag for that: http://id3.org/id3v2.3.0#Synchronised_lyrics.2Ftext
- ogg/vorbis files may contain a "kate" stream that contains lyrics https://wiki.xiph.org/OggKate
  - `katedec -t lrc file.ogg` can be used to extract that stream into .lrc format
- kid3 has a usable synchronized lyrics editor for mp3; http://kid3.sourceforge.net/

It would be nice if the synchronized lyrics plugin would support reading the synchronized lyrics embedded in the audio files instead of requiring external files.

declension:
See also #1628 

I suggest that ogg/vobis Kate support is excluded from this feature (and another raised in Mutagen as necessary - AIUI it's not supported currently). This ticket can be about `~lyrics support` which should be easy enough

declension:
Hmm and editing "info" area (see #1068) that deals with lyrics editing should maybe support embedded lyrics, but that's _another_ ticket.

lazka:
We could also handle lyrics similar to embedded images by provide extra API in AudioFile to load/save. That would mean it's not searchable like tags, but we could load it form multiple sources.

declension:
@lazka I'm really keen on searchable lyrics if possible, and to an extent this is do-able in QL (e.g `~lyrics=sunshine`), but the API idea sounds good; maybe this could be used to retrieve (more like Album Art Downloader then I guess) before embedding *or* saving to `.lrc`, etc.

Synchronised ones would make this harder, but I'm sure a data structure that spans `.lrc` and `SYLT` and Kate is feasible

Nikephor:
So, has anything been done to fix this? I spend a lot of time manually syncing my mp3 files' lyrics because I'm a perfectionist. Not being able to use the embbeded ones is stupid, the whole thing is already there, I just need the plugin to read them from the tag itself.

elfalem:
@Nikephor I have always intended to add the option to read LRC formatted lyrics in embedded tags.  Your changes in #3173 seems to achieve this. However this issue also mentions reading from the ID3v2 SYLT tag and ogg/vorbis kate streams which requires significant additional work.

cweiske:
[...]

PR #4523, opened 2024-05-05
Add option to hide timestamps from lyrics panel

viltered:
Check-list
----------

Implements #4522 

See also #1989 - if embedded lyrics are to be supported by the synchronized lyrics plugin, the lyrics panel should be able to deal with them as well.


declension:
Thanks

Issue #4543, opened 2024-07-02
Symbolic links are replaced with regular files


Peptis:
Steps to reproduce
------------------
Context:
- I store my `~/.config/quodlibet/config` file in a `dotfiles` folder that is tracked by git.
- I use GNU Stow to create a symbolic link back to the `dotfiles` folder's `config` file.

Action:
- I changed program configuration via the UI.

Expected Output
---------------
I expect the `config` file in my `dotfiles` folder to be updated, and the symbolic link to remain in place.

Actual Output
-------------
The symbolic link is replaced by a regular file. The upshot is that I lose the ability to track my config change in my `dotfiles` git repo.

Test System
-----------

### Which version of Quod Libet?
4.6.0

### Which operating system 
EndeavourOS (Arch Linux) -- rolling release

### If it's audio-related, what back-end?
N/A

Additional Information
----------------------
The [atomic_save](https://github.com/quodlibet/quodlibet/blob/3dcf31dfc8db9806d1f73a47fdabc950d35ded1d/quodlibet/util/atomic.py#L41) method seems to be the culprit.

This uses a safe method to atomically write to `filename` by following these steps:
1. creates a temporary file
2. writes to the temporary file
3. moves the temporary file to `filename`

The last operation replaces the symbolic link with the regular (temporary) file.

In the case where `filename` is a symbolic link, would it make sense to overwrite the link destination, rather than the link itself?

Peptis:
Hi, I've developed a fix for this bug, please see attached patch:

[symbolic-link.PATCH](https://github.com/user-attachments/files/16093567/symbolic-link.PATCH)

MilesBHuff:
> Hi, I've developed a fix for this bug, please see attached patch:
> 
> [symbolic-link.PATCH](https://github.com/user-attachments/files/16093567/symbolic-link.PATCH)

Would you be willing to please make a pull request for this?  It'd be nice to have this fixed in mainline.

Peptis:
Sure, will do :)

declension:
Closed by #4548

PR #4548, opened 2024-07-07
Fix: Dereference symbolic links when writing to file (Issue 4543)

Peptis:
Fix for atomic_save overwriting symbolic links. Move the temporary file to the symbolic link target, rather than replacing the symbolic link.

Check-list
----------

 * [N/A] I've added / updated documentation for any user-facing features.

Fix for atomic_save overwriting symbolic links - [Issue 4543](https://github.com/quodlibet/quodlibet/issues/4543). The temporary file created in atomic_save is moved to the symbolic link target, rather than replacing the symbolic link itself.


declension:
Thanks

PR #4550, opened 2024-07-14
Make waveform height configurable (in plugin configuration)

eudoxos:
Check-list
----------

I created a simple int setting for height of the waveform (I prefer it bigger, as I navigate in speech recording with numerous gaps).


declension:
Thanks

Issue #4558, opened 2024-08-08
Windows installer build fails due to dependency issues


peacegiverman:
Steps to reproduce
------------------
1. Set up development environment according to the [Windows Installer instructions](https://github.com/quodlibet/quodlibet/tree/main/dev-utils/win_installer#setting-up-the-msys2-environment).
2. Run `./build.sh`

Expected Output
---------------
The build script runs without errors and creates a Windows installer.

Actual Output
-------------
Two separate errors.

1. pacman error while removing transitional dependencies:
    
    ```bash
    error: target not found: mingw-w64-ucrt-x86_64-gsl
    error: target not found: mingw-w64-ucrt-x86_64-vulkan-headers
    ```

2. After working around the first issue, the build script breaks when it gets to the `depcheck.py` stage.

    ```bash
    Process Process-16:
    Traceback (most recent call last):
      File "D:/dev/quodlibet/dev-utils/win_installer/_build_root/ucrt64/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
        self.run()
      File "D:/dev/quodlibet/dev-utils/win_installer/_build_root/ucrt64/lib/python3.11/multiprocessing/process.py", line 108, in run
        self._target(*self._args, **self._kwargs)
      File "D:/dev/quodlibet/dev-utils/win_installer/misc/depcheck.py", line 27, in _get_shared_libraries
        repo.require(namespace, version, 0)
    gi.repository.GLib.GError: g-irepository-error-quark: Typelib file for namespace 'GIRepository', version '3.0' not found (0)
    ```

Test System
-----------

### Which version of Quod Libet?
Development (`main`)

### Which operating system 
Windows 10 22H2

Additional Information
----------------------

### 1. Missing transitional dependencies

It looks like none of the project dependencies pull `vulkan-headers` and `gsl` packages anymore. Unfortunately, when pacman tries to delete non-existing dependencies it fails with the `target not found` error.

Removing these these two packages from the cleanup list, makes this part of the build process pass fine.

### 2. GIRepository version error

Due to changes in _GLib_, `python-gobject` now ships with two versions of _GIRepository_. This causes `depcheck.py` to softlock during the build process with the following error:

    ```bash
    gi.repository.GLib.GError: g-irepository-error-quark: Typelib file for namespace 'GIRepository', version '3.0' not found (0)
    ```

[...]

PR #4559, opened 2024-08-08
wininst: Fix dependency issues (#4558)

peacegiverman:
Check-list
----------

Removes transitional dependencies that aren't used anymore and adds a workaround for `GIRepository` version errors. (#4558)

Note that the current state of MSYS packages that include `GIRepository` is probably transitional and may transition to having only one, `GIRepository` 3.0 at some point.


lazka:
works I guess, but it's just hiding an error that should never happen. I'll have a look.

PR #4566, opened 2024-08-23
tests: Filter out unsupported file formats properly

luk1337:
Check-list
----------

This change fixes tests on Fedora 41 as https://src.fedoraproject.org/rpms/gdk-pixbuf2/c/a5f3d593c19d4ec7a46699422fbbf4ccbbc05db9 is missing from 41 and rawhide for whatever reason.


declension:
Thanks

PR #4576, opened 2024-09-04
test_brainz: Mark as requiring network

luk1337:
Check-list
----------

This test requires network, so skip it when pytest is ran with -m "not network".


declension:
Thanks

PR #4577, opened 2024-09-04
gdist: Fix filtering dirs logic in get_missing()

luk1337:
We can't iterate and modify the variable at the same time.

Check-list
----------


declension:
Thanks

Issue #54, opened 2015-03-14
Batch lyrics downloading


lazka:
Original [issue 54](https://code.google.com/p/quodlibet/issues/detail?id=54) created by redsun82 on 2008-11-21T10:40:26.000Z:

I find the lyrics feature to be a little uncomfortable to use for two reasons:
- Lyrics are too much clicks/keytrokes away from the song: you select the
  song, right click on it, go to information, then to the Lyrics tab and then
  most of the time you still have to click on &quot;dowload&quot; to retrieve them. Too
  much, I think! I like it very much in Rhythmbox that the lyrics plugin is
  called just by the combination ctrl+L. At least there should be an
  automatic download when the tag is empty and queried.
- This other one would be a killer feature in my opinion: batch lyrics
  download in tag editor/exfalso. It could work much like album art: you
  select files, right click and there you go, all files tagged (instead of
  doing a file at a time). Can be most appreciated by ipod users (or others
  media player, I don't know), where lyrics tags can be displayed.

kellogsgit:
Getting back to the issue.
http://lyrics.wikia.com/wiki/LyricWiki:Permission

`We give permission to the MusicBrainz project and/or its users and/or its parent organization the MetaBrainz Foundation to link to LyricWiki. `

So we are able to link using MusicBrainz (but it is broken as we cannot today link to recordings, only work, but this could change).

`If you would like to show our lyrics to your end-users, the only valid way currently is to display the whole LyricWiki page inside of an HTML pane or browser window (including the ads) with javascript enabled. `

Not workable.

kellogsgit:
MusicBrainz allow to store lyrics relationships:
https://beta.musicbrainz.org/search/edits?auto_edit_filter=&combinator=and&conditions.0.args=271&conditions.0.field=link_type&conditions.0.operator=%3D&negation=0&order=desc&page=3

kellogsgit:
After for hastle, I could add:
https://musicbrainz.org/work/ce407c72-7a2a-4515-a66f-6098d0427960

Tamriel:
[Here](https://bigishdata.com/2016/09/27/getting-song-lyrics-from-geniuss-api-scraping/) is the code for crawling lyrics from Genius. One could create a Quod Libet plugin. The user has to set his Genius API key in the Quod Libet setting once. It adds a new context menu entry 'Fetch lyrics for selected songs'.

declension:
FWIW I found that Genius crawling unreliable (recently anyway)

PR #4619, opened 2024-12-02
Add a plugin for downloading song lyrics

panchaea:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix (#54)

Adds a plugin that downloads song lyrics for a selection of songs. The heavy lifting is offloaded to a [`beets`](https://beets.io/) plugin called [`lyrics`](https://beets.readthedocs.io/en/stable/plugins/lyrics.html), but as most users are probably not interested in using `beets`, this is treated as an implementation detail (although I think Quod Libet would make for a good platform for a `beets` UI of sorts and may look into this in the future). This approach has been pretty reliable in my testing, at least for songs which are in MusicBrainz and have lyrics on Genius, but it is not very fast - I would probably not want to download lyrics for more than a few albums at a time.

I also updated the `nix` flake since it was missing some libraries which are necessary for the UI to work.

By the way, I am open to adding a unit test for the plugin but I have not done so yet as it involves adding a test fixture which actually has lyrics on a supported lyrics service (i.e., is identified by `beets` as being a real song). I think a blank audio file with the correct ID3 tags would be sufficient, but asserting anything meaningful about the downloaded lyrics might be problematic for copyright reasons.


declension:
Thanks for the PR and apologies for the delay. 

I don't think we can really include `beets` as part of QL (note though beets does use [Mutagen](../../mutagen) FWIW, but as is I don't think this can be merged.

Would be happy to look at other (even partial) solutions to #54 though of course. Note some minor improvements to the lyrics editing now due in #4712.

But note there is increasing need for a separate plugins org / repo / something for unofficial plugins where this might be able to live in future. Some users have their own (e.g. https://github.com/pschwede/quodlibet-plugins/)

PR #4625, opened 2024-12-21
Exfalso: add progress bar (fixes #4604)

### LINKED ISSUE #4604: Exfalso: no progress bar for background tasks
Steps to reproduce
------------------

- Run the MusicBrainz lookup plugin from exfalso to retag big files on a slow file system.
- When validating, background tasks are rerun in background to write the modified files
- There's no progress bar in the exfalso window, so you think it's finished and you close the window, *but* it's not really finished and some files are not yet written

That's a regression introduced by #4189 = the older waitwindow mechanism showed a big progress bar on top of the exfalso window.

No problem in quodlibet as there's a statusbar widget showing the progress. I guess some code can be adapted from there

Expected Output
---------------

A progress bar showing that the process is still on-going

Actual Output
-------------

Test System
-----------

4.6.0

### Which operating system 

Ubuntu 24.04

n/a

Additional Information
----------------------

abretaud:
Check-list
----------

A small fix for #4604
Added a progress bar in Exfalso, to monitor the progress of tasks.
It looks like this:

progress

### DIFF
--- a/quodlibet/qltk/exfalsowindow.py
+++ b/quodlibet/qltk/exfalsowindow.py
@@ -33,6 +33,7 @@ from quodlibet.qltk.x import Align, SeparatorMenuItem, ConfigRHPaned, \
     SymbolicIconImage, MenuItem
 from quodlibet.qltk.window import PersistentWindowMixin, Window
 from quodlibet.qltk.msg import CancelRevertSave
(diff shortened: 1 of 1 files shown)


declension:
Thanks!

PR #4649, opened 2025-01-22
Add SSL support to bundled Python for use by plugins.

### LINKED ISSUE #4648: QuodLIbet 4.6.1 unable to make web request on MacOS
Originally reported by @erixtekila to [https://github.com/jdykstra/quodlibet/](https://github.com/jdykstra/quodlibet/).

Steps to reproduce

Install Musikbrainz plugin.
Use it on a playlist will produce an error.
It seems that no request could be emitted from the sonoma application…?
Which operating system

MacOS 10.14...

Image

jdykstra:
Correct build of bundled Python interpreter so that it includes the SSL module.  A bug in the 3.11.1 Python release (and others) caused the "--with-openssl" configure parameter to have no effect when the SSL module was built through setup.py.  Change the module build specification so that it is built through Modules/Setup, which correctly implements the relevant configure parameter.

Fixes #4648.

Check-list
----------

 * [X ] There is a linked issue discussing the motivations for this feature or bugfix
 * [X ] Performance seems to be comparable or better than current `main`

The immediate cause of this is that the Python interpreter we include in the application bundle isn't being built with SSL support. (MusicBrainz requires https.) The root cause is a deficiency in the Python build scripts. The "--with-openssl" configure parameter is ignored when the "_ssl" module is built by the top-level setup.py script.

### DIFF
--- a/dev-utils/osx_bundle/modulesets/quodlibet.modules
+++ b/dev-utils/osx_bundle/modulesets/quodlibet.modules
@@ -406,20 +406,19 @@
     </dependencies>
   </autotools>
 
-  <!-- INSTALL_PREFIX because it doesn't use DESTDIR and LIBS="" to skip installing static libs -->
   <autotools id="openssl" autogen-sh="Configure" autogenargs="shared zlib-dynamic"
              autogen-template="%(srcdir)s/%(autogen-sh)s --prefix=%(prefix)s --openssldir=%(prefix)s/etc/ssl %(autogenargs)s"
-             makeinstallargs="INSTALL_PREFIX=$JHBUILD_PREFIX/_jhbuild/root-openssl LIBS= install_sw"
              supports-non-srcdir-builds="no">
     <branch module="openssl-1.1.1t.tar.gz" version="1.1.1.t" repo="openssl"
             hash="sha256:8dee9b24bdb1dcbf0c3d1e9b02fb8f6bf22165e807f45adeb7c9677536859d3b"/>
   </autotools>
 
-  <autotools id="python" autogenargs="--enable-shared --without-ensurepip"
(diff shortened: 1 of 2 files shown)


dev-utils/osx_bundle/modulesets/quodlibet.modules
```diff
@@ -406,20 +406,19 @@
     </dependencies>
   </autotools>
 
-  <!-- INSTALL_PREFIX because it doesn't use DESTDIR and LIBS="" to skip installing static libs -->
   <autotools id="openssl" autogen-sh="Configure" autogenargs="shared zlib-dynamic"
              autogen-template="%(srcdir)s/%(autogen-sh)s --prefix=%(prefix)s --openssldir=%(prefix)s/etc/ssl %(autogenargs)s"
-             makeinstallargs="INSTALL_PREFIX=$JHBUILD_PREFIX/_jhbuild/root-openssl LIBS= install_sw"
```
lazka, line 412: Why is this change needed?

jdykstra: The definitions of INSTALL_PREFIX and LIBS are unnecessary, and I'm deleting them to avoid future confusion.  QL on Mac's prefix directory tree contains almost every library it needs, and we control its contents.  There's no need for  openssl to coexist with a potentially different version from the OS, so we don't need to put it a separate directory.

declension:
Shall we merge this? Or does is still need reviewing?

jdykstra:
I think it's safe.  If I missed something, SSL might break, but it's already broken and used only by a couple of plugins.

declension:
Thanks @jdykstra

PR #4667, opened 2025-03-30
🆕 Auto-enable more plugins:

declension:
* `mpris` (D-Bus) if on Linux - it's vital for so many Linux things
* `WaveformSeekBar` - stable for years, it's in every modern screenshot and there seems little reason to ship without it

Probably some more we could turn on...


lazka:
lgtm

PR #4670, opened 2025-04-04
Require poetry v2 and update pyproject.toml

lazka:
Try to use more of the new standards that are not
poetry specific.

Remove wheel from build-system, not required with setuptools for a long time.

Switch poetry to non-package mode, since we only use it for development.


lazka:
@declension are you OK with depending on poetry v2?

declension:
> @declension are you OK with depending on poetry v2?

I guess. Elsewhere we've 100% ditched Poetry, for [uv](https://docs.astral.sh/uv/) and was actually going to propose using that for QL. But not super-fussed either way

lazka:
Yeah, uv would be nice. I need to add MSYS2 support to it :/

At least this migrates most of the metadata to a format also uv could use, except "tool.poetry.group.dev.dependencies". Once poetry supports dependency groups (https://github.com/python-poetry/poetry/issues/9751) we could have a pyproject.toml that works with both poetry and uv.

lazka:
Both uv and poetry should work now.

declension:
Perhaps if I get some time I'll do a port to `uv` - as I haven't kept up with Poetry 2, plus uv is pretty nice

lazka:
Sure, Windows CI is still stuck on poetry for now, but the others can move.

Issue #4011, opened 2022-06-10
Add light / dark mode detection


louwers:
The proposed change
-------------------

QuodLibet has a Theme Switcher plugin. This plugin has a 'Prefer dark theme version'. I suggest a 'Set dark theme to match system theme on startup' is added.

There is now a freedesktop.org standard that defines how systems can communicate dark preference. See [here](https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.impl.portal.Settings.xml#L40). I recently added this to Anki, which is also an application written in Python ([source code](https://github.com/louwers/anki/blob/7b6893e31dd1b9692bfb50f98c645a32e2f1dd05/qt/aqt/theme.py#L376)).

Benefits of this change
-----------------------

Users that like using dark mode, but who only like dark mode in the evening or when their system has been set to a dark theme, will be able to use the application in dark mode, without having to go into the settings every time to toggle it manually.

Any issues this might cause
---------------------------

- There is no impact to users that do not enable the toggle in the Plugin settings.
- The added code is simple and of low complexity.
- The only downside of this approach is that there is another one: D-Bus also allows you to 'listen' to dark mode changes. This would be an even better implementation, since then users won't have to restart Quod Libet when they toggle their system dark mode preference.

Okanda:
I miss this a lot.

clockworkgeek:
The above approach no longer works with newer versions of Gnome. Instead of gsettings "org.gnome.desktop.interface gtk-theme" or "org.freedesktop.appearance color-scheme" we must now look at "org.gnome.desktop.interface color-scheme".

I made a quick go at fixing this myself but ran into a wall. This correctly reads the current dark mode:

    settings = Gio.Settings.new('org.gnome.desktop.interface')
    darkmode = settings.get_string('color-scheme') == 'prefer-dark'

But then fails to respond to any change as I had expected:

    def on_color_scheme(settings, value):
        # do something with the new value here

    settings.connect('changed::color-scheme', on_color_scheme)

Any other suggestion I can find online is outdated with respect to modern Gnome. Does anyone else know what to do here?

lazka:
> But then fails to respond to any change as I had expected:

I think because the "settings" gets out of scope and destroyed.

Question is if this should be done by default?

I'd suggest:

* enable it by default, so it works out of the box
* in the plugin add an option for the dark setting to "use system defaults", so the plugin can be used to adjust if needed.

[...]

PR #4676, opened 2025-04-15
Match desktop color scheme as it changes

clockworkgeek:
Check-list
----------

Add light / dark mode detection #4011
Honestly I only tested with Gnome Shell on Arch Linux. It relies on Gnome specific settings so might conflict with other environments.
Also there's a slight bug where opening plugin preferences will preset the "Prefer dark theme version" checkbox even if the config value for it does not exist.
However this method does prove that automatic detection is possible.

### DIFF
--- a/quodlibet/ext/events/themeswitcher.py
+++ b/quodlibet/ext/events/themeswitcher.py
@@ -41,6 +41,11 @@ class ThemeSwitcher(EventPlugin):
         self.__default_theme = settings.get_property("gtk-theme-name")
         self.__default_dark = settings.get_property("gtk-application-prefer-dark-theme")
 
+        self.__desktop = Gio.Settings.new('org.gnome.desktop.interface')
+        def on_color_scheme(settings, value):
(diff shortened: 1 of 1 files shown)


declension:
Thanks. I can't replicate this working locally (Gnome on Manjaro) - i.e. when I change it system theme preference, QL doesn't update (but Firefox does for example).

Can you give steps to replicate it working?

clockworkgeek:
Gnome 48 here. For me the theme must be "Adwaita" or "Default Theme" not "Default" (which is misleading tbh). Clicking the associated checkbox does override the system preference.

image
image

For testing I simply hit the given "Dark Style" button in the system menu.

image

The crucial part of this method is checking for a single dconf setting. This has been consistent since around Gnome 40 that I'm aware of.
```
$ dconf read /org/gnome/desktop/interface/color-scheme
'prefer-dark'
```

declension:
Hmm, I still can't get it to set _light_ (or dark, I guess) properly (with default or Adwaita):


In light mode:

image

PR #4731, opened 2025-08-13
⚗️ Add `stale` action to manage PRs and issues better

declension:
* Very conservative values
* Seems like the least bad option at this point


declension:
CC @lazka

PR #4741, opened 2025-08-26
test_formats_xiph: Make _get_jpeg() read file from disk instead of fd

luk1337:
Check-list
----------

It seems like some Gdk behavior got changed and now h.read() returns empty list when trying to read back the JPEG image.

This incidentally makes code match TCoverManagerBuiltin::setUp(), so I guess it's ok anyway.


declension:
Thanks

PR #4750, opened 2025-09-12
Update _id3.py to use album artist for TPE2

### LINKED ISSUE #1179: ID3: Use TPE2 for Album Artist and a TXXX tag for Performer

### LINKED ISSUE #3810: ID3V2 dsf file tag identity
The proposed change
-------------------

Change TPE2 tag from "performer" to "albumartist"

Benefits of this change
To comply with the defacto standard
-----------------------
Music players such as mpd interpret TPE2 as albumartist, so a separate free text entry for TXXX albumartist confuses its database.

Any issues this might cause
---------------------------

I don't think there are significant problems.

mdpenguin:
Use 'albumartist' for TPE2 instead of 'performer' to maintain consistency with other music players and tagging for other file formats.

Closes #1179
Closes #3810

Check-list
----------

Change: Modified _id3.py to make TPE2 map to albumartist.
Reason: This is how other music players and tag editors handle the tag.  It also brings mp3 tagging in line with other file formats that have an albumartist tag so that tracks in collections with multiple formats can be handled the same way (e.g. allows using albumartist tag consistently for both mp3 and flac in searches or in Paned Browser and Album Collection views).

### DIFF
--- a/quodlibet/formats/_id3.py
+++ b/quodlibet/formats/_id3.py
@@ -42,7 +42,7 @@ class ID3File(AudioFile):
         "TIT2": "title",
         "TIT3": "version",
         "TPE1": "artist",
-        "TPE2": "performer",
+        "TPE2": "albumartist",
         "TPE3": "conductor",
         "TPE4": "arranger",
         "TEXT": "lyricist",


declension:
This won't work without a migration - as all users will have their entire libraries changed without warning.

mdpenguin:
What would a migration entail?  You brought that up five years ago and I don't see a record of a discussion about it after that.  Could you tell me who would need to be involved in approving and implementing a migration plan?  Does this project have that kind of organization?

I've made this change in my local installation and it doesn't seem to be causing problems, but my usage is fairly limited (mostly I use Ex Falso).  What issues are you anticipating this causing for users?  Are they worse than the compatibility issues that people are complaining about?

declension:
> What would a migration entail? You brought that up five years ago and I don't see a record of a discussion about it after that. 

Yes, there have been a few ideas, but all have gone stale and not many people around from then now unfortunately.

Quickly thinking through, I guess it would want to 
* scan the library
* filter for ID3 tracks and for each
  * read the old `albumartist` (`TXXX`) and `performer` (`TPE2`)
  * before saving with new mappings 

A random example I just found might help:
```
IDv2 tag info for .../Antonio Carlos Jobim - Wave.mp3
COMM=c0=XXX=YEAR: 2000  ID3G:   8
RVA2=[unrepresentable data]
RVA2=[unrepresentable data]
TALB=The Very Best of Latin Jazz 2
TCON=Latin Jazz
TDRC=1999
TIT2=Wave
TPE1=Antônio Carlos Jobim
TPOS=2
TRCK=9
TXXX=QuodLibet::albumartist=Various Artists
TXXX=replaygain_album_gain=-4.539062 dB
TXXX=replaygain_album_peak=0.999969482888
TXXX=replaygain_track_gain=-4.910156 dB
TXXX=replaygain_track_peak=0.989868164523
```

Or, trickier:

```
IDv2 tag info for .../George Michael & Queen - Somebody To Love.mp3
POPM=EMAIL@DOMAIN=0 204/255
RVA2=[unrepresentable data]
RVA2=[unrepresentable data]
TALB=Ladies & Gentlemen: The Best Of George Michael
TCON=Rock
TIT2=Somebody To Love
TOPE=Queen
TPE2=Queen / George Michael
TPOS=2/2
TRCK=15
TXXX=QuodLibet::albumartist=George Michael
TXXX=replaygain_album_gain=-7.480469 dB
TXXX=replaygain_album_peak=0.928405762151
TXXX=replaygain_track_gain=-7.480469 dB
TXXX=replaygain_track_peak=0.928405762151
```


> Could you tell me who would need to be involved in approving and implementing a migration plan? Does this project have that kind of organization?

I can help a bit, but TBH, no not really :(


> What issues are you anticipating this causing for users? Are they worse than the compatibility issues that people are complaining about?

Experience has shown that there are quite a few QL users with extremely large, complex setups that understandably don't want to be changed.
And without telemetry it's very hard for us to tell users sizes let alone how many people might be affected (or some kind of judgement on whose pain is worse).

Concretely (for my own example of part FLAC, part MP3) - it'll be weird to have half the library suddenly lose (and gain) album artists and performers with no warning.

To be clear though, I am (personally) largely in favour of this change. 

I now think the only pragmatic way forwards is to 
* make it configurable, with (initially at least) the default being as-is (`TPE2` -> performer)
* document this somewhere
* Write or at least map out _some_ kind of possibly lofi migration (Python console script comes to mind) that does as above maybe (but probably better)
* ...probably something else I've forgotten from #1179

mdpenguin:
> Make it configurable, with (initially at least) the default being as-is (TPE2 -> performer)

That makes good sense to me.  Unfortunately, it's more than I'll be able to do for a while.  Would you be willing to implement this or to put out a call for volunteers to help with it?

> Write or at least map out some kind of possibly lofi migration (Python console script comes to mind) that does as above maybe (but probably better)

In lieu of that, how about making the setting require the user to rescan their library or throw a warning stating that it would be required to do so?  We'd also probably want to include a warning to look out for any custom albumartist tags.

Offerel:
As a beginner, I started using Quod Libet yesterday: Right away, I stumbled upon this very problem and wondered why some tracks had no album artist but did have a performer. Is there a way to use a script to transfer the value from Performer to Album Artist?

mdpenguin:
@Offerel, pull requests aren't the best place to ask questions like this.  In all likelihood, no changes will ever be made to Quod Libet to address this issue.  However, you can manually update the file `_id3.py` in `/usr/lib/python3/dist-packages/quodlibet/formats/` to correct the tag.  Look at the Files Changed tab in this PR and it will show you what line you need to change.  You will need to rescan your library after making the change for it to show up in any files that you already have in your Quod Libet library, but any new files added will have the tag labeled correctly.

Keep in mind that this change will be overwritten every time that you update Quod Libet, so you'll have to apply it again each time.

declension:
> > Make it configurable, with (initially at least) the default being as-is (TPE2 -> performer)
> 
> That makes good sense to me. Unfortunately, it's more than I'll be able to do for a while. Would you be willing to implement this or to put out a call for volunteers to help with it?

Yes, I had a look again yesterday actually as the other big PR is now in at least.
Note also #1590 which is the truly good way, but on further look, seems too hard again :| 

> 
> > Write or at least map out some kind of possibly lofi migration (Python console script comes to mind) that does as above maybe (but probably better)
> 
> In lieu of that, how about making the setting require the user to rescan their library or throw a warning stating that it would be required to do so? We'd also probably want to include a warning to look out for any custom albumartist tags.

Yes, something like this maybe. But knowing when to alert the user is itself non-trivial...

mdpenguin:
> Yes, something like this maybe. But knowing when to alert the user is itself non-trivial...

Honestly, I think you're asking a lot more of this software than any users affected by this will.  Take your second example: why would someone who has already purposefully started using a custom Album Artist field -- which is the only way that example could happen -- want to change the current behavior?

All anyone really needs is to be aware of the risk of checking the box.  There are tools available that they can use to clean up data if they need to: e.g. Ex Falso can change the tag being used for a field in bulk, or Puddletag, which can automate the kind of retagging pattern that someone might need to clean up their library.  And, of course, they can always opt not to check the box at all if it is too complicated for their library: not all features have to work for all users.

Issue #4761, opened 2025-10-04
Crashes on Startup (MacOS 15.4)


umikoio:
Steps to reproduce
------------------
After startup, the application crashes after a few seconds on MacOS. Sometimes you're able to go to `File` and pick a audio file to play, but it usually only gets through ~5 seconds before it crashes (if it doesn't crash before).

Expected Output
---------------
The application should not crash on MacOS.

Actual Output
-------------
Here is the error message:
```
Terminating app due to uncaught exception 'OC_PythonException', reason: '<class 'AttributeError'>: 'NoneType' object has no attribute 'applicationDockMenu_''
```

Test System
-----------

### Which version of Quod Libet?
Quod Libet v4.7

### Which operating system 
MacOS 15.4

### If it's audio-related, what back-end?
This was not an audio issue.

Additional Information
----------------------
I've already remediated the issue. I just wanted to create an issue to have a breadcrumb trail for the future. This was a pretty simple patch, however, I went ahead and included a required MacOS package within Poetry alongside the PR. Feel free to ping me if you want it removed. It's not required for the patch, but it is required when configuring the app for MacOS development.

Issue #2060, opened 2016-10-09
Running quodlibet.py fails on macOS 10.12


timpulver:
Running

```
source env.sh; jhbuild shell; ../quodlibet/quodlibet.py
```

fails with:

```
/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file 'shell': [Errno 2] No such file or directory
2016-10-09 15:56:05.961 Python[40296:2294311] *** WARNING: Method userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be used in new applications. Use convertRectToBacking: instead. 
E: 0.553: util.importhelper.load_dir_modules: empty.py:22:EmptyBar: NameError: name '_' is not defined

(quodlibet:40296): Gdk-WARNING **: GdkQuartzDisplay does not implement the monitor vfuncs
Illegal instruction: 4

```

or :

```
/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python: can't open file 'shell': [Errno 2] No such file or directory
2016-10-09 16:00:24.818 Python[40368:2315652] *** WARNING: Method userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be used in new applications. Use convertRectToBacking: instead. 
E: 0.429: util.importhelper.load_dir_modules: empty.py:22:EmptyBar: NameError: name '_' is not defined

(quodlibet:40368): Gdk-WARNING **: GdkQuartzDisplay does not implement the monitor vfuncs
2016-10-09 16:01:03.581 Python[40368:2315652] *** Terminating app due to uncaught exception 'OC_PythonException', reason: '<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'applicationDockMenu_''
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff81ff452b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00007fff966cccad objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff81ff4479 -[NSException raise] + 9
	3   _objc.so                            0x0000000103e3895e PyObjCErr_ToObjCWithGILState + 46
	4   _objc.so                            0x0000000103e2076e method_stub + 5070
	5   _objc.so                            0x0000000103e4f1a2 ffi_closure_unix64_inner + 674
[...]

PR #4762, opened 2025-10-04
Do not try to load GTK `applicationDockMenu_` unless it exists

umikoio:
Check-list
----------

This patch fixes the MacOS crash when starting the application. Previously, the app would crash almost immediately due to `applicationDockMenu_` being `None`. This makes sure that it exists and is not an empty variable.

While not the same issue, it was similar to https://github.com/quodlibet/quodlibet/issues/2060, which caused the app to crash after only a few seconds of being open on MacOS.

NOTE: I included a required package for MacOS in this PR, however, it is not required for the patch. I'm happy to remove it if you just want to keep it "patch-only". This just makes it easier to develop on MacOS since it now includes all required packages in Poetry.

Fixes https://github.com/quodlibet/quodlibet/issues/4761


declension:
> NOTE: I included a required package for MacOS in this PR, however, it is not required for the patch. I'm happy to remove it if you just want to keep it "patch-only". 

Yes please :+1: 

That will make this _much_ easier to merge

akizasoft:
No problem, I removed it

declension:
Thanks

PR #4778, opened 2025-11-05
Rework and enhance Discord Status Message plugin

tulser:
Check-list
----------

 * [X] There is a linked issue discussing the motivations for this feature or bugfix
   * Fixes #3423
   * Fixes #4168
 * [X] Unit tests have been added where possible
   * Created `tests/plugin/test_discord_status.py` in 5ff5beb to test `DiscordStatusMessage.truncate_unicode_text()`
     - **Required manually adding dependencies to workflows**
 * [X] I've added / updated documentation for any user-facing features.
   * Current configuration options for the plugin were not added or changed and still largely reflect current functionality.
 * [X] Performance seems to be comparable or better than current `main`
   * No noticeable performance hit in run-time testing

What this change is adding / fixing
-----------------------------------
This PR reworks the Discord Rich Presence plugin to take advantage of current features/parameters available in pypresence and Discord Rich Presence.

Contrasted with the current plugin, there are several notable changes:
* When playing music in Quod Libet, rich presence in Discord will show a progress slider that indicates current music progress.
* When playing music in Quod Libet, rich presence in Discord displays the "Listening to" prefix (`ActivityType.LISTENING`) to indicate that the user is currently listening to music in the status below the username in the user list. The plugin will use the rich presence details (Status Line 1) to append to the status.
* When music is paused in Quod Libet, rich presence in Discord shows the details with a "(Paused)" suffix and removes the progress slider.
* When music is stopped (i.e. no song is loaded), the rich presence in Discord will be cleared.

This PR also fixes #4168 by truncating characters/graphemes for long detail and state strings that could raise errors when updating rich presence in Discord.

With the above fix, this implementation introduces the [`regex`](mrabarnett/mrab-regex) module (`python3-regex` in Debian, Ubuntu) as a dependency of this plugin for grapheme isolation as it supports Unicode grapheme matching that current `re` does not appear to implement.

Screenshots
-----------
Screenshot_20251104_134803

Screenshot_20251104_132504

Screenshot_20251104_132444

PR #4795, opened 2025-11-25
Fix release workflow for introduced testing dependencies

tulser:
Fixes the release workflow failing in the midst of performing unit tests due to missing imports for the Discord status plugin unit tests. This specifically adds the optional dependencies group `plugin_test` and alters the workflow to install those extras via poetry.

**This is a follow up to #4778.**


declension:
Thanks for finding the issue. Have gone with the less invasive #4806.

Issue #4794, opened 2025-11-24
Add automatic ID3 tagging when downloading podcast episodes *(linked with a Pull Request for a tested solution)*


Stevensonjr:
# **The proposed change**
(edit 12.12.2025 : added a pull request #4818 , written with the help of chatgpt 5.1. tested succesfully on my setup : file was downloaded, tagged, album_art embedded. This PR is only tagging if the podcast is in mp3 format. FLAC or other format could be added separately)

Quod Libet should add support for **automatic ID3 tagging of downloaded podcast episodes**.

Today, Quod Libet’s Podcast Browser correctly displays rich metadata extracted from the RSS feed (title, author, podcast name, description, publication date, episode artwork, etc.), but once an episode is downloaded, the resulting MP3 file contains **no ID3 tags at all**. Consequently, the same episode that appears richly annotated in the Podcast Browser appears in the Library as:

* Unknown title
* Unknown artist
* Unknown album
* No cover art
* No date
* No description

This happens because most podcast feeds serve raw MP3 files without embedded tags, and Quod Libet’s downloader currently writes the audio bytes to disk without applying any metadata.

**Proposed behaviour:**
After a podcast episode is successfully downloaded, Quod Libet should automatically write a standard ID3 tag set into the MP3 file, using the metadata already parsed from the RSS feed. Suggested mappings:

* `TIT2`: episode title
* `TPE1`: author / artist
* `TALB`: podcast name
* `COMM`: episode description / summary
* `TDRC`: publication date
* `APIC`: cover art (if available)
* Optional: `TRCK` episode number, canonical ID, or episode GUID

This makes the downloaded file consistent with the metadata already shown in the UI.

---

# **Benefits of this change**

**For all users of podcast downloads:**

* **Consistency:**
  The metadata visible in the Podcast Browser will match the metadata seen in the Library and in the downloaded file.

* **Improved usability:**
  Downloaded podcast episodes will display correct information inside:

  * Quod Libet’s Library
  * mobile music players
  * car infotainment systems
  * external audio players and tag-based organizers

* **Standards compliance:**
  This behaviour matches that of well-known podcast clients:

  * iTunes / Apple Podcasts
  * gPodder
  * PocketCasts
  * AntennaPod
    These clients all tag the audio file using RSS metadata after download.

* **Predictable indexing:**
  Music libraries and search tools will no longer list downloaded episodes as “Unknown Artist” or “Unknown Title”.

* **No additional parsing or fetching required:**
[...]

PR #4818, opened 2025-12-12
Update download.py to add mp3 tagging after download

Stevensonjr:
link to enhancement proposal #4794
add mp3 tagging based on rss fields and also cover-art.

Check-list
----------

 * [X] I've added / updated documentation for any user-facing features.

Check linked issue/enhancement

### DIFF
--- a/quodlibet/qltk/download.py
+++ b/quodlibet/qltk/download.py
@@ -97,6 +97,69 @@ class DownloadProgress(GObject.Object):
                 f.write(result)
             self.success(song)
             print_d(f"Downloaded to {path} successfully!")
+            # -------------------- Podcast tagging (MP3 / ID3) --------------------
+            try:
+                from mutagen.mp3 import MP3
+                from mutagen.easyid3 import EasyID3
+                from mutagen.id3 import ID3, ID3NoHeaderError, APIC
+                import urllib.request
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


declension:
Thanks. I think the problem here is this place is too general (it's not just for podcasts); that download applies to _any_ song that has a URI. These may be remote Soundcloud stream URLs, remote other URLs, or even localish files `file:///` ones (or `smb://` maybe), which have all the metadata intended in their ID3 tags. 

I'd say if we want to go down this route, it'd need to do some kind of more intelligent detection and merging of any existing tags, and also it might want to be configurable whether we want to overwrite existing tags at all.

Stevensonjr:
> Thanks. I think the problem here is this place is too general (it's not just for podcasts); that download applies to _any_ song that has a URI. These may be remote Soundcloud stream URLs, remote other URLs, or even localish files `file:///` ones (or `smb://` maybe), which have all the metadata intended in their ID3 tags.
> 
> I'd say if we want to go down this route, it'd need to do some kind of more intelligent detection and merging of any existing tags, and also it might want to be configurable whether we want to overwrite existing tags at all.

Thanks @declension ,make sense, did not realise. in this case I agree, I wish I could have this option, but I don't know the product enough where to put it. It's an ideal. Worse case I keep the code on my side an readd it when there is a new release.

PR #4820, opened 2025-12-15
fix: synchronizedLyrics trackname checking for illegal chars

Egezenn:
Hello,

The `synchronizedLyrics` plugin was unable to set it's text and background colors and causing errors when the metadata contains illegal characters such as `?` on Windows.

~~These changes _should_ fix the issues, however I couldn't verify it as I don't have the necessary build environment yet.~~


Egezenn:
I couldn't reproduce the color button issue on a fresh local build, just added the related exception.

declension:
Thanks. Maybe add a `print_w` there to show the user / dev that the exception happened at all.

declension:
Thanks

PR #4826, opened 2025-12-21
WIP: GTK4

declension:
Reanimated the second attempt, and have the LLM on the case now, so this one might actually have a chance one day...


declension:
Definitely progress:

image

declension:
More progress:

image

PR #4827, opened 2025-12-22
Add hierarchical collection art feature to Cover Grid browser

Sappelen:
Implements idea #4702: Adds support for organizing albums into collections with visual cover art navigation.

Features:
- Hierarchical navigation: Collections view -> Albums view
- Uses Album Collection browser's grouping key setting
- Custom collection covers from configurable directory
- Automatic detection of collections based on song tags
- All Collections and No Collection special items
- Back button navigation and breadcrumb display
- Compatible with multi-value tags like ~people

Changes:
- main.py: Added collection view logic and navigation handlers
- models.py: Added CollectionListItem class for collection items
- prefs.py: Added Collection Art preferences section

Related to: https://github.com/quodlibet/quodlibet/discussions/4702

Tests I performed:
-------------------
See https://www.youtube.com/watch?v=4exYJD7-4xc

Cover Grid Preferences:
- A new block "Collection Art" is added in the middle, below the "Album Art" block. It contains two fields with a description each: (1) Show Collection Art" with a yes/no field and (2) "Collection cover directory" with a text field with a prompt. collection_covers_dir in main.py is filled with this new field. If it is empty, the cover grid behaves as it did before any of the adjustments, t.i. showing only album covers and no collection covers. OK, works like it should

Cover Grid:
- Navigation from collection to album and back: OK, works like you expect. If you didn't check the checkbox for collection grid, you don't see any different behaviour. 
- Show "no collection" with the default blank cover for albums and/or tracks that do not fall under a collection, in the same way as there is a "no album" image for tracks without an album defined, so you still have navigation for all trakcs: OK
- Test for tracks with album+album grouping, tracks with only album grouping, tracks with only album, tracks with neither: OK, navigation works for all of those situations
- Show "All collections" image, behaving in the same way as the "All albums" image does for album grid (so, do not show it or show it depending on the flag "Show All Albums Item" flag in Cover Grid Preferences): 1st test TOO SLOW, 2nd test OK
- New text "Collections" (in bold font) plus the name of the collection on top of the grid: OK
- Test if view is refreshed at the right moments by clicking through the various options, preferences and browsers: OK
- Test what happens if you leave the collection folder path empty, but set Show Collection Art to yes: OK, that works too, you see collections with empty covers, as you would expect

TODOs. 
The following 2 things I know not to be 100% perfect yet: 
- the selection of the folder in the preferences popup. That should have a prompt similar to menu file->preferences->Library, Scan Directories. I asked Claude code to reuse that "+" button logic, but it did something slightly different. So as a workaround, you copy the folder text from the prompt and then paste it, or you type it
-  it does work for custom album collection album_grouping_key, but not for all other album collection groupings


Sappelen:
Still waiting for someone to implement this PR

Sappelen:
Could someone remove the stale label please? This PR is still active

Sappelen:
Could someone remove the stale label please? This PR is still active

Issue #4788, opened 2025-11-13
Fedora 43 / Python 3.14: PEP 765 warning fires


j39m:
This issue is specific to Fedora 43, [which ships Python 3.14](https://discussion.fedoraproject.org/t/f43-change-proposal-python-3-14-system-wide/133979/7). This version of Python incorporates [PEP 765](https://peps.python.org/pep-0765/), which warns on using a `break` inside a `finally:` block. (Casual readers can refer to [LWN's coverage](https://lwn.net/Articles/1000635/).)

QL will warn on this, e.g. when using `quodlibet --random`. Example:

```
$ quodlibet --random=album
/usr/lib/python3.14/site-packages/quodlibet/util/winpipe.py:161: SyntaxWarning: 'break' in a 'finally' block
  break  # noqa
```

The line is [here](https://github.com/quodlibet/quodlibet/blob/7d3a98eb875519e507bbd559777693a92e54c0f0/quodlibet/util/winpipe.py#L161).

Stevensonjr:
HI. Not sure if related. I tried to download a podcast, also on fedora 43, and sorry I'm still somhow noob. It fails aswell, maybe also python 3.14 related ?

```
AttributeError: 'int' object has no attribute 'replace'
------
Traceback (most recent call last):

  File "/usr/lib/python3.14/site-packages/quodlibet/qltk/songsmenu.py", line 375, in _finished
    f"{util.bold(successes)} "
       ~~~~~~~~~^^^^^^^^^^^

  File "/usr/lib/python3.14/site-packages/quodlibet/util/__init__.py", line 234, in bold
    return f"<b>{escaper(string)}</b>"
                 ~~~~~~~^^^^^^^^

  File "/usr/lib/python3.14/site-packages/quodlibet/util/__init__.py", line 225, in escape
    return string.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
           ^^^^^^^^^^^^^^

AttributeError: 'int' object has no attribute 'replace'

```

PR #4838, opened 2026-01-18
winpipe: Address PEP 765 warning

j39m:
Rework `NamedPipeServer.run()` by moving some control flow into a new `_run_loop()` method. We can break out of this new method by raising an exception or by noticing that `self._stopped` has been set.

This addresses quodlibet/quodlibet#4788.

Check-list
----------

I don't have a Windows device to test with, so this pull is on thin ice to start with :(

Basic playback works on my Linux setup, but I don't think that tells us anything useful about the code changed here. `pylint` doesn't even know what to make of `ctypes.WinError`.

 * ✅ There is a linked issue discussing the motivations for this feature or bugfix
 * ❌ Unit tests have been added where possible
 * ✅ I've added / updated documentation for any user-facing features.
 * ❌ Performance seems to be comparable or better than current `main`

Note: since using a `break` in `finally` swallows any raised exceptions, I've tried to carry this behavior forward by silently `break`ing the outer loop when that particular path is hit.

### DIFF
--- a/quodlibet/util/winpipe.py
+++ b/quodlibet/util/winpipe.py
@@ -63,6 +63,8 @@ class NamedPipeServer(threading.Thread):
 
     """
 
+    BUFFER_SIZE = 4096
+
     def __init__(self, name, callback):
         """name is the name of the pipe file (should be unique I guess)
         callback will be called with new data until close() is called.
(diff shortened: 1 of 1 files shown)


j39m:
Hmm. I don't see the connection between `test_https` and `winpipe`, so I'm mainly rereading the code I changed to see what I messed up. Any pointers would be much appreciated.

declension:
Not sure, but some of those tests are flaky. Maybe try updating from `main` anyway, a re-run might fix it.

j39m:
Thanks — looks like the failure went away. The new flatpak failure is flush with `main`.

declension:
Great - thanks

Issue #4844, opened 2026-01-23
Import loader is not compatible with Python 3.15


luk1337:
Steps to reproduce
------------------
Just try to start it on system with Python 3.15

Expected Output
---------------
Works.

Actual Output
-------------
Broken.

Test System
-----------

### Which version of Quod Libet?
02505e48752b609dd7ecca25df6aaba69587eab2

### Which operating system 
Fedora rawhide

### If it's audio-related, what back-end?
n/a

Additional Information
----------------------
```
+ /usr/bin/pytest -m 'not network'
ImportError while loading conftest '/builddir/build/BUILD/quodlibet-4.7.1-build/quodlibet-4.7.1/tests/conftest.py'.
tests/__init__.py:26: in <module>
    import quodlibet
quodlibet/__init__.py:23: in <module>
    from .util.i18n import _, C_, N_, ngettext, npgettext
quodlibet/util/__init__.py:29: in <module>
    from senf import fsnative
E   AttributeError: 'RedirectImportHook' object has no attribute 'exec_module'. Did you mean: 'find_module'?
```
https://bugzilla.redhat.com/show_bug.cgi?id=2432343

luk1337:
I tried this, but it fails to load raven:

```diff
diff --git a/quodlibet/_import.py b/quodlibet/_import.py
index eb70c8ea9..e1b005572 100644
--- a/quodlibet/_import.py
+++ b/quodlibet/_import.py
@@ -57,6 +57,15 @@ class RedirectImportHook:
         sys.modules[loadname] = mod
         return mod
 
+    def create_module(self, spec):
+        return None
+
+    def exec_module(self, module):
+        target_name = f"{self._name}.{module.__name__}"
+        target_mod = importlib.import_module(target_name)
+        sys.modules[module.__name__] = target_mod
+        module.__dict__.update(target_mod.__dict__)
+
 
 def install_redirect_import_hook():
     """Install the import hook, does not import anything"""
diff --git a/quodlibet/util/importhelper.py b/quodlibet/util/importhelper.py
index fa3cc4215..7c32b0470 100644
--- a/quodlibet/util/importhelper.py
+++ b/quodlibet/util/importhelper.py
@@ -103,7 +103,9 @@ def load_module(name, package, path):
         parent_spec = importlib.machinery.ModuleSpec(package, None, is_package=True)
         sys.modules[package] = importlib.util.module_from_spec(parent_spec)
 
-    mod = spec.loader.load_module(fullname)
[...]

PR #4846, opened 2026-01-23
Update module loader for Python 3.15

luk1337:
Check-list
----------

https://github.com/quodlibet/quodlibet/issues/4844


luk1337:
@lazka @declension can you take a look at this change? I've been using it for couple days with no issues. BTW load_module() has been deprecated since 3.4 according to docs: https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module

Issue #3109, opened 2019-01-13
Set list of column (including custom columns) from CLI


eauchat:
The proposed change
-------------------

The CLI quodlibet provides is really great to highly customize interface and implement any system shortcut to a filtered display for example.
However, for some setups (example a panned browser showing music filtered as classical) you might want a specific list of columns different than one for another setup (example a panned browser showing music filtered as audiobook).
Changing the list of columns, as far as I know can only be achieved through the UI, and take some time. If I could change it from CLI, I could create a simple bash script with a preset of what columns I want to see and I just have to run it through a shortcut to completely change the look of quodlibet). This would be really cool and powerful feature to allow user to create any preset, and change them on the go with just one command line.

Benefits of this change
-----------------------

Instant interface customization, allowing having multiple radically different interfaces with the panel browsing view.

Any issues this might cause
---------------------------

ø

PR #4851, opened 2026-01-31
Added --set-panes cli option, to be able to change currently visible panes from cli

eauchat:
All is in the title, it's a direct implementation of the feature asked in #3109.

It's my first time contributing to quodlibet code, so I hope this PR is formed in an appropriate way. Also if there's anything missing to it let me know and I'll do my best to add the necessary.
I haven't added a test, but looking through them, I did not really know if it was necessary to add some for this. If necessary, let me know which test I should add where and I'll try my best to make one.

As mentioned in the issue, this is a small addition, but allow to add quite a bit of flexibility to quodlibet, making it relevant for many type of ways to organize at once. Hoping this is a welcome change :)

Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix: #3109
 * [x] I've added / updated documentation for any user-facing features. — *hope I added it everywhere relevant, if desired I could also add a bit of documentation on how to use the feature*

Simply adds a `--set-panes` cli option.
This allows to run from cli for example `quodlibet --set-panes genre\tgrouping\tartist\t'<albumyear> - <album>'` to actually trigger the modification of panes, if current browser is paned, otherwise nothing is done.

### DIFF
--- a/docs/guide/commands/quodlibet.rst
+++ b/docs/guide/commands/quodlibet.rst
@@ -140,6 +140,9 @@ OPTIONS
 --set-browser=BrowserName
     Set the current browser
 
+--set-panes=<PanesConfig>
+    Set panes for current browser (config for each pane should be separated by \t)
+
 --show-window
     Show main window
 
--- a/quodlibet/browsers/paned/main.py
+++ b/quodlibet/browsers/paned/main.py
@@ -28,7 +28,7 @@ from quodlibet.util import connect_destroy
 from quodlibet.qltk.paned import ConfigMultiRHPaned
 
 from .prefs import PreferencesButton, ColumnMode
-from .util import get_headers
+from .util import get_headers, save_headers
 from .pane import Pane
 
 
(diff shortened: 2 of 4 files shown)


eauchat:
I didn't add autocompletion bits to relevant files because it was saying they are autobuilt. I didn't know if I had to run a command to build them, if necessary let me know how I can do that, or feel welcome to do it :)

declension:
Thanks and sorry for the delay.

Obviously the CI build needs fixing (and merging / rebasing from `main` at this point), and a test of this new feature would be good here. What happens when the user specifies an invalid pane?

eauchat:
Thanks for the attention @declension :)
I've never written any tests for quodlibet so that won't be an easy task for me, also at the moment I won't be able to do it soon I'm afraid, so if anyone feels like doing so it's very welcome. If not I'll try to do it when I can find time for that.

About the behaviour when an invalid pane is specified, it's actually the same as when one does this from the panes preference UI, it show the pane, but since no matching tag can be found with this invalid tag value it displays one only entry in this pane: `Unknown` containing all songs.

PR #4853, opened 2026-01-31
Move show_all() call down in AlbumArtWindow::__init__()

luk1337:
Check-list
----------

Somehow this fixes Flatpak unit tests with runtime 49.


luk1337:
ah nvm, this seems to just be unlucky timing on my end.

Issue #4641, opened 2025-01-09
Discord Status Message plugin causes error if plugin is active & Discord is closed


curtissimo41:
Steps to reproduce
------------------
Enable the Discord Stats Message plugin, close Discord, play a song.

Expected Output
---------------
The song will play, and the program will not pull up an error message.

Actual Output
-------------
Error details:

```
PipeClosed: The pipe was closed. Catch this exception and re-connect your instance.
------
Traceback (most recent call last):

  File "/home/curtis/.local/lib/python3.10/site-packages/pypresence/baseclient.py", line 79, in read_output
    preamble = await asyncio.wait_for(self.sock_reader.read(8), self.response_timeout)

  File "/usr/lib/python3.10/asyncio/tasks.py", line 445, in wait_for
    return fut.result()

  File "/usr/lib/python3.10/asyncio/streams.py", line 650, in read
    raise self._exception

  File "/usr/lib/python3.10/asyncio/selector_events.py", line 924, in write
    n = self._sock.send(data)

BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "/usr/lib/python3/dist-packages/quodlibet/plugins/events.py", line 140, in __invoke
    handler(*args)

  File "/usr/lib/python3/dist-packages/quodlibet/ext/events/discord_status.py", line 101, in plugin_on_unpaused
    self.handle_unpaused()

  File "/usr/lib/python3/dist-packages/quodlibet/ext/events/discord_status.py", line 90, in handle_unpaused
    self.handle_play()

  File "/usr/lib/python3/dist-packages/quodlibet/ext/events/discord_status.py", line 82, in handle_play
    self.update_discordrp(details, state)

  File "/usr/lib/python3/dist-packages/quodlibet/ext/events/discord_status.py", line 64, in update_discordrp
    self.discordrp.update(details=details, state=state,

  File "/home/curtis/.local/lib/python3.10/site-packages/pypresence/presence.py", line 34, in update
    return self.loop.run_until_complete(self.read_output())

  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()

  File "/home/curtis/.local/lib/python3.10/site-packages/pypresence/baseclient.py", line 83, in read_output
    raise PipeClosed

pypresence.exceptions.PipeClosed: The pipe was closed. Catch this exception and re-connect your instance.

```

Test System
-----------

[...]

PR #4855, opened 2026-02-08
Discord presence plugin: Add new Exception type emitted by pypresence

klausman:
More recent versions of pypresence emit a new exception type. This usually fires if Discord was running at some point (and we had an open pipe to it), but has since closed. When we then emit another song change/status update, the exception `pypresence.exceptions.PipeClosed` is thrown. Since there is not much we can do except ignore it, we do just that. If Discord comes back, the client will automatically recover and status updates will happen again.

Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix: #4641

Fixes new spurious error messages due to pypresence using a new/different exception type, see issue #4641


declension:
Thanks

PR #4860, opened 2026-02-11
Update deps

lazka:


declension:
Oh - I was deliberately pinning poetry at 2.2.1 to be [the supported Nixpkgs 25.11](https://search.nixos.org/packages?channel=25.11&query=poetry&show=poetry) version. 

But I guess I can globally install or something

lazka:
oh, sorry, I didn't notice. I bumped it since the lockfile form 2.3 isn't compatible. I can revert if wanted and use an older lock file.

I should really look into the Windows uv issue...

lazka:
reverted

declension:
Thanks! I'll merge from `main` again (into `gtk4` branch... don't ask...)

Issue #2016, opened 2016-08-27
MPD search support


nkuttler:
I have implemented the mpd search command, 12927521780695384f402868f46049396870eed1 and can use it in the sonata client. My main goal is to use it with MPDroid though, but that seems to work differently. I'll have to have another look at the code before creating a PR but would be happy about any feedback.

nkuttler:
Seems like all there is to do for MPDroid is to implement the `find` and `add` methods. The lookup is very dumb at the moment. For `add` I think the files should be added to the queue, I'm just looking into how I can do that.

nkuttler:
MPDroid mostly works. There is apparently a glitch with add to queue + clear it. Btw, yes, I only use the queue, that's good enough for MPDroid I think. Other clients might try to create playlists on the fly etc.

[my branch](https://github.com/nkuttler/quodlibet/tree/mpd_search)

lazka:
Looks good, some mpdroid crashes, I guess due to the queue state not getting synced.
Only using the queue sounds good for a start.

I think we will need a revision counter for the song lists, I'll look into adding one

nkuttler:
I think some crashes are related to a lack of input validation, e.g.
newlines in tags or missing modification time for some queries. I hope
to get another look next weekend.

There are also some differences in how search and find work that I
haven't implemented yet. They don't seem super important though.

TombusPL:
I connect to the subject. Is it possible to add for a plugin MPD a function queuing, browsing the collections and playlists?

By the way, I use the SMPC on Sailfish OS smartphone and Sonata for Linux;)

nkuttler:
@TombusPL I'm not really sure what you mean? Yes, quodlibet could send the quodlibet playlists to the mpd client (I have already written some code related to that). Collections could be added as well I guess, by I'm not sure how useful it would be. The only way I can see to add them would be through the normal file browser, where collections become part of the "filesystem" as a directory. The mpd client I use to test offers poor support for directories (e.g. you can't add a directory to the current playlist). I've been too busy lately to work more on this code, but I've pushed my latest changes and added a `mpd_search_playlists` branch (not really sure that code is viable).

To be honest, I might focus more on playlists and not work a lot on the search any more, as I'm beginning to think that a quodlibet that can import/export playlists to and from mpd is all I really need. Running mpd on my SBC seems like a better option than running quodlibet.

TombusPL:
I have sorted my music in playlists.
[...]

PR #4864, opened 2026-02-21
Mpd query

nemethf:
Check-list
----------

 * [?] There is a linked issue discussing the motivations for this feature or bugfix

I use KDE Connect to remotely interact with Quod Libet (through the mpris procol), but it does not make possible to remotely enqueue songs.  This PR allows the user to  run queries against the QL database and enqueue songs from the result.  Issue #2016 seems to request the same feature.  This one only implements search related mpd commands, the find related ones are omitted.

I don't know what is the policy here, but I used an AI assistant.  However, the resulting patch is mostly readable, or at least better than what I could have written.  Although maybe the filter expression parser of filtering.py is not easy to understand.

I'm not sure I added and modified copyright notices correctly.  Please, double check them.

If this PR is accepted, I can can work on implementing additional mpd commands (playlist, channels, and others).

Thanks!

### DIFF
--- a/quodlibet/ext/events/mpdserver/__init__.py
+++ b/quodlibet/ext/events/mpdserver/__init__.py
@@ -62,8 +62,7 @@ class MPDServerPlugin(EventPlugin, PluginConfigMixin):
     PLUGIN_NAME = _("MPD Server")
     PLUGIN_DESC = _(
         "Allows remote control of Quod Libet using an MPD Client. "
-        "Streaming, playlist and library management "
-        "are not supported."
+        "Streaming and library management are not supported."
     )
     PLUGIN_ICON = Icons.NETWORK_WORKGROUP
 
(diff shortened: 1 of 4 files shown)


declension:
Sorry about the delay, and thanks. It's a bit hard to review (any simplifications possible would be good), but appreciate the passing tests and the updated one.

The main problem is I have no mpd setup to test this with, so we'd have to trust that you've tested this manually

nemethf:
>  It's a bit hard to review (any simplifications possible would be good), but appreciate the passing tests and the updated one.

The PR is split into three commits.  Maybe it is easier to review them separately. 

> The main problem is I have no mpd setup to test this with, so we'd have to trust that you've tested this manually

Would it help if I find you a simple local mpd client to test the code with?  Or help you to use the M.A.L.P. android client?  

At any rate, I did test this and it works, but I ran into small annoyances with `myMPD`, so I kept adding protocol features and now I have more than 40 additional commits in my local branch and an almost full implementation of the mpd protocol.  If you accept this PR, do you think it would make sense for me to contribute the whole implementation?

Thanks.

PR #4365, opened 2023-08-31
Connect TreeViewHints->view leave-notify-event to __undisplay

luk1337:
Check-list
----------

 * [x] There is a linked issue discussing the motivations for this feature or bugfix
 * [ ] Unit tests have been added where possible
 * [ ] I've added / updated documentation for any user-facing features.
 * [ ] Performance seems to be comparable or better than current `main`

What this change is adding / fixing
-----------------------------------

This fixes stuck tooltips when leaving the window.

See: https://www.youtube.com/watch?v=x7Ey-y0TPIs
Closes: #4259

Issue #4259, opened 2023-01-21
Residual tooltip


jej:
Steps to reproduce
------------------
1. Hover the mouse on a title, so that a tooltip appears. It appears if the column width is smaller than the title.
2. `Alt-Tab` to switch on another application.
3. The tooltip still appears, even if QL is behind.

Some screenshots:
![screenshot_20230121-185559](https://user-images.githubusercontent.com/33821/213880545-533fcec4-ea2b-4ebb-8f9c-db698c442788.png)

![screenshot_20230121-185513](https://user-images.githubusercontent.com/33821/213880551-5e6dd8ee-3852-4ba1-a993-9f5714d51a41.png)

The problem occurs also when the window is minimized.
1. Hover the mouse on a title, so that a tooltip appear.
2. `Alt-Space` to open the window menu.
3. Minimize the windows.
4. The tooltip still appears, even if QL is minimized.

Expected Output
---------------
Tooltip should appears only on QL window.

Actual Output
-------------
Tooltip appears as a separate layer on the desktop.

Test System
-----------

### Which version of Quod Libet?
4.6 (development)

### Which operating system 
```
~$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
```
with Gnome 3.68

declension:
Thanks for raising! I've seen this for years actually (on Gnome both Ubuntu and Manjaro), never got round to investigating

jej:
Yes, me too! I always had in mind to open this issue.

I tried to investigate, but I am too far from gtk for now...

jej:
Somewhere here maybe:
https://github.com/quodlibet/quodlibet/blob/fb65cb3d7126c8dd9023d6acf3641c99f41bf87c/quodlibet/qltk/views.py#L26

b0ssi:
I can second this issue (Arch, Mate as well as Xfce - former Wayland now, I believe, latter X11). Thanks for keeping this issue open.

Npa972:
This bug disappeared for a while in the development version but has since reappeared.
Not serious at all but a little  annoying.

sebastian-65:
Happens to me too all the time, quodlibet 4.7.1-2, Arch, Xfce.

declension:
[...]

PR #4870, opened 2026-03-13
Fix residual tooltips on some systems (#4259)

nemethf:
Check-list
----------

There was an old pull request (#4365) by @luk1337, which fixed the residual tooltip (#4259), but it seem the PR caused troubles on windows.  Since windows is a non-free platform, I think, it should not hinder the progress on other platforms.  So, this PR fixes the issue where we know a fix.  See the discussion in #4259. 

I've added @luk1337 as co-author, because this PR just selectively applies the original idea.


declension:
Thanks

Issue #4661, opened 2025-02-27
No macOS Build of latest release?


rafo:
The proposed change
-------------------
macOS builds for the latest versions. The last macOS Build is 4.4. Installing 4.4 with brew results in a non-working app under macOS 15.3.1 (24D70). It simply does not start.

There might be problems with building the app #4638

Benefits of this change
-----------------------
A working app.

Any issues this might cause
---------------------------
More users ;D

rafo:
Sorry! I did not check the PR #4639

Maybe this issue is obsolet.

VGerris:
The above PR is merged, but I do not see a MacOS release for 4.7.1. 
Happy to see one and test it!

jost-s:
@declension @lazka The build for macos is fixed and automated now. I assume that the release process has been manual.

Are you interested in automated releases?

declension:
Yes most things are manual still due to legacy / coordination / release notes / distribution channels / secrets etc. An increasing amount of scripting / CI support is there at least. But naturally automation is what we do want really (not on every commit, naturally)

Issue #3477, opened 2020-10-15
MacOS build failing


declension:
Steps to reproduce
------------------
Push (probably) working code to a branch

Expected Output
---------------
No build failures

Actual Output
-------------
MacOS build consistently failing
```
2020-10-15T07:55:27.5315200Z Collecting iniconfig
2020-10-15T07:55:27.5369690Z   Downloading iniconfig-1.1.1.tar.gz (8.1 kB)
2020-10-15T07:55:27.5563480Z   Installing build dependencies: started
2020-10-15T07:55:30.9268430Z   Installing build dependencies: finished with status 'done'
2020-10-15T07:55:30.9332420Z   Getting requirements to build wheel: started
2020-10-15T07:55:30.9753970Z   Getting requirements to build wheel: finished with status 'error'
2020-10-15T07:55:30.9755480Z   ERROR: Command errored out with exit status 2:
2020-10-15T07:55:30.9757560Z    command: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/_app/Contents/MacOS/python3 /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/_app/Contents/Resources/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmppz5lyrqw
2020-10-15T07:55:30.9759350Z        cwd: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-v83hf8hv/iniconfig
2020-10-15T07:55:30.9759930Z   Complete output (1 lines):
2020-10-15T07:55:30.9761950Z   /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/_app/Contents/MacOS/python3: can't open file '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/_app/Contents/Resources/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py': [Errno 2] No such file or directory
```

[...]

Issue #3348, opened 2020-04-16
fails to build from source on macOS 10.15 due to PyObjC issue


smammy:
Steps to reproduce
------------------
```
git clone https://github.com/quodlibet/quodlibet.git
cd quodlibet/dev-utils/osx_bundle
./bootstrap.sh
./build.sh
```

Expected Output
---------------
```
[thousands of lines of output, removed]
I: Install complete: 703 files copied
*** success *** [94/94]
```

Actual Output
-------------
```
[thousands of lines of output, removed]
*** Building python-pyobjc-core *** [49/94]
python3 setup.py build --build-base /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1
setup.py:624: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
  cfg.readfp(fp)
running build
running build_py
Overriding build_packages to copy PyObjCTest
creating /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1
creating /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1/lib.macosx-10.9-x86_64-3.7
creating /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1/lib.macosx-10.9-x86_64-3.7/objc
copying Lib/objc/_bridges.py -> /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1/lib.macosx-10.9-x86_64-3.7/objc
copying Lib/objc/_protocols.py -> /Users/sam/Files/Projects/Music/quodlibet-bug-report/quodlibet/dev-utils/osx_bundle/_home/.cache/jhbuild/build/python-pyobjc-core-4.2.1/lib.macosx-10.9-x86_64-3.7/objc
[...]

PR #4880, opened 2026-03-23
Upgrade dependencies and fix macos build

jost-s:
Check-list
----------

   https://github.com/quodlibet/quodlibet/issues/3477
   https://github.com/quodlibet/quodlibet/issues/4504
   https://github.com/quodlibet/quodlibet/issues/3348
   https://github.com/quodlibet/quodlibet/issues/4661

- Fix build locally on macOS 15.7 by upgrading dependencies and build scripts
- Adds a job to the release workflow that builds quodlibet on macOS and uploads the bundled artifacts.
- This new job caches build artifacts, keyed by the hashed dependencies.
- Upgrades GitHub actions of the release workflow to latest version

The [build on macOS from scratch](https://github.com/jost-s/quodlibet/actions/runs/23445322017/job/68207300258) takes about 1 h on CI. This is [a run of the release workflow that uploads the bundled artifacts](https://github.com/jost-s/quodlibet/actions/runs/23457204923), using the build cache.

Let me know if this is a desired direction. The `.dmg` is uploaded as an artifact. I can also see how the build process could be done with nix, but that would mean a larger refactor of the build process.

Also let me know if I should squash commits.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -88,13 +88,15 @@
                 gtksourceview4
                 kakasi
                 keybinder3
-                libappindicator-gtk3
                 libmodplug
                 libnotify
                 librsvg
                 libsoup_3
                 pcre2
                 shared-mime-info
+              ]
+              ++ lib.optionals stdenv.isLinux [
+                libappindicator-gtk3
                 xorg.xvfb
               ]
               ++ (with gst_all_1; [
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,18 +57,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
-
-      # - name: Cache jhbuild dependencies
-      #   uses: actions/cache@v4
-      #   with:
-      #     path: |
-      #       dev-utils/osx_bundle/_home/jhbuild_checkoutroot/pkgs
-      #       dev-utils/osx_bundle/_home/jhbuild_prefix
-      #       dev-utils/osx_bundle/_home/.cache/jhbuild/build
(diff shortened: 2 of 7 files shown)


flake.nix
```diff
@@ -88,13 +88,15 @@
...
-                libappindicator-gtk3
                 libmodplug
                 libnotify
                 librsvg
                 libsoup_3
                 pcre2
                 shared-mime-info
+              ]
+              ++ lib.optionals stdenv.isLinux [
+                libappindicator-gtk3
```
jost-s, line 99: These packages don't exist for macos systems.

declension:
Sorry, been away / sick so only just seeing this but sounds promising (and thanks).

Not a macOS user so limited insight here. Obviously we can't have a 1h job in the main CI loop, so there's a risk of bitrot which might need some thought.

But seems like an improvement to what we have. @lazka any thoughts?

lazka:
Maybe separate like https://github.com/quodlibet/quodlibet/blob/main/.github/workflows/wininst.yml so we can trigger it manually for starters?

jost-s:
[The build only takes 8 mins from cache](https://github.com/jost-s/quodlibet/actions/runs/23457204923), and the cache is keyed by the macos depdendencies, which seem to get updated ~2x per year. Maybe it's acceptable to wait 1 h if it's that rare.

declension:
> [The build only takes 8 mins from cache](https://github.com/jost-s/quodlibet/actions/runs/23457204923), and the cache is keyed by the macos depdendencies, which seem to get updated ~2x per year. Maybe it's acceptable to wait 1 h if it's that rare.

Yes that's much more acceptable actually. Maybe we leave like this and split it out if and only if it gets in the way @lazka

lazka:
Ok

jost-s:
@declension @lazka Who merges PRs around here?

declension:
Thanks for the PR @jost-s

Issue #4025, opened 2022-06-24
Multimedia keys either don't work in Quod Libet, or ONLY work in Quod Libet


PisauraXTX:
Steps to reproduce
------------------
1. Play media in an application other than Quod Libet, for example a Youtube video in your browser. Pause playback with the multimedia keys on your keyboard. 
2. Open Quod Libet. Try to start playback in Quod Libet with the multimedia keys on your keyboard. 

Alternatively, 
1. Open Quod Libet and use the multimedia keys on your keyboard to control playback. 
2. Switch to a different media stream in a different application and try to control playback there with the multimedia keys. 

Expected Output
---------------
The focus of the multimedia keys should always lie on the media stream you last interacted with. 

Actual Output
-------------
Quod Libet ignores input from multimedia keys when another application with a media stream has been opened first. If Quod Libet is opened first, focus of the multimedia keys will always remain on Quod Libet and it is not possible to control other media streams with them. 

Test System
-----------

### Which version of Quod Libet?
4.5.0

### Which operating system 
Arch Linux, Kernel 5.18, KDE Plasma 5.25.1

### If it's audio-related, what back-end?
No pipeline selected within Quod Libet. The system is using Pipewire. 

Additional Information
----------------------
Tested with Microsoft Sidewinder X6 keyboard using sidewinderd. Switching focus of the multimedia keys back and forth works flawlessly with other applications such as Firefox, Chromium, VLC, Kasts, etc. The problem only occurs when using Quod Libet.

smiledra:
Hi, I had the same problem and installing `python-paho-mqtt` fixed it for me after enabling MPRIS D-Bus Support in Plugins. This previously helped me on a PopOS machine and worked again on my ArchLinux machine. cf. https://github.com/quodlibet/quodlibet/issues/2692#issuecomment-759090404

I'm also on Quod Libet 4.5.0 and using `pipewire`. Wondering if this would help!

PisauraXTX:
> Hi, I had the same problem and installing `python-paho-mqtt` fixed it for me after enabling MPRIS D-Bus Support in Plugins. This previously helped me on a PopOS machine and worked again on my ArchLinux machine. cf. [#2692 (comment)](https://github.com/quodlibet/quodlibet/issues/2692#issuecomment-759090404)
> 
> I'm also on Quod Libet 4.5.0 and using `pipewire`. Wondering if this would help!

Hello smiledra, 
[...]

Issue #3316, opened 2020-03-03
Media keys don't work on Mac with 4.3


elcste:
Media keys (play/pause, forward, back) don't work on Mac with new version 4.3. I'm using Catalina (10.15.3). If I go back to 4.2.1, it works again. (Even with the previous versions, sometimes I would have to reboot to get this to work if I accidentally opened iTunes/Apple Music, but I've tried rebooting with 4.3 with no luck, and then opening 4.2.1 works without issue.

LivArt01:
Confirmed issue there too.

elcste:
Back with Quod Libet after a hiatus (it's too good to stay away!), and this is still an issue with QL 4.4.0 and macOS 12.2.1.

But, there's an easy workaround: use [Beardie](https://github.com/Stillness-2/beardie/). Just run it and it works without any configuration, and you can set it to autolaunch and hide the menubar icon.

jost-s:
I suppose it's just a matter of re-adding QL to the apps under Accessibility https://github.com/quodlibet/quodlibet/issues/2692#issuecomment-4322285373

Issue #2692, opened 2018-01-04
macOS 10.13 + mmkeys


lazka:
Check if it's working: https://github.com/quodlibet/quodlibet/issues/1817#issuecomment-355088672

palfrey:
Not working for me with 10.14.3 and 4.2.1

palfrey:
Hmm. Not sure what I did there. It's now working fine. I downloaded a copy of the repo, ran it directly from there, and it was working. Tried looking at the mmkeys history, found nothing notable since 4.2.1 for OS X, and then just tried 4.2.1 and suddenly it works!

elcste:
I have 10.14.3 at work, and it seems to work the first time I launch Quod Libet after restarting the computer. But if I ever quit and reopen it, the keys don't work until I reboot again. I only tried once, but logging out and back in didn't seem to work either.

tmikaeld:
Doesn't work on 10.4.4 either, Quodlibet 4.2.1

LivArt01:
Not working 4.3.0

onasipim:
Also not working on macOS 10.15.5 with Quodlibet 4.3.0

faiwer:
mac: 10.15.7
Quodlibet: 4.3.0
mmkeys don't work

cmd+. & cmd+, work, but only when the window is active

fduchatea:
Running MacOS 10.14.6, the mmkeys worked fine with Quodlibet 4.2.1. 
I updated to Quodlibet 4.3.0, and the mmkeys do not work anymore.
In systems preferences, Quodlibet is authorized to manipulate these keys (tried to delete and add again authorization, but still not working).

smiledra:
Not sure if it's related but: I have PopOS 20.10 and media keys (F8 F9 F10) are not working for Quod Libet 4.3.0. The same keys work for all other things (browser media playing, rhythmbox, etc.). In Plugins > Advanced Preferences, "Disable multimedia keys" is set to false.

smiledra:
> Not sure if it's related but: I have PopOS 20.10 and media keys (F8 F9 F10) are not working for Quod Libet 4.3.0. The same keys work for all other things (browser media playing, rhythmbox, etc.). In Plugins > Advanced Preferences, "Disable multimedia keys" is set to false.

Aaand after I saw an error in the plugins page "Couldn't find 'paho-mqtt'" I went to install this module (cf. #2828) through pip, and then mmkeys are recognized after restarting Quod Libet.

[...]

PR #4899, opened 2026-05-08
Fix multimedia keys on macos

jost-s:
Check-list
----------

Multimedia keys on macos weren't working without granting accessibility privileges in system settings. When granted, QL would claim all events and other apps weren't controllable with media keys any longer. macOS's current media management service wasn't used at all.

Now QL registers the proper event handlers for the Remote Command Center. It displays the info of the currently played artist and title, allows for scrubbing in the timeline and displays artwork. Accessibility privileges are no longer needed.

Tests on macos can now be run on the current source code without downloading the static .DMG file. Tests for the mm keys were added. The documentation for running tests on macos was updated too.

Related issues https://github.com/quodlibet/quodlibet/issues/2692, https://github.com/quodlibet/quodlibet/issues/4025, https://github.com/quodlibet/quodlibet/issues/3316

### DIFF
--- a/quodlibet/mmkeys/keybinder.py
+++ b/quodlibet/mmkeys/keybinder.py
@@ -6,7 +6,12 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import sys
+
 from ._base import MMKeysBackend, MMKeysAction, MMKeysImportError
+
+if sys.platform != "linux":
+    raise MMKeysImportError()
 
 import gi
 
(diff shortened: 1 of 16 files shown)


quodlibet/mmkeys/keybinder.py
```diff
@@ -6,8 +6,13 @@
 # the Free Software Foundation; either version 2 of the License, or
 # (at your option) any later version.
 
+import sys
+
 from ._base import MMKeysBackend, MMKeysAction, MMKeysImportError
 
+if sys.platform != "linux":
+    raise MMKeysImportError
```
declension:
```suggestion
    raise MMKeysImportError()
```

declension (approved):
If this works for your setup, then LGTM

quodlibet/mmkeys/keybinder.py
jost-s: Right, fixed in https://github.com/quodlibet/quodlibet/pull/4899/commits/502bb48686d444593223da29d7d6afeba774b444

declension:
Thanks

PR #4902, opened 2026-05-12
fix(brainz): preserve existing albumartist on MusicBrainz lookup

### LINKED ISSUE #4900: MusicBrainz Lookup plugin deletes albumartist tag
Steps to reproduce
------------------
Scenario 1:
1. Enable MusicBrainz Lookup plugin
2. Run on a single-artist album that has some songs that have additional artists and has an albumartist tag
    - E.g. Syntesizer by Outkast off of Aquemini or Shady by Adam Lambert off of Tresspassing
3. Note that albumartist is deleted

Scenario 2:
1. Uncheck option "Write albumartist when needed" for plugin
2. Run on any album, including multi-artist albums, where albumartist is already set
3. Note that albumartist is deleted

Expected Output
---------------
Existing albumartist tags should not be deleted in either scenario.  Missing albumartist tags should be populated if 'Write "albumartist" when needed' is selected and existing tags should be overwritten.

Actual Output
-------------
Tag is deleted

Test System
-----------

4.7.1

### Which operating system 
Mint 22.3

N/A

Additional Information
----------------------

SAY-5:
Check-list
----------

Fixes #4900.

`apply_to_song` treats an empty meta value as a delete (it calls `song.remove(key)`), so the MusicBrainz Lookup plugin's two places that emit `meta["albumartist"] = ""` end up silently dropping the existing albumartist tag from the file:

1. `build_song_data`, on a single-artist or various-artists release.
2. `apply_options`, when the "Write `albumartist` when needed" option is disabled.

This patch fixes both:

- `build_song_data` now always populates `albumartist` / `albumartistsort` / `musicbrainz_albumartistid` from the release artists. The MB release artist is the correct value on any release, including single-artist ones.
- `apply_options` now `pop`s `albumartist` from the meta dict when the option is off, so `apply_to_song` simply leaves the on-disk tag alone.

Two regression tests are added to `tests/plugin/test_brainz.py` covering both scenarios.

### DIFF
--- a/tests/plugin/test_brainz.py
+++ b/tests/plugin/test_brainz.py
@@ -602,13 +602,13 @@ class TBrainz(PluginTestCase):
         assert release.is_single_artist
         track = release.tracks[1]
         meta = build_song_data(release, track)
-        self.assertEqual(meta["albumartist"], "Autechre")
+        self.assertEqual(meta["albumartist"], "Autechre\nThe Hafler Trio")
 
         # albumartist option ON: write the MB value.
         apply_options(meta, True, True, False, False, False)
         dummy = AudioFile({"albumartist": "preexisting"})
         apply_to_song(meta, dummy)
-        self.assertEqual(dummy("albumartist"), "Autechre")
+        self.assertEqual(dummy("albumartist"), "Autechre\nThe Hafler Trio")
 
     def test_apply_options_albumartist_off_preserves_existing(self):
         # Regression test for #4900 (Scenario 2): if the user disables
--- a/tests/plugin/test_brainz.py
+++ b/tests/plugin/test_brainz.py
@@ -588,6 +588,48 @@ class TBrainz(PluginTestCase):
         )
         self.assertEqual(dummy("labelid"), "pgram002")
 
+    def test_build_metadata_single_artist_preserves_albumartist(self):
+        # Regression test for #4900: on a single-artist release, the plugin
+        # used to write meta["albumartist"] = "", which caused
+        # apply_to_song() to delete an existing albumartist tag. The release
+        # artist should now be written instead.
+        Release = brainz.mb.Release
+        build_song_data = brainz.widgets.build_song_data
+        apply_options = brainz.widgets.apply_options
+        apply_to_song = brainz.widgets.apply_to_song
+
+        release = Release(TEST_DATA)
+        assert release.is_single_artist
+        track = release.tracks[1]
+        meta = build_song_data(release, track)
+        self.assertEqual(meta["albumartist"], "Autechre")
+
+        # albumartist option ON: write the MB value.
+        apply_options(meta, True, True, False, False, False)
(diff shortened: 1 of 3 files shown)


tests/plugin/test_brainz.py
```diff
@@ -588,6 +588,48 @@ def test_build_mbids_labelid(self):
...
+        assert release.is_single_artist
+        track = release.tracks[1]
+        meta = build_song_data(release, track)
+        self.assertEqual(meta["albumartist"], "Autechre\nThe Hafler Trio")
+
+        # albumartist option ON: write the MB value.
+        apply_options(meta, True, True, False, False, False)
+        dummy = AudioFile({"albumartist": "preexisting"})
+        apply_to_song(meta, dummy)
+        self.assertEqual(dummy("albumartist"), "Autechre\nThe Hafler Trio")
```
declension: We've long since ditched these, but there are quite a few left over:

```suggestion
        assert dummy("albumartist") == "Autechre\nThe Hafler Trio"
```

tests/plugin/test_brainz.py
```diff
@@ -588,6 +588,48 @@ def test_build_mbids_labelid(self):
...
+        build_song_data = brainz.widgets.build_song_data
+        apply_options = brainz.widgets.apply_options
+        apply_to_song = brainz.widgets.apply_to_song
+
+        release = Release(TEST_DATA)
+        track = release.tracks[1]
+        meta = build_song_data(release, track)
+        # albumartist option OFF.
+        apply_options(meta, True, False, False, False, False)
+        self.assertNotIn("albumartist", meta)
```
declension:
```suggestion
        assert "albumartist" not in meta
```

tests/plugin/test_brainz.py
```diff
@@ -588,6 +588,48 @@ def test_build_mbids_labelid(self):
...
+
+        dummy = AudioFile({"albumartist": "preexisting"})
+        apply_to_song(meta, dummy)
+        self.assertEqual(dummy("albumartist"), "preexisting")
```
declension:
```suggestion
        assert dummy("albumartist") == "preexisting"
```

tests/plugin/test_brainz.py
```diff
@@ -588,6 +588,48 @@ def test_build_mbids_labelid(self):
...
+        apply_to_song = brainz.widgets.apply_to_song
+
+        release = Release(TEST_DATA)
+        assert release.is_single_artist
+        track = release.tracks[1]
+        meta = build_song_data(release, track)
+        self.assertEqual(meta["albumartist"], "Autechre\nThe Hafler Trio")
+
+        # albumartist option ON: write the MB value.
+        apply_options(meta, True, True, False, False, False)
```
declension, line 608: Side note: we really should switch to kwargs-only here, it's Boolean Blindness in the extreme

declension (approved):

Issue #1288, opened 2015-03-15
Space bar should be the shortcut for play / pause


lazka:
Original [issue 1288](https://code.google.com/p/quodlibet/issues/detail?id=1288) created by simon.boulier on 2013-11-07T11:45:52.000Z:

I would want to change the space bar shortcut so that it toogles play/pause. Don't you? (Currently it activates the selected button).
It seems to me that it's an usual shortcut in other audio players.

Quodlibet is amazing, thank you very much!

Simon

marcelpaulo:
The accelerator for `<Actions>/QuodLibetWindowActions/PlayPause` is always reset to `<Primary>space` when `QL` starts. I tried:

1. Stop `QL`
2. In `~/.quodlibet/accels`, uncomment the entry for `<Actions>/QuodLibetWindowActions/PlayPause`, and change the accelerator to `space`
3. Start `QL`

The entry was reset to the default: commented and accelerator set to `<Primary>space`.

alexforencich:
This bug is absolutely infuriating.

PatronBernard:
Do the devs actually use their program? It shouldn't be "configurable", spacebar should be the **default** play/pause toggle...

CreamyCookie:
> Also, shouldn't the play button change to the pause button if there is music playing?

Huh? It already does that. Which version are you using?

PatronBernard:
3.5.3., downloaded it in Ubuntu 16.04's software center...

CreamyCookie:
Oh, okay. You need to upgrade to at least [3.6](https://quodlibet.readthedocs.io/en/latest/changelog.html#it-is-altogether-fitting-and-proper-that-we-should-do-this) if you want that:

> Player controls: use a normal button with two icons instead of a toggle button. #1814

You could add the PPA from here: http://quodlibet.readthedocs.io/en/latest/downloads.html#ubuntu-logo-ubuntu

scottyeager:
+1 This is maddening. I really can't imagine a workflow where it makes any sense. Like,
1. Song is playing
2. User navigates to another song
3. User wants to start the new song
4. User presses *Enter* (this works already, why use the space bar?)

As others have mentioned, every single other music/video player in existence works this way. That might not even be an exaggeration. Configurable would be nice, but space bar as a play/pause toggle is the sane default.

Can this be fixed via a plugin or other means that don't require directly modifying the source?

jacebot:
[...]

PR #4919, opened 2026-07-06
fixed spacebar

ironhak:
The spacebar is not play/pause, but it first skip a song and then it becomes play/pause on the first time you play an album. See here: https://github.com/quodlibet/quodlibet/issues/1288

This issue has been running for a long time and it's very frustrating. This is the fix for it.


ironhak:
Hello can you please look into this?

PR #4926, opened 2026-09-01
Fix double-escaping in file renaming error messages

gwymor:
Check-list
----------

The user-controlled text is already escaped by util.bold, so it ended up being escaped twice.


gwymor:
The contributing guidelines say I should add a copyright line to files I touch. Does that include trivial changes like this?

declension:
Thanks

PR #4931, opened 2026-09-13
Fix invalid escape sequences in radio_crawler dev utilities

MBK-fr:
### Problem

The `dev-utils/radio_crawler/` helper scripts build regexes from non-raw string literals with invalid escapes:
- `fetch_xiph.py` - `\[`, `\d`
- `fetch_soma.py` - `\d`

Python 3.12+ raises `SyntaxWarning: invalid escape sequence` for these.

### Why it matters for Quod Libet

The project targets `>=3.10,<4.0` and runs Python 3.12 in CI, where the warnings appear when the scripts are imported.

### Fix

Escape the backslashes; the compiled regexes are unchanged (ruff W605-style).

### Verification

`ast.dump` unchanged per file; `python3 -W error` py_compile passes on both after, failed before.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014HBQNzAf5C2E3MiJC48HQw


declension:
Thanks

